PR #1, opened 2020-02-12
Add TPMConnection and associated helpers

chrisccoulson:


pedronis:
@chrisccoulson should this already include get-deps and .travis.yaml ?  a short README about installing the simulator snap and how to run the tests would also be good

pedronis:
@chrisccoulson in the underlying go-tpm2  the panics about wrong types seem ok, but there are some panics like:

```
 panic(fmt.Sprintf("cannot marshal raw command auth area to temporary buffer: %v", err))
```

etc, what's the rational to panic vs error there?

errors.go
```diff
@@ -0,0 +1,53 @@
...
+	// ErrProvisioning indicates that the TPM is not provisioned correctly for the requested operation. Please note that other errors
+	// that can be returned may also be caused by incomplete provisioning, as it is not always possible to detect incomplete or
+	// incorrect provisioning in all contexts.
+	ErrProvisioning = errors.New("the TPM is not correctly provisioned")
+)
+
+// EkCertVerificationError is returned from SecureConnectToDefaultTPM if verification of the EK certificate against the built-in
+// root CA certificates fails, or the EK certificate does not have the correct properties, or the supplied certificate data cannot
+// be unmarshalled correctly because it is invalid.
+type EkCertVerificationError struct {
```
pedronis: this is annoying but I would expect  the spelling to be s/Ek/EK/ everywhere

tpm.go
```diff
@@ -0,0 +1,937 @@
...
+// TPMDeviceAttributes contains details about the TPM extracted from a manufacturer issued endorsement key certificate.
+type TPMDeviceAttributes struct {
+	Manufacturer    tpm2.TPMManufacturer
+	Model           string
+	FirmwareVersion uint32
+}
+
+// TPMConnection corresponds to a connection to a TPM device, and is a wrapper around *tpm2.TPMContext.
+type TPMConnection struct {
+	*tpm2.TPMContext
```
pedronis: will caller use methods from this directly?

tpm.go
```diff
@@ -0,0 +1,937 @@
...
+	hmacSession              tpm2.SessionContext
+}
+
+// VerifiedEkCertChain returns the verified certificate chain for the endorsement key certificate obtained from this TPM. It was
+// verified using one of the built-in TPM manufacturer root CA certificates.
+func (t *TPMConnection) VerifiedEkCertChain() []*x509.Certificate {
+	return t.verifiedEkCertChain
+}
+
+// VeririedDeviceAttributes returns the TPM device attributes for this TPM, obtained from the verified endorsement key certificate.
```
pedronis: typo in VeririedDeviceAttributes

tpm.go
```diff
@@ -0,0 +1,937 @@
...
+// In order for the TPM to prove it is the device for which the endorsement key certificate was issued, an endorsement key is
+// required. If the TPM doesn't contain a valid persistent endorsement key at the expected location (eg, if ProvisionTPM hasn't been
+// executed yet), this function will attempt to create a transient endorsement key. This requires knowledge of the endorsement
+// hierarchy authorization value, which will be empty on a newly cleared device. If there is no object at the persistent endorsement
+// key index and creation of a transient endorement key fails, ErrProvisioning will be returned.
+//
+// If the TPM cannot prove it is the device for which the endorsement key certificate was issued, a TPMVerificationError error will be
+// returned. This can happen if there is an object at the persistent endorsement key index but it is not the object for which the
+// endorsement key certificate was issued, and creation of a transient endorsement key fails because the correct endorsement hierarchy
+// authorization value hasn't been provided.
```
pedronis: all this doc doesn't mention endorsementAuth []byte which is a bit strange, also given it's []byte is not super obvious what exactly goes in there

tpm.go
```diff
@@ -0,0 +1,937 @@
...
+// which must contain a URL for the parent certificate.
+//
+// This function will stop when it encounters a certificate that doesn't specify an issuer URL, or when it encounters a self-signed
+// certificate.
+//
+// If no endorsement key certificate can be obtained, an error will be returned.
+//
+// If parentsOnly is true, this function will only save the parent certificates as long as the endorsement key certificate can be
+// reliably obtained from the TPM.
+func FetchAndSaveEkCertificateChain(tpm *TPMConnection, parentsOnly bool, dest string) error {
```
pedronis: s/dest/destPath/

tpm.go
```diff
@@ -0,0 +1,937 @@
...
+
+// SaveEkCertificateChain will save the specified EK certificate and associated parent certificates atomically to the specified file
+// in a form that can be decoded by SecureConnectToDefaultTPM. This is useful in scenarios where the EK certificate cannot be located
+// automatically, or the EK certificate or any intermediate certificates lack the Authority Information Access extension but the
+// certificates have been downloaded manually by the caller.
+//
+// If the EK certificate can be obtained reliably from the TPM during establishment of a connection, then it can be omitted in order
+// to save a file that only contains parent certificates. In this case, SecureConnectToDefaultTPM will attempt to obtain the EK
+// certificate from the TPM and verify it against the supplied parent certificates.
+func SaveEkCertificateChain(ekCert *x509.Certificate, parents []*x509.Certificate, dest string) error {
```
pedronis: s/dest/destPath/

tpm.go
```diff
@@ -0,0 +1,937 @@
...
+		if len(certData.Cert) == 0 {
+			// The supplied data only contains parent certificates. Retrieve the EK cert from the TPM.
+			if cert, err := readEkCertFromTPM(tpm); err != nil {
+				return nil, EkCertVerificationError{fmt.Sprintf("cannot obtain endorsement key certificate from TPM: %v", err)}
+			} else {
+				certData.Cert = cert
+			}
+		}
+	} else {
+		data, err := fetchEkCertificateChain(tpm, false)
```
pedronis: what's the advantage of having this path here, vs asking user to call FetchAndSaveEkCertificateChain before, it feels a bit too magical this involving the network implicitly here

chrisccoulson:
> @chrisccoulson in the underlying go-tpm2 the panics about wrong types seem ok, but there are some panics like:
> 
> ```
>  panic(fmt.Sprintf("cannot marshal raw command auth area to temporary buffer: %v", err))
> ```
> 
> etc, what's the rational to panic vs error there?

MarshalToWriter only returns errors in a very limited number of circumstances, and one of those is IO errors which don't apply in this particular case because it's using bytes.Buffer (bytes.Buffer.Write panics rather than returning errors). Other reasons for returning errors are passing a nil pointer for a type that implement tpm2.CustomMarshaller or passing variable length types where the length / size overflows the size field for the type, and none of these apply here (the auth area is marshalled without a length field) so MarshalToWriter can never return an error. The error value could be ignored, but the panic is there in case I make a change to the marshalling code that alters the behaviour of the API which makes it return an error for reasons where it previously didn't.

chrisccoulson:
> @chrisccoulson should this already include get-deps and .travis.yaml ? a short README about installing the simulator snap and how to run the tests would also be good

Yes, it probably should. I forgot to add those.

errors.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"errors"
+	"fmt"
+)
+
+var (
+	// ErrProvisioning indicates that the TPM is not provisioned correctly for the requested operation. Please note that other errors
+	// that can be returned may also be caused by incomplete provisioning, as it is not always possible to detect incomplete or
+	// incorrect provisioning in all contexts.
+	ErrProvisioning = errors.New("the TPM is not correctly provisioned")
```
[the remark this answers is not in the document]
chrisccoulson: Yeah, that sounds ok

tpm.go
chrisccoulson: Yes, possibly - specifying the authorization value for permanent handles requires it (eg, see https://godoc.org/github.com/chrisccoulson/go-tpm2#TPMContext.OwnerHandleContext and https://godoc.org/github.com/chrisccoulson/go-tpm2#ResourceContext). Also, it's possible that some errors could require corrective action, such as undefining a NV index (using tpm2.TPMContext.NVUndefineSpace).

errors.go
chrisccoulson: Ack, that sounds ok.

tpm.go
chrisccoulson: I can probably remove that - examples/verify-tpm in the original repo does use it, but there probably isn't a reason for it to be there.

tpm.go
```diff
@@ -0,0 +1,937 @@
...
+	}
+
+NextCert:
+	for i := len(chain) - 1; i >= 0; i-- {
+		cert := chain[i]
+		if len(cert.ExtKeyUsage) == 0 && len(cert.UnknownExtKeyUsage) == 0 {
+			continue
+		}
+
+		for _, usage := range cert.ExtKeyUsage {
```
mvo5: (nitpick) with a small helper like `isExtKeyUsageAny()` this would probably be slightly more readable and avoids the continue to label.

tpm.go
```diff
@@ -0,0 +1,937 @@
...
+	return true
+}
+
+func parseTPMDeviceAttributesFromDirectoryName(dirName pkix.RDNSequence) (*TPMDeviceAttributes, pkix.RDNSequence, error) {
+	var attrs TPMDeviceAttributes
+	var rdnsOut pkix.RelativeDistinguishedNameSET
+
+	hasManufacturer, hasModel, hasVersion := false, false, false
+
+	for _, rdns := range dirName {
```
mvo5: Nice checking here!

tpm.go
```diff
@@ -0,0 +1,937 @@
...
+	if !ok {
+		return errors.New("cannot obtain RSA public key from certificate")
+	}
+
+	// Insert the RSA public key in to the EK template to compute the name of the EK object we expected to read back from the TPM.
+	var ekPublic *tpm2.Public
+	b, _ := tpm2.MarshalToBytes(ekTemplate)
+	tpm2.UnmarshalFromBytes(b, &ekPublic)
+
+	if pubKey.E != 65537 {
```
mvo5: Maybe a comment here similar to the one below in https://github.com/snapcore/secboot/pull/1/files#diff-7d1c2a3334601b6c1958aae0a594cba8R72

constants.go
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package secboot
+
+import (
+	"github.com/chrisccoulson/go-tpm2"
```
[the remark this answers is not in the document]
chrisccoulson: Not yet - it might be best to do one repo at a time, as I think things might get a bit messy already whilst we're copying stuff here, doing some cleanups and making other changes in the process, whilst still committing other changes to chrisccoulson/ubuntu-core-fde-utils at the same time.

tools/gen-certdata/gen-certdata.go
```diff
@@ -0,0 +1,90 @@
...
+	"crypto"
+	_ "crypto/sha256"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+)
+
+func main() {
```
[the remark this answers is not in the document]
chrisccoulson: Oops, deleted one of them :)

gen-certdata/gen-certdata.go
```diff
@@ -0,0 +1,90 @@
...
+
+	buffer.WriteString("package secboot\n\n")
+	buffer.WriteString("var (\n")
+	buffer.WriteString("\trootCAHashes = [][]byte{\n")
+
+	for _, fi := range files {
+		buffer.WriteString("\t\t[]byte{")
+		path := filepath.Join(in, fi.Name())
+		data := func() []byte {
+			f, err := os.Open(path)
```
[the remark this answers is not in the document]
chrisccoulson: Yes, that helps quite a bit.

tpm.go
```diff
@@ -0,0 +1,928 @@
...
+// which must contain a URL for the parent certificate.
+//
+// This function will stop when it encounters a certificate that doesn't specify an issuer URL, or when it encounters a self-signed
+// certificate.
+//
+// If no endorsement key certificate can be obtained, an error will be returned.
+//
+// If parentsOnly is true, this function will only save the parent certificates as long as the endorsement key certificate can be
+// reliably obtained from the TPM.
+func FetchAndSaveEKCertificateChain(tpm *TPMConnection, parentsOnly bool, destPath string) error {
```
pedronis: why fetching and saving together is more convenient it doesn't feel very orthogonal also it forces to save to a file. It seems a form that is fetch only and returns (ekCert *x509.Certificate, parents []*x509.Certificate) would be more general and can be combined with either Encode or Save as useful.  A form that only does the readEkCertFromTPM might also be useful for flexibility?

chrisccoulson: The reason this is here is because saving them to a file will be the most common way of storing them, but it's probably worth having a FetchEKCertificateChain which returns (*x509.Certificate, []*x509.Certificate, error) as well. I only really added EncodeEKCertificateChain originally because it allows the tests to use SecureConnectToDefaultTPM without having to save a file.

tpm.go
chrisccoulson: Thanks, although this reminds me - some of this code and the code involved with certificate chain verification could probably do with some additional unit tests (it does get tested at a high level because the tests certify the simulator with a fake certificate).

tpm.go
chrisccoulson: I've added a brief comment here now.

tpm.go
chrisccoulson: Yeah, that makes it a bit neater - I've added isExtKeyUsageAny and isExtKeyUsageEkCertificate as well.

export_test.go
```diff
@@ -0,0 +1,51 @@
...
+	EkTemplate                     = ekTemplate
+	OidExtensionSubjectAltName     = oidExtensionSubjectAltName
+	OidTcgAttributeTpmManufacturer = oidTcgAttributeTpmManufacturer
+	OidTcgAttributeTpmModel        = oidTcgAttributeTpmModel
+	OidTcgAttributeTpmVersion      = oidTcgAttributeTpmVersion
+	OidTcgKpEkCertificate          = oidTcgKpEkCertificate
+)
+
+func SetOpenDefaultTctiFn(fn func() (io.ReadWriteCloser, error)) {
+	openDefaultTcti = fn
```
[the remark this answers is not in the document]
chrisccoulson: Thanks - I've updated it to use a similar pattern.

tools/gen-certdata/gen-certdata.go
```diff
@@ -0,0 +1,90 @@
...
+
+	buffer.WriteString("package fdeutil\n\n")
+	buffer.WriteString("var (\n")
+	buffer.WriteString("\trootCAHashes = [][]byte{\n")
+
+	for _, fi := range files {
+		buffer.WriteString("\t\t[]byte{")
+		path := filepath.Join(in, fi.Name())
+		data := func() []byte {
+			f, err := os.Open(path)
```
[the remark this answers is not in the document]
chrisccoulson: Ah, this one is a commenting on the same code in a different file. I've deleted the duplicate file now.

tpm.go
pedronis: yes, let's reconsider adding that in a follow up

PR #2, opened 2020-02-13
Add ProvisionTPM, RequestTPMClearUsingPPI and ProvisionStatus

chrisccoulson:


provisioning.go
```diff
@@ -0,0 +1,370 @@
...
+// hierarchy. If called with mode set to ProvisionModeClear, or if called just after clearing the TPM via the physical presence
+// interface, the authorization value for the storage hierarchy will be empty at the point that it is required. If called with any
+// other mode and if the authorization value for the storage hierarchy has previously been set, it will need to be provided by calling
+// TPMConnection.OwnerHandleContext().SetAuthValue() prior to calling this function. If the wrong value is provided for the storage
+// hierarchy authorization, then a AuthFailError error will be returned. If the correct authorization value is not known, then the
+// only way to recover from this is to call the function with mode set to ProvisionModeClear. If there is an object already stored at
+// the location used for the storage root key then this function will evict it automatically from the TPM.
+//
+// This function will create a pair of NV indices used for locking access to sealed key objects, if necessary. These indices will be
+// created at 0x01801100 and 0x01801101. This requires knowledge of the authorization value for the storage hierarchy. If called with
```
pedronis: I suppose these are referred by hex notation because the user of the library never needs to specify them ?

provisioning.go
```diff
@@ -0,0 +1,370 @@
...
+// clear the TPM via the physical presence interface.
+//
+// If mode is ProvisionModeClear, this function will attempt to clear the TPM before provisioning it. If owner clear has been disabled
+// (which will be the case if the TPM has previously been provisioned with this function), then ErrTPMClearRequiresPPI will be returned.
+// In this case, the TPM must be cleared via the physical presence interface by calling RequestTPMClearUsingPPI and performing a
+// system restart.
+//
+// This function will create and persist an endorsement key which requires knowledge of the authorization values for the storage
+// and endorsement hierarchies. If called with mode set to ProvisionModeClear, or if called just after clearing the TPM via the
+// physical presence interface, the authorization values for these hierarchies will be empty at the point that they are required.
```
pedronis: is this a TPM termology thing to use empty vs unset,  is it actually conceptual?  (I'm asking because unset would be a bit natural to me)

provisioning.go
```diff
@@ -0,0 +1,370 @@
...
+// indices need to be created, then the only way to recover from this is to call the function with mode set to ProvisionModeClear. If
+// there are already NV indices defined at either of these handles but they don't meet the requirements of this function, a
+// TPMResourceExistsError error will be returned. In this case, the caller will either need to manually undefined these using
+// TPMContext.NVUndefineSpace, or clear the TPM.
+//
+// If mode is not ProvisionModeWithoutLockout, the authorization value for the lockout hierarchy will be set to newLockoutAuth.
+func ProvisionTPM(tpm *TPMConnection, mode ProvisionMode, newLockoutAuth []byte) error {
+	status, err := ProvisionStatus(tpm)
+	if err != nil {
+		return xerrors.Errorf("cannot determine the current status: %w", err)
```
pedronis: s/status/TPM status/ ?

provisioning.go
```diff
@@ -0,0 +1,370 @@
...
+// value for the storage hierarchy will be empty at the point that it is required. If called with any other mode and if the
+// authorization value for the storage hierarchy has previously been set, it will need to be provided by calling
+// TPMConnection.OwnerHandleContext().SetAuthValue() prior to calling this function. If the wrong value is provided for the storage
+// hierarchy authorization, then a AuthFailError error will be returned. If the correct authorization value is not known and new NV
+// indices need to be created, then the only way to recover from this is to call the function with mode set to ProvisionModeClear. If
+// there are already NV indices defined at either of these handles but they don't meet the requirements of this function, a
+// TPMResourceExistsError error will be returned. In this case, the caller will either need to manually undefined these using
+// TPMContext.NVUndefineSpace, or clear the TPM.
+//
+// If mode is not ProvisionModeWithoutLockout, the authorization value for the lockout hierarchy will be set to newLockoutAuth.
```
pedronis: it is in parenthetical earlier but probably this should repeat that we will also disable owner clear.

provisioning.go
```diff
@@ -0,0 +1,370 @@
...
+				Scheme:   tpm2.RSAScheme{Scheme: tpm2.RSASchemeNull},
+				KeyBits:  2048,
+				Exponent: 0}},
+		Unique: tpm2.PublicIDU{Data: make(tpm2.PublicKeyRSA, 256)}}
+)
+
+// ProvisionTPM prepares the TPM associated with the tpm parameter for full disk encryption. The mode parameter specifies the
+// behaviour of this function.
+//
+// If mode is not ProvisionModeWithoutLockout, this function performs operations that require knowledge of the lockout hierarchy
```
pedronis: I would move this later together with the explanation of what we try to do in ProvisionModeWithoutLockout.

provisioning.go
```diff
@@ -0,0 +1,370 @@
...
+// the TPM will have entered dictionary attack lockout mode for the lockout hierarchy. Further calls will result in a ErrTPMLockout
+// error being returned. The only way to recover from this is to either wait for the pre-programmed recovery time to expire, or to
+// clear the TPM via the physical presence interface.
+//
+// If mode is ProvisionModeClear, this function will attempt to clear the TPM before provisioning it. If owner clear has been disabled
+// (which will be the case if the TPM has previously been provisioned with this function), then ErrTPMClearRequiresPPI will be returned.
+// In this case, the TPM must be cleared via the physical presence interface by calling RequestTPMClearUsingPPI and performing a
+// system restart.
+//
+// This function will create and persist an endorsement key which requires knowledge of the authorization values for the storage
```
pedronis: In all modes (list of modes) ... maybe ?

provisioning.go
```diff
@@ -0,0 +1,370 @@
...
+// and endorsement hierarchies. If called with mode set to ProvisionModeClear, or if called just after clearing the TPM via the
+// physical presence interface, the authorization values for these hierarchies will be empty at the point that they are required.
+// If called with any other mode and if the authorization values have previously been set, they will need to be provided by calling
+// TPMConnection.EndorsementHandleContext().SetAuthValue() and TPMConnection.OwnerHandleContext().SetAuthValue() prior to calling
+// this function. If the wrong value is provided for either authorization, then a AuthFailError error will be returned. If the
+// correct authorization values are not known, then the only way to recover from this is to call the function with mode set to
+// ProvisionModeClear. If there is an object already stored at the location used for the endorsement key then this function will
+// evict it automatically from the TPM.
+//
+// This function will create and persist a storage root key, which requires knowledge of the authorization value for the storage
```
pedronis: will create as well

provisioning.go
```diff
@@ -0,0 +1,370 @@
...
+// hierarchy. If called with mode set to ProvisionModeClear, or if called just after clearing the TPM via the physical presence
+// interface, the authorization value for the storage hierarchy will be empty at the point that it is required. If called with any
+// other mode and if the authorization value for the storage hierarchy has previously been set, it will need to be provided by calling
+// TPMConnection.OwnerHandleContext().SetAuthValue() prior to calling this function. If the wrong value is provided for the storage
+// hierarchy authorization, then a AuthFailError error will be returned. If the correct authorization value is not known, then the
+// only way to recover from this is to call the function with mode set to ProvisionModeClear. If there is an object already stored at
+// the location used for the storage root key then this function will evict it automatically from the TPM.
+//
+// This function will create a pair of NV indices used for locking access to sealed key objects, if necessary. These indices will be
```
pedronis: further somewhere in here

errors.go
```diff
@@ -22,15 +22,51 @@ package secboot
...
 
+
+// TPMResourceExistsError is returned from any function that creates a persistent TPM resource if a resource already exists
+// at the specified handle.
+type TPMResourceExistsError struct {
+	Handle tpm2.Handle
+}
+
+func (e TPMResourceExistsError) Error() string {
+	return fmt.Sprintf("a resource already exists on the TPM at handle 0x%08x", e.Handle)
```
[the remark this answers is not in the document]
chrisccoulson: You're right - I'd completely forgotten about that.

errors.go
```diff
@@ -22,15 +22,51 @@ package secboot
...
+// for other reasons that would cause a HMAC check failure, such as a communication failure between the host CPU and the TPM
+// or the name of a resource on the TPM not matching the name of the ResourceContext passed to the function that failed - this
+// latter issue can occur when using a resource manager if another process accesses the TPM and makes changes to persistent
+// resources or sessions.
+type AuthFailError struct {
+	Handle tpm2.Handle
+}
+
+func (e AuthFailError) Error() string {
+	return fmt.Sprintf("an authorization check failed for the hierarchy associated with %v", e.Handle)
```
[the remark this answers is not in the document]
chrisccoulson: That could possibly work, but I should actually check that "hierarchy" is appropriate here. That was the case when the error was first added, but it might not be now.

provisioning.go
```diff
@@ -0,0 +1,370 @@
...
+	// Presence Interface Specification", version 1.30, revision 00.52, 28 July 2015.
+	clearPPIRequest string = "5"
+
+	// DA lockout parameters.
+	maxTries        uint32 = 32
+	recoveryTime    uint32 = 7200
+	lockoutRecovery uint32 = 86400
+)
+
+const (
```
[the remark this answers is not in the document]
chrisccoulson: Sure, I can move those.

provisioning.go
```diff
@@ -0,0 +1,370 @@
...
+
+	AttrLockNVIndex // The TPM has a valid NV index used for locking access to keys sealed with SealKeyToTPM
+)
+
+const (
+	// ProvisionModeClear specifies that the TPM should be fully provisioned after clearing it.
+	ProvisionModeClear ProvisionMode = iota
+
+	// ProvisionModeWithoutLockout specifies that the TPM should be refreshed without performing
+	// operations that require knowledge of the lockout hierarchy authorization value.
```
pedronis: this should probably expand a bit on what those operations are,  if I understand, they are  disabling owner clear and configuring lockout itself ?

provisioning.go
chrisccoulson: The TCG specifications refer to an unset authorization value as empty.

provisioning.go
chrisccoulson: That sounds ok, although I will probably remove this call to ProvisionStatus in a followup as it isn't really necessary here.

policy.go
```diff
@@ -0,0 +1,309 @@
...
+
+	// Initialize the index
+	if err := tpm.NVWrite(index, index, nil, 0, policySession, session.IncludeAttrs(tpm2.AttrAudit)); err != nil {
+		return xerrors.Errorf("cannot initialize NV index: %w", err)
+	}
+
+	// Marshal key name and cut-off time for writing to the NV index so that they can be used for verification in the future.
+	data, err := tpm2.MarshalToBytes(lockNVIndexVersion, keyName, time.ClockInfo.Clock)
+	if err != nil {
+		panic(fmt.Sprintf("cannot marshal contents for policy data NV index: %v", err))
```
mvo5: Could this be an errors as well? panic in the library code is a bit risky

policy.go
```diff
@@ -0,0 +1,309 @@
...
+	index, err := tpm.NVDefineSpace(tpm.OwnerHandleContext(), nil, public, session)
+	if err != nil {
+		if isNVIndexDefinedError(err) {
+			return &nvIndexDefinedError{public.Index}
+		}
+		return xerrors.Errorf("cannot create NV index: %w", err)
+	}
+
+	succeeded := false
+	defer func() {
```
mvo5: (nitpick) Could you check for `if err == nil { return }` here too?

provisioning.go
```diff
@@ -0,0 +1,370 @@
...
+	// Provision an endorsement key
+	ekContext, err := tpm.CreateResourceContextFromTPM(ekHandle)
+	if err == nil {
+		if _, err := tpm.EvictControl(tpm.OwnerHandleContext(), ekContext, ekHandle, session); err != nil {
+			if isAuthFailError(err) {
+				return AuthFailError{tpm2.HandleOwner}
+			}
+			return xerrors.Errorf("cannot evict existing object at handle required by endorsement key: %w", err)
+		}
+	} else if _, notFound := err.(tpm2.ResourceUnavailableError); !notFound {
```
mvo5: (nitpick) this feels a bit unusual, most go code hanldes the `err != nil` case first which is I think would also make this easier to read. Something like:
```
        if err != nil {
		if _, notFound := err.(tpm2.ResourceUnavailableError); !notFound {
			return xerrors.Errorf("cannot create context for object at handle required by endorsement key: %w", err)
		}
	}
	if _, err := tpm.EvictControl(tpm.OwnerHandleContext(), ekContext, ekHandle, session); err != nil {
		if isAuthFailError(err) {
			return AuthFailError{tpm2.HandleOwner}
		}
		return xerrors.Errorf("cannot evict existing object at handle required by endorsement key: %w", err)
	}
```
(or even the first line with a helper: `if err != nil && !tpm2.IsNotFound(err) {` (iirc xerrors also provides something for this).

provisioning.go
```diff
@@ -0,0 +1,370 @@
...
+func ProvisionStatus(tpm *TPMConnection) (ProvisionStatusAttributes, error) {
+	var out ProvisionStatusAttributes
+
+	session := tpm.HmacSession().IncludeAttrs(tpm2.AttrAudit)
+
+	if ek, err := tpm.CreateResourceContextFromTPM(ekHandle, session); err != nil {
+		if _, unavail := err.(tpm2.ResourceUnavailableError); !unavail {
+			return 0, err
+		}
+	} else if ekInit, err := tpm.EndorsementKey(); err == nil && bytes.Equal(ekInit.Name(), ek.Name()) {
```
mvo5: I'm not a fan of these nested if/else as I feel it requires to keep a lot of state in the head. I wonder if it could be written more explicit. Something like (please double check my logic!):
```
func isResourceUnavailable(err error) bool {
	_, unavail := err.(tpm2.ResourceUnavailableError)
	return unavail
}
...
        ek, err := tpm.CreateResourceContextFromTPM(ekHandle, session)
	if err != nil && !isResourceUnavailable(err) {
		return 0, err
	}
	if isResourceUnavailable(err) {
		ekInit, err := tpm.EndorsementKey()
		if err != nil {
			// yes?
			return err
		}
		if bytes.Equal(ekInit.Name(), ek.Name()) {
			out |= AttrValidEK
		}
        }
    ...
```

provisioning.go
```diff
@@ -0,0 +1,370 @@
...
+		out |= AttrValidEK
+	}
+
+	if srk, err := tpm.CreateResourceContextFromTPM(srkHandle, session); err != nil {
```
mvo5: Similar for these nested if/else

policy.go
pedronis: the target here is memory, the question is can there be untested bugs in MarshalToBytes that would trigger this vs runtime serious issues?  in any case if this is changed there is quite a lot of similar code in go-tpm2 as well that would need changing too

provisioning.go
```diff
@@ -0,0 +1,370 @@
...
+
+	if lockIndex, err := tpm.CreateResourceContextFromTPM(lockNVHandle, session); err != nil {
+		if _, unavail := err.(tpm2.ResourceUnavailableError); !unavail {
+			return 0, err
+		}
+	} else if _, err := readAndValidateLockNVIndexPublic(tpm.TPMContext, lockIndex, session); err == nil {
+		out |= AttrLockNVIndex
+	}
+
+	return out, nil
```
cmatsuoka: The logic in this block is a bit hard to follow, perhaps we could refactor it a little bit to make it easier for the casual reader to understand what's going on? Maybe something like the expanded version below could be more readable.
```
        if lockIndex, err := tpm.CreateResourceContextFromTPM(lockNVHandle, session); err != nil {
                if _, unavail := err.(tpm2.ResourceUnavailableError); !unavail {
                        return 0, err
                } else {
                        return out, nil
                }       
        }

        if _, err := readAndValidateLockNVIndexPublic(tpm.TPMContext, lockIndex, session); err != nil {
                return out, nil
        }

        return out | AttrLockNVIndex, nil
```

provisioning.go
chrisccoulson: I can reorder it, but the logic here wouldn't quite work right - TPMContext.EvictControl should only run if TPMContext.CreateResourceContextFromTPM succeeded (it will just return an error anyway), and this function should only return early with an error if TPMContext.CreateResourceContextFromTPM returns an error other than tpm2.ResourceUnavailableError.

provisioning.go
chrisccoulson: The issue is that whilst we want to return early with an error if TPMContext.CreateResourceContextFromTPM returns an error that isn't tpm2.ResourceUnavailableError, we have a subsequent block of code that should only run if no error was returned. So, it would look more like this:
```
	ek, err := tpm.CreateResourceContextFromTPM(ekHandle, session)
	if err != nil && !isResourceUnavailable(err) {
		return 0, err
	}
	if err == nil {
		ekInit, err := tpm.EndorsementKey()
		if err == nil && bytes.Equal(ekInit.Name(), ek.Name()) {
			out |= AttrValidEK
		}
        }
```
Is that better though?

mvo5: Aha, sorry, again sorry if I'm too much bikesheeding, I will try to avoid this, maybe one more suggestion?
```
ek, err := tpm.CreateResourceContextFromTPM(ekHandle, session)
	switch {
	case err != nil && !isResourceUnavailable(err):
		// pass real error up
		return 0, err
	case isResourceUnavailable(err):
		// nothing to do
	case err == nil:
		ekInit, err := tpm.EndorsementKey()
		if err == nil && bytes.Equal(ekInit.Name(), ek.Name()) {
			out |= AttrValidEK
		}
	}
```
I like the very explicit nature of this. But feel free to ignore at this point it's probably more about personal preference (of me) than real improvements.

provisioning.go
mvo5: Ok, sorry for that. Feel free to just leave as is if there is no better way to write this.

errors.go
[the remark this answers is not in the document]
chrisccoulson: Thinking about this - "cannot access hierarchy at handle %v" probably isn't specific enough - there are other reasons a hierarchy can be inaccessible other than an authorization failure (eg, they can be disabled by the owner). How about "cannot access resource at handle %v because an authorization check failed"?

provisioning.go
chrisccoulson: Yes, these are deliberately hardcoded so that they are the same on every device, and they aren't defined by any specs so there aren't any nice names for them.

provisioning.go
chrisccoulson: I ended up refactoring this bit of documentation quite a bit to try to factor in some of these comments but also reduce some of the duplication.

provisioning.go
chrisccoulson: Done

provisioning.go
chrisccoulson: There was quite a bit of duplicated code for creating the 2 primary keys, so I ended up splitting this out in to another helper (provisionPrimaryKey) and I've tried to clean it up a little.

errors.go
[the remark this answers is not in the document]
pedronis: the longer form is fine, and yes would be better to start with "cannot..."

PR #10, opened 2020-02-20
Fix TestProvisionWithInvalidEkCert failure

chrisccoulson:
--- FAIL: TestProvisionWithInvalidEkCert (4.06s)
provisioning_test.go:561: ProvisionTPM should have returned an error

The intention of the test is for it to temporarily modify the template used for
creating the endorsement key, but this fails in go 1.10 because the template
exported to the test is a copy of the one that ProvisionTPM uses.

### DIFF
--- a/export_test.go
+++ b/export_test.go
@@ -37,6 +37,7 @@ const (
 var (
 	EkTemplate                     = ekTemplate
 	LockNVIndexAttrs               = lockNVIndexAttrs
+	MakeDefaultEKTemplate	       = makeDefaultEKTemplate
 	OidExtensionSubjectAltName     = oidExtensionSubjectAltName
 	OidTcgAttributeTpmManufacturer = oidTcgAttributeTpmManufacturer
 	OidTcgAttributeTpmModel        = oidTcgAttributeTpmModel
--- a/export_test.go
+++ b/export_test.go
@@ -33,7 +33,7 @@ const (
 )
 
 var (
-	EkTemplate                     = ekTemplate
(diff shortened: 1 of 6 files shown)


tpm_test.go
```diff
@@ -171,7 +171,7 @@ func createTestCA() ([]byte, crypto.PrivateKey, error) {
 }
 
 func createTestEkCert(tpm *tpm2.TPMContext, caCert []byte, caKey crypto.PrivateKey) ([]byte, error) {
-	ekContext, pub, _, _, _, err := tpm.CreatePrimary(tpm.EndorsementHandleContext(), nil, &EkTemplate, nil, nil, nil)
+	ekContext, pub, _, _, _, err := tpm.CreatePrimary(tpm.EndorsementHandleContext(), nil, EkTemplate, nil, nil, nil)
```
pedronis: I'm quite confused, do some/which of these functions modify the template in place? don't we have the same issue in the real code, that the template cananot be reused?

chrisccoulson: There aren't any other functions that modify the template.

export_test.go
```diff
@@ -58,3 +60,11 @@ func InitTPMConnection(t *TPMConnection) error {
 func AppendRootCAHash(h []byte) {
 	rootCAHashes = append(rootCAHashes, h)
 }
+
+func MockEKTemplate(mock *tpm2.Public) (func()) {
```
pedronis: naming the return (restore func()) is a bit more self-documenting

PR #11, opened 2020-03-02
Add computeSecureBootPolicyDigests

chrisccoulson:
This is used for computing secure boot policy PCR digests based on pending EFI
signature database updates and a list of EFI image load event sequences, which
can be used in a TPM authorization policy via the TPM2_PolicyPCR assertion and
is required in order to support atomic updates of components that affect the
secure boot policy.

Some things to note on this PR:
- It currently depends on my own fork of https://github.com/fullsailor/pkcs7 which includes a small bug fix for handling of Microsoft Authenticode signatures. Mozilla appear to have a fork of this repository too, and I'll investigate switching to that (and supply another PR): https://github.com/mozilla-services/pkcs7.
- There are some binaries required for testing purposes - currently these are just checked in. It might be worth investigate compiling these from source later on.
- On that note, some of the tests only work on x86-64 because they depend on mock EFI binaries that are only compiled for one architecture.
- Quite a lot of things need to be exported for testing of this code - I'll investigate refactoring this later on (maybe an internal package providing test utilities that can be shared between test code that's inside both the secboot and secboot_test packages).
- It contains a copy of some code from https://github.com/golang/go/commit/3b92f36d15c868e856be71c0fadfc7ff97039b96 in order to avoid a hard dependency on go 1.14 (see pe.go).
- The API is not exported, but some types are exported because they will be used by code that initiates the sealing (EFIImage, EFIImageLoadEvent, EFIImageLoadEventSource).


secureboot_policy.go
```diff
@@ -0,0 +1,1292 @@
...
+	Container snap.Container
+	Path      string // The path of the snap image (used by the implementation of fmt.Stringer)
+	FileName  string // The filename within the snap squashfs
+}
+
+func (f SnapFileEFIImage) String() string {
+	return f.Path + ":" + f.FileName
+}
+
+func (f SnapFileEFIImage) ReadAll() ([]byte, error) {
```
pedronis: I'm not sure I understand how this API will be used, won't []byte be quite large in some cases to just get all of it into memory?

secureboot_policy.go
```diff
@@ -0,0 +1,1292 @@
...
+
+	return certData, nil
+}
+
+// secureBootDbIterator provides a mechanism to iterate over a set of EFI_SIGNATURE_LIST entries in a EFI signature database.
+type secureBootDbIterator struct {
+	r io.ReadSeeker
+}
+
+// nextSignatureList returns the SignatureType, SignaureHeader and EFI_SIGNATURE_DATA entries associated with the next
```
pedronis: typo, SignatureHeader

secureboot_policy.go
pedronis: do we need to do something on the snapd side to help with this?

secureboot_policy.go
```diff
@@ -0,0 +1,1292 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package secboot
+
```
pedronis: this file feels a bit too  large, is there a way to split into lower level/higher level bits?  or spit out the low level efi bits and the signature db bits ?

secureboot_policy.go
pedronis: snapd never implemented anything but ReadFile on Container because we didn't have large file use cases to consider

chrisccoulson: ReadAll() is certainly suboptimal - ideally, we'd have access to a io.ReaderAt to avoid having to read the entire file in to memory (which, in the case of the kernel image is over 10MB)

secureboot_policy.go
chrisccoulson: I think what I'd like to do in a future PR is split some of this code out in to an internal package - that will make this file quite a bit smaller but also reduce the number of test-only exports that are defined in export_test.go.

secureboot_policy.go
pedronis: I looked, we should be able to add something like RandomAccessFile() ReadAtCloser to Container,  where ReadAtCloser is basically  io.ReaderAt + io.Closer. I'll try to have a PR up today, shouldn't block this PR though.

chrisccoulson: Yeah, that would work for this - thanks!

pedronis: I proposed this:

https://github.com/snapcore/snapd/pull/8248

export_test.go
```diff
@@ -35,48 +37,116 @@ const (
...
-
+// Alias some unexported types for testing. These are required in order to pass these between functions in tests, or to access
+// unexported members of some unexported types.
 type DynamicPolicyData dynamicPolicyData
 
-type MockPolicyPCRParam struct {
-	PCR     int
-	Alg     tpm2.HashAlgorithmId
-	Digests tpm2.DigestList
+type EFISignatureData efiSignatureData
```
pedronis: a type alias woudl be:

type EFISignatureData = efiSignatureData

pe.go
```diff
@@ -0,0 +1,197 @@
+package secboot
+
```
pedronis: this code doesn't seem to be tested ATM

cmatsuoka:
snapcore/snapd#8248 has landed.

export_test.go
chrisccoulson: Aha, I didn't know there was a difference but I've just had a look at the documentation. I'll fix that and others in another PR.

export_test.go
```diff
@@ -24,8 +24,10 @@ import (
 	"io"
 
 	"github.com/chrisccoulson/go-tpm2"
+	"github.com/chrisccoulson/tcglog-parser"
```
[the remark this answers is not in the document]
chrisccoulson: This might be better under github.com/canonical, along with go-tpm2.

PR #12, opened 2020-03-03
Make policyPCRData serializable

chrisccoulson:
The policyPCRData type needs to be serializable but currently isn't because it contains a member of type int.


chrisccoulson:
Actually, closing this one as it's superseded by https://github.com/snapcore/secboot/pull/13, which removes this type entirely.

PR #13, opened 2020-03-04
Execute a single TPM2_PolicyPCR command regardless of the number of PCRs

chrisccoulson:
This will make it easier to generate authorization policies with more restrictive combinations of PCR values.

### DIFF
--- a/policy.go
+++ b/policy.go
@@ -518,7 +518,23 @@ func computeDynamicPolicy(alg tpm2.HashAlgorithmId, input *dynamicPolicyComputeP
 	var pcrOrDigests tpm2.DigestList
 	// Keep track of the index in to each policyPcrParam.digests slice without having to do a recursive iteration.
 	indices := make([]int, len(input.pcrParams))
-Outer:
+	advanceIndices := func() bool {
+		for i := range input.pcrParams {
+			indices[i]++
+			if indices[i] < len(input.pcrParams[i].digests) {
+				break
+			}
+			// The index for this policyPcrParam has advanced beyond the end of its digests slice - roll over to zero before moving on to
+			// advance the index for the next policyPcrParam.
+			indices[i] = 0
+			if i == len(input.pcrParams)-1 {
+				// We've rolled all of the indices back to zero, so we're finished now
+				return false
+			}
+		}
+		return true
+	}
+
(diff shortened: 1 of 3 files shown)


policy.go
```diff
@@ -508,28 +488,75 @@ func computeStaticPolicy(alg tpm2.HashAlgorithmId, input *staticPolicyComputePar
...
+			indices[i]++
+			if indices[i] < len(input.pcrParams[i].digests) {
+				break
+			}
+			// The index for this policyPcrParam has advanced beyond the end of its digests slice - roll over to zero before moving on to
+			// advance the index for the next policyPcrParam
+			indices[i] = 0
+			if i == len(input.pcrParams)-1 {
+				// We've rolled all of the indices back to zero, so we're finished now
+				break Outer
```
mvo5: (nitpick) I wonder if this would be easier to read if this would get extracted into a helper and then it could return here instead of breaking out to "outer". But just a hunch

policy_test.go
```diff
@@ -720,34 +773,61 @@ func TestComputeDynamicPolicy(t *testing.T) {
...
+				}
+			}
+			numPcrOrDigests := 0
+			for _, p := range data.pcrParams {
+				if numPcrOrDigests == 0 {
+					numPcrOrDigests = 1
 				}
+				numPcrOrDigests *= len(p.Digests)
 			}
+			//if len(dataout.PCROrDigests) != numPcrOrDigests {
```
mvo5: Why is this commented out? Should there be a comment about the why?

policy_test.go
```diff
@@ -917,6 +998,7 @@ func TestExecutePolicy(t *testing.T) {
 	})
 
 	t.Run("Single/2", func(t *testing.T) {
+		// Test with a policy that includes a single digest for 2 PCRs
```
mvo5: Thanks for adding those comments

policy_test.go
chrisccoulson: It should have been deleted - I've got another branch that makes additional changes to this code and removes this commented out block already.

policy.go
chrisccoulson: I've updated it to use a helper function now.

PR #16, opened 2020-03-12
Support compound authorization policies with more than 8 conditions

chrisccoulson:
TPM2_PolicyOR only supports 8 conditions, but merging all TPM2_PolicyPCR assertions in to a single assertion makes it more likely that we may generate compound authorization policies with more than 8 conditions in some more complex transient cases. This adds support for more than 8 conditions.

### DIFF
--- a/policy.go
+++ b/policy.go
@@ -64,9 +64,16 @@ type dynamicPolicyComputeParams struct {
 	policyCount uint64
 }
 
+type policyOrDataNode struct {
+	Next    uint32
+	Digests tpm2.DigestList
+}
+
+type policyOrDataTree []policyOrDataNode
+
 type dynamicPolicyData struct {
 	PCRSelection              tpm2.PCRSelectionList
-	PCROrDigests              tpm2.DigestList
+	PCROrData                 policyOrDataTree
 	PolicyCount               uint64
 	AuthorizedPolicy          tpm2.Digest
 	AuthorizedPolicySignature *tpm2.Signature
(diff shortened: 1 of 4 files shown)


policy.go
```diff
@@ -64,9 +64,16 @@ type dynamicPolicyComputeParams struct {
 	policyCount uint64
 }
 
+type policyOrDataNode struct {
+	Next    uint32
```
pedronis: this probably merits a comment about how it's defined, is it absolute or relative to the current node?

policy.go
```diff
@@ -485,6 +492,61 @@ func computeStaticPolicy(alg tpm2.HashAlgorithmId, input *staticPolicyComputePar
...
+		nextDigests = append(nextDigests, trial.GetDigest())
+
+		// We've consumed n digests, so adjust the slice to point to the next ones to consume to produce a sibling node.
+		digests = digests[n:]
+
+		if len(digests) == 0 {
+			// There are no digests left to produce sibling nodes, and we have a collection of digests to produce parent nodes. Update the
+			// data produced for the nodes at this level to point to the parent nodes we're going to produce on the subsequent iterations.
+			for i := range nextDigests {
+				data[current+i].Next = uint32(len(nextDigests) - i + (i / 8))
```
pedronis: related to the comment above,  the -i  here is not obvious over casual reading

policy.go
```diff
@@ -485,6 +492,61 @@ func computeStaticPolicy(alg tpm2.HashAlgorithmId, input *staticPolicyComputePar
...
+// for the final TPM2_PolicyOR assertion, and leaf nodes containing digests for each OR condition. Whilst the returned data is
+// conceptually a tree, the layout in memory is just a slice of tables of up to 8 digests, each with an index that enables the code
+// executing the assertions to traverse upwards through the tree by just advancing to another entry in the slice. This format is
+// easily serialized. After the computations are completed, the provided *tpm2.TrialAuthPolicy will be updated.
+//
+// The returned data is used by firstly finding the leaf node which contains the current session digest. Once this is found, a
+// TPM2_PolicyOR assertion is executed on the digests in that node, and then the tree is traversed upwards to the root node, executing
+// TPM2_PolicyOR assertions along the way - see executePolicyORAssertions.
+func computePolicyORData(alg tpm2.HashAlgorithmId, trial *tpm2.TrialAuthPolicy, digests tpm2.DigestList) policyOrDataTree {
+	var current int
```
pedronis: current is an index in data, it should be called curNode, also for clary I would define it after the tree and list, as:

curNode := 0

PR #17, opened 2020-03-16
Make it possible to generate more restrictive PCR policies

chrisccoulson:
Previously, if a policy contained multiple values for multiple PCRs,
all combinations of values would be considered valid. But the API for
generating PCR policies (PCRProtectionProfile) will be able to generate
policies that are more restrictive than this, so support it in the core
policy code.

As an example, say you want to seal a key against digest A for PCR7 and
digest B for PCR8, or digest C for PCR7 and digest D for PCR8, the logic
would look something like this:

(PCR7[A] && PCR8[B]) || (PCR7[C] && PCR8[D])

...this is a policy that has 2 conditions. The PCRProtectionProfile API
can create this, but prior to this commit we would actually generate a
policy that would work more like this:

(PCR7[A] || PCR7[C]) && (PCR8[B] || PCR8[D])

...and this policy has 4 conditions, including combinations that weren't
explicitly allowed.


chrisccoulson:
> the new code looks simpler, it's not obvious in the tests, as they were changed but no new test added(?) that the behavior has switched to OR of ANDs as described?

The tests haven't changed much because I've added a helper (MakeMockPolicyPCRValues) which works like computeDynamicPolicy used to - it takes a list of PCR digests for each individual PCR and then creates a set of PCR values for each combination of condition. I could probably add some other test cases.

pedronis:
>  I could probably add some other test cases.

yes, it's probably good to exercise the new behavior in some way

PR #21, opened 2020-03-17
Add PCRProtectionProfile API

chrisccoulson:
This adds a low-level API for building arbitrary PCR protection policies, and
will be used by the key sealing code. The API supports building policies by
specifying static PCR values, specifying current PCR values or by computing PCR
values by performing a sequence of extend operations. Also supported are
policies with multiple conditions.

This also adds another public API - AddEFISecureBootPolicyProfile which allows
a key to be sealed to a device's EFI secure boot configuration and chain of
trust. In the future, additional mechanisms will be added to seal a key to
different properties - such as the kernel commandline when using systemd's EFI
stub, an Ubuntu Core model, boot manager code (PCR4) etc.


pcr_profile.go
```diff
@@ -0,0 +1,162 @@
...
+func (l pcrValuesList) setValue(alg tpm2.HashAlgorithmId, pcr int, value tpm2.Digest) {
+	for _, v := range l {
+		v.SetValue(pcr, alg, value)
+	}
+}
+
+// extendValue extends the specified PCR with the supplied value for all branches.
+func (l pcrValuesList) extendValue(alg tpm2.HashAlgorithmId, pcr int, value tpm2.Digest) {
+	for _, v := range l {
+		if _, ok := v[alg]; !ok {
```
pedronis: the two if feel like the should be a helper on PCRValues,  not a fan of the partial encapsulation going here

pcr_profile.go
```diff
@@ -0,0 +1,162 @@
...
+		h.Write(v[alg][pcr])
+		h.Write(value)
+		v[alg][pcr] = h.Sum(nil)
+	}
+}
+
+func (l pcrValuesList) copy() (out pcrValuesList) {
+	for _, v := range l {
+		ov := make(tpm2.PCRValues)
+		for alg := range v {
```
pedronis: again it feels like PCValues should have a Copy

pcr_profile.go
```diff
@@ -0,0 +1,162 @@
...
+
+// pcrProtectionProfileInstr is a building block of PCRProtectionProfile.
+type pcrProtectionProfileInstr struct {
+	op   pcrProfileOp
+	args []interface{}
+}
+
+// PCRProtectionProfile defines the PCR profile used to protect a key sealed with SealKeyToTPM. It contains a sequence of instructions
+// for computing combinations of PCR values that a key will be protected against. The profile is built using the methods of this type.
+type PCRProtectionProfile []pcrProtectionProfileInstr
```
pedronis: I find the naked unexported slice confusing, given how slice works (they are not lisp  cons) you cannot really build trees with them so the sliceness is not very useful, I think this should be wrapped in a struct, and the method below become mutating-in-place

pcr_profile.go
```diff
@@ -0,0 +1,162 @@
...
+	pcrProfileOpAdd pcrProfileOp = iota
+	pcrProfileOpAddFromTPM
+	pcrProfileOpExtend
+	pcrProfileOpOr
+)
+
+// pcrProtectionProfileInstr is a building block of PCRProtectionProfile.
+type pcrProtectionProfileInstr struct {
+	op   pcrProfileOp
+	args []interface{}
```
pedronis: this feels weak,  if we don't want to use a lot of structs and interfaces, then have fields for the various kind of possible argument and fill only the relevant ones

secureboot_policy.go
```diff
@@ -655,13 +655,14 @@ func isShimExecutable(r io.ReaderAt) (bool, error) {
 	return pefile.Section(".vendor_cert") != nil, nil
 }
 
-// secureBootProtectionParams provide the arguments to computeSecureBootPolicyDigests.
-type secureBootProtectionParams struct {
-	loadSequences []*EFIImageLoadEvent // A list of EFI image load sequences for which to compute PCR digests for
+// EFISecureBootPolicyProfileParams provide the arguments to AddEFISecureBootPolicyProfile.
+type EFISecureBootPolicyProfileParams struct {
+	PCRAlgorithm  tpm2.HashAlgorithmId
+	LoadSequences []*EFIImageLoadEvent // A list of EFI image load sequences for which to compute PCR digests for
```
pedronis: now that this is exported, it would be better documented in the style of:

// LoadSequences is a list ...

before the def

secureboot_policy.go
```diff
@@ -1051,7 +1051,7 @@ func (g *secureBootPolicyGen) processPreOSEvents(path *secureBootPolicyGenPath,
 	}
 
 	// The verification event associated with the initial OS load event was recorded as part of a UEFI driver load, so we need to keep it.
-	g.extendFirmwareVerificationMeasurement(path, tpm2.Digest(initialOSVerificationEvent.event.Digests[tcglog.AlgorithmId(g.alg)]))
+	g.extendFirmwareVerificationMeasurement(path, tpm2.Digest(initialOSVerificationEvent.event.Digests[tcglog.AlgorithmId(g.params.PCRAlgorithm)]))
```
pedronis: this is not new code but I noticed the tests don't reach here

secureboot_policy.go
```diff
@@ -743,19 +744,18 @@ type loadEventAndPaths struct {
 
 // secureBootPolicyGen is the main structure involved with computing secure boot policy PCR digests.
 type secureBootPolicyGen struct {
-	alg    tpm2.HashAlgorithmId
-	params *secureBootProtectionParams
+	params *EFISecureBootPolicyProfileParams
```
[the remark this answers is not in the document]
chrisccoulson: Thanks, I've updated it to use an anonymous member. I don't think there will be any more fields here - in the original repo, this struct used to be much larger. But most of that state is in secureBootPolicyGenPath now.

secureboot_policy.go
chrisccoulson: I've updated this now, and added some documentation for PCRAlgorithm too.

PR #22, opened 2020-03-17
Add key sealing and unsealing APIs

chrisccoulson:
This adds SealKeyToTPM for creating a new sealed TPM object,
UpdateKeyPCRProtectionPolicy for making changes to the PCR protection
policy for an existing key, and SealedKeyObject.UnsealFromTPM for accessing
a secret sealed with SealKeyToTPM.

### DIFF
--- a/keydata.go
+++ b/keydata.go
@@ -360,7 +360,7 @@ func readAndValidateKeyData(tpm *tpm2.TPMContext, keyFile, keyPolicyUpdateFile i
 
 	pinNVPublic, err := validateKeyData(tpm, data, policyUpdateData, session)
 	if err != nil {
-		return nil, nil, nil, xerrors.Errorf("key data validation failed: %w", err)
+		return nil, nil, nil, xerrors.Errorf("cannot validate key data: %w", err)
 	}
 
 	return data, policyUpdateData, pinNVPublic, nil
--- a/keydata.go
+++ b/keydata.go
@@ -1,0 +1,402 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package secboot
+
+import (
+	"bytes"
+	"crypto"
+	"crypto/rsa"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"io"
+	"math/big"
+	"os"
+
+	"github.com/chrisccoulson/go-tpm2"
+	"github.com/snapcore/snapd/osutil"
+	"github.com/snapcore/snapd/osutil/sys"
+
+	"golang.org/x/xerrors"
+)
+
+const (
+	currentVersion            uint32 = 0
+	keyDataHeader             uint32 = 0x55534b24
+	keyPolicyUpdateDataHeader uint32 = 0x55534b50
+)
+
+// AuthMode corresponds to an authentication mechanism.
+type AuthMode uint8
+
+const (
+	AuthModeNone AuthMode = iota
+	AuthModePIN
+)
+
+// keyPolicyUpdateData corresponds to the private part of a sealed key object that is required in order to create new dynamic
+// authorization policies.
+type keyPolicyUpdateData struct {
+	Data struct {
+		AuthKey []byte
+	}
+	CreationData   *tpm2.CreationData
+	CreationTicket *tpm2.TkCreation
+}
+
+// keyData corresponds to the part of a sealed key object that contains the TPM sealed object and associated metadata required
+// for executing authorization policy assertions.
+type keyData struct {
+	KeyPrivate        tpm2.Private
+	KeyPublic         *tpm2.Public
+	AuthModeHint      AuthMode
+	StaticPolicyData  *staticPolicyData
+	DynamicPolicyData *dynamicPolicyData
+}
+
+// readKeyPolicyUpdateData deserializes keyPolicyUpdateData from the provided io.Reader.
+func readKeyPolicyUpdateData(buf io.Reader) (*keyPolicyUpdateData, error) {
+	var header uint32
+	var version uint32
+	if err := tpm2.UnmarshalFromReader(buf, &header, &version); err != nil {
+		return nil, xerrors.Errorf("cannot unmarshal header and version number: %w", err)
+	}
+
+	if header != keyPolicyUpdateDataHeader {
+		return nil, fmt.Errorf("unexpected header (%d)", header)
+	}
+	if version != currentVersion {
+		return nil, fmt.Errorf("unexpected version number (%d)", version)
+	}
+
+	var d keyPolicyUpdateData
+	if err := tpm2.UnmarshalFromReader(buf, &d); err != nil {
+		return nil, xerrors.Errorf("cannot unmarshal key data: %w", err)
+	}
+
+	return &d, nil
+}
+
+// write serializes keyPolicyUpdateData to the provided io.Writer.
+func (d *keyPolicyUpdateData) write(buf io.Writer) error {
+	return tpm2.MarshalToWriter(buf, keyPolicyUpdateDataHeader, currentVersion, d)
+}
+
+type keyFileError struct {
+	err error
+}
+
+func (e keyFileError) Error() string {
+	return e.err.Error()
+}
+
+func (e keyFileError) Unwrap() error {
+	return e.err
+}
+
+func isKeyFileError(err error) bool {
+	var e keyFileError
+	return xerrors.As(err, &e)
+}
+
+// readKeyData deserializes keyData from the provided io.Reader.
+func readKeyData(buf io.Reader) (*keyData, error) {
+	var header uint32
+	var version uint32
+	if err := tpm2.UnmarshalFromReader(buf, &header, &version); err != nil {
+		return nil, keyFileError{xerrors.Errorf("cannot unmarshal header and version number: %w", err)}
+	}
+
+	if header != keyDataHeader {
+		return nil, keyFileError{fmt.Errorf("unexpected header (%d)", header)}
+	}
+	if version != currentVersion {
+		return nil, keyFileError{fmt.Errorf("unexpected version number (%d)", version)}
+	}
+
+	var d keyData
+	if err := tpm2.UnmarshalFromReader(buf, &d); err != nil {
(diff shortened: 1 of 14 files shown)


keydata.go
```diff
@@ -0,0 +1,402 @@
...
+	// It's loaded ok, so we know that the private and public parts are consistent.
+	defer tpm.FlushContext(keyContext)
+
+	lockIndex, err := tpm.CreateResourceContextFromTPM(lockNVHandle)
+	if err != nil {
+		return nil, xerrors.Errorf("cannot create context for lock NV index: %v", err)
+	}
+	lockIndexPub, err := readAndValidateLockNVIndexPublic(tpm, lockIndex, session)
+	if err != nil {
+		return nil, xerrors.Errorf("cannot determine if NV index at 0x%08x is global lock index: %w", lockNVHandle, err)
```
pedronis: usual comment about Handles having %v

keydata.go
```diff
@@ -0,0 +1,402 @@
...
+	}
+	if lockIndexPub == nil {
+		return nil, xerrors.Errorf("NV index at 0x%08x is not a valid global lock index", lockNVHandle)
```
pedronis: same

keydata.go
```diff
@@ -0,0 +1,402 @@
...
+func (k *SealedKeyObject) AuthMode2F() AuthMode {
+	return k.data.AuthModeHint
+}
+
+// PINIndexHandle indicates the handle of the NV index used for PIN support for this sealed key object.
+func (k *SealedKeyObject) PINIndexHandle() tpm2.Handle {
+	return k.data.StaticPolicyData.PinIndexHandle
+}
+
+// LoadSealedKeyObject loads a sealed key data file created by SealKeyToTPM from the specified path. If the file cannot be opened,
```
pedronis: this actually loads data in the tpm only temporarily in validateKeyData?  should it be called ReadSealedObject instead ?

pin.go
```diff
@@ -58,7 +95,7 @@ func createPinNVIndex(tpm *tpm2.TPMContext, handle tpm2.Handle, updateKeyName tp
 
 	nameAlg := tpm2.HashAlgorithmSHA256
 
-	// The NV index requires 4 policies:
+	// The NV index requires 5 policies:
 	// - A signed policy for initialization, signed with a key that is discarded so that the index cannot be recreated.
 	// - A signed policy for updating the index to revoke old dynamic authorization policies.
 	// - A policy for updating the authorization value (PIN / passphrase).
```
pedronis: maybe the code below maybe should mark each section with a comment again repeating which policy it is about

policy_test.go
```diff
@@ -1554,7 +1555,7 @@ func TestExecutePolicy(t *testing.T) {
 					data:  "foo",
 				},
 			}}, nil)
-		if err == nil {
+		if IsStaticPolicyDataError(err) || !IsDynamicPolicyDataError(err) {
 			t.Fatalf("Expected an error")
```
pedronis: should the messages be changed accordingly:  expected a dynamic policy data error etc, also below

seal.go
```diff
@@ -0,0 +1,347 @@
...
+	if err != nil {
+		return nil, xerrors.Errorf("cannot compute dynamic authorization policy: %w", err)
+	}
+
+	return policyData, nil
+}
+
+// KeyCreationParams provides the arguments for SealKeyToTPM.
+type KeyCreationParams struct {
+	PinHandle tpm2.Handle // Handle at which to create a NV index for PIN support
```
pedronis: will this grow over time, it's a bit unclear why a struct?

should the pcrPolicy live in here?

seal.go
```diff
@@ -0,0 +1,347 @@
...
+// either file cannot be created and opened for writing.
+//
+// The caller is expected to provide a handle at which a NV index should be created via the PinHandle filed of the KeyCreationParams
+// struct. If the handle is already in use, a TPMResourceExistsError error will be returned. The handle must be a valid NV index
+// handle (MSO == 0x01), and the choice of handle should take in to consideration the reserved indices from the "Registry of reserved
+// TPM 2.0 handles and localities" specification. It is recommended that the handle is in the block reserved for owner objects
+// (0x01800000 - 0x01bfffff).
+//
+// The key will be protected with a PCR policy computed from the supplied PCRProtectionProfile.
+func SealKeyToTPM(tpm *TPMConnection, keyPath, policyUpdatePath string, params *KeyCreationParams, pcrProfile *PCRProtectionProfile, key []byte) error {
```
pedronis: I would expect key []byte to be the second argument, it's conceptually the main input

seal.go
```diff
@@ -0,0 +1,347 @@
...
+		key:                  authPublicKey,
+		pinIndexPub:          pinIndexPub,
+		pinIndexAuthPolicies: pinIndexAuthPolicies,
+		lockIndexName:        lockIndexName})
+	if err != nil {
+		return xerrors.Errorf("cannot compute static authorization policy: %w", err)
+	}
+
+	// Define the template for the sealed key object, using the computed policy digest
+	template := tpm2.Public{
```
pedronis: maybe this should be global,  and used to get the relevant values in the validation code?

seal.go
```diff
@@ -0,0 +1,347 @@
...
+	case isAuthFailError(err, tpm2.CommandNVDefineSpace, 1):
+		return AuthFailError{tpm2.HandleOwner}
+	case err != nil:
+		return xerrors.Errorf("cannot create new pin NV index: %w", err)
+	}
+	defer func() {
+		if succeeded {
+			return
+		}
+		index, err := tpm2.CreateNVIndexResourceContextFromPublic(pinIndexPub)
```
pedronis: this bit of undo code doesn't seem exercised by tests

policy.go
```diff
@@ -727,10 +791,18 @@ func executePolicySession(tpm *tpm2.TPMContext, policySession tpm2.SessionContex
 
 	authorizeTicket, err := tpm.VerifySignature(authorizeKey, h.Sum(nil), dynamicInput.AuthorizedPolicySignature)
 	if err != nil {
+		if tpm2.IsTPMParameterError(err, tpm2.AnyErrorCode, tpm2.CommandVerifySignature, 2) {
+			// dynamicInput.AuthorizedPolicySignature is invalid.
+			return dynamicPolicyDataError{errors.New("dynamic authorization policy signature verification failed")}
+		}
 		return xerrors.Errorf("dynamic authorization policy signature verification failed: %w", err)
```
pedronis: cannot verify ...

policy.go
```diff
@@ -727,10 +791,18 @@ func executePolicySession(tpm *tpm2.TPMContext, policySession tpm2.SessionContex
 
 	authorizeTicket, err := tpm.VerifySignature(authorizeKey, h.Sum(nil), dynamicInput.AuthorizedPolicySignature)
 	if err != nil {
+		if tpm2.IsTPMParameterError(err, tpm2.AnyErrorCode, tpm2.CommandVerifySignature, 2) {
+			// dynamicInput.AuthorizedPolicySignature is invalid.
+			return dynamicPolicyDataError{errors.New("dynamic authorization policy signature verification failed")}
```
pedronis: cannto verify ...

keydata.go
chrisccoulson: Thanks, fixed!

keydata.go
```diff
@@ -0,0 +1,402 @@
...
+		var err error
+		policyUpdateData, err = readKeyPolicyUpdateData(keyPolicyUpdateFile)
+		if err != nil {
+			return nil, nil, nil, xerrors.Errorf("cannot read dynamic policy authorization data: %w", err)
+		}
+	}
+
+	pinNVPublic, err := validateKeyData(tpm, data, policyUpdateData, session)
+	if err != nil {
+		return nil, nil, nil, xerrors.Errorf("key data validation failed: %w", err)
```
pedronis: cannot validate key data...

keydata.go
chrisccoulson: I've renamed this to ReadSealedKeyObject. This function actually doesn't load anything in to the TPM. The validateKeyData function is only called from functions that aren't called during in early boot, as a way to perform additional diagnostic checks.

pin.go
chrisccoulson: I've updated the comments and added comments to each subsequent block too.

seal.go
chrisccoulson: There's a good chance it will grow with additional features, such as additional 2nd factor mechanisms or if we add a network-bound element to unlocking. That said, moving pcrProfile here probably makes sense.

unseal.go
```diff
@@ -0,0 +1,159 @@
...
+// If access to sealed key objects created by this package is disallowed until the next TPM reset or TPM restart, then a
+// ErrSealedKeyAccessLocked error will be returned.
+//
+// If the authorization policy check fails during unsealing, then a InvalidKeyFileError error will be returned. Note that this
+// condition can also occur as the result of an incorrectly provisioned TPM, which will be detected during a subsequent call to
+// SealKeyToTPM.
+//
+// On success, the unsealed cleartext key is returned. If the lock argument is true, then subsequent access to all sealed key
+// objects created by this package will be denied until the next TPM reset or TPM restart.
+func (k *SealedKeyObject) UnsealFromTPM(tpm *TPMConnection, pin string, lock bool) ([]byte, error) {
```
pedronis: will this typically be used with lock=true?

pin and lock could go in some kind of UnsealOptions struct

seal.go
chrisccoulson: I've added a test to exercise this now (TestSealKeyToTPMErrorHandling/InvalidPCRProfile)

unseal.go
chrisccoulson: Yes, I imagine that it will be used with lock=true. There will also need to be a separate API for triggering the locking without unsealing too.

seal.go
chrisccoulson: Ok, I've added a PCRProfile field to KeyCreationParams.

seal.go
chrisccoulson: It doesn't really work as a global because the AuthPolicy field needs to be mutated here. I've added makeSealedKeyTemplate() though which returns a *tpm2.Public without the AuthPolicy field set.

unseal.go
chrisccoulson: Following the conversation on IRC, I've removed the lock option from this in favour of a separate API (which will come in a future PR).

seal.go
```diff
@@ -0,0 +1,370 @@
...
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/chrisccoulson/go-tpm2"
+
+	"golang.org/x/xerrors"
+)
+
+func makeSealedKeyTemplate() *tpm2.Public {
```
pedronis: I would move this to keydata.go,  another approach (but maybe overkill) would be to combine my proposal and this:

in keydata.go:

```
var sealedKeySkeleton = tpm2.Public{
		Type:    tpm2.ObjectTypeKeyedHash,
		NameAlg: tpm2.HashAlgorithmSHA256,
		Attrs:   tpm2.AttrFixedTPM | tpm2.AttrFixedParent,
		Params:  tpm2.PublicParamsU{Data: &tpm2.KeyedHashParams{Scheme: 
                 tpm2.KeyedHashScheme{Scheme: tpm2.KeyedHashSchemeNull}}}}
}
```
leave out Params if they are mutable, sealedKeySkeleton.Xyz can be used in lots of places

here in seal.go:

```
func makeSealedKeyTemplate(authPolicy) *tmp.Public {
     t := sealedKeySkeleton // copy
     t.AuthPolicy = authPolicy
     return &t
}
```

PR #24, opened 2020-03-19
Add AddSystemdEFIStubProfile API

chrisccoulson:
The AddSystemdEFIStubProfile API enables the creation of a PCR policy that
seals a key to a permitted set of kernel commandlines when booting a
system using systemd's EFI stub loader.

### DIFF
--- a/sdefistub_policy.go
+++ b/sdefistub_policy.go
@@ -1,0 +1,74 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package secboot
+
+import (
+	"bytes"
+	"errors"
+
+	"github.com/chrisccoulson/go-tpm2"
+	"github.com/chrisccoulson/tcglog-parser"
+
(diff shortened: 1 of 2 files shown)


sdefistub_policy.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// AddSystemdEFIStubProfile adds the systemd EFI linux loader stub profile to the PCR protection profile, in order to generate a
+// PCR policy that restricts access to a key to a defined set of kernel commandlines when booting a linux kernel using the systemd
+// EFI stub.
+//
+// The PCR index that the EFI stub measures the kernel commandline too can be specified via the PCRIndex field of params.
+//
+// The set of kernel commandlines to add to the PCRProtectionProfile is specified via the KernelCmdlines field of params.
+func AddSystemdEFIStubProfile(profile *PCRProtectionProfile, params *SystemdEFIStubProfileParams) error {
```
pedronis: I'm probably getting a bit confused with naming,  why is just Profile and PolicyProfile?

sdefistub_policy.go
```diff
@@ -0,0 +1,77 @@
...
+			return xerrors.Errorf("cannot encode kernel commandline event: %w", err)
+		}
+
+		h := params.PCRAlgorithm.NewHash()
+		buf.WriteTo(h)
+
+		subProfiles = append(subProfiles, NewPCRProtectionProfile().ExtendPCR(params.PCRAlgorithm, params.PCRIndex, h.Sum(nil)))
+	}
+
+	profile.AddProfileOR(subProfiles...)
```
pedronis: question: how will this combine later  with the model measurement?  we will want a policy like (cmdLineA OR cmdLineB ...) AND (model measure) ?

sdefistub_policy.go
chrisccoulson: AddEFISecureBootPolicyProfile is named as it is because it's a profile for the EFI Secure Boot Policy measurements (with the name being taken from the TCG spec).

sdefistub_policy.go
chrisccoulson: There will be another function to add that to the profile. If you do:
```
profile := NewPCRProtectionProfile()
AddSystemdEFIStubProfile(profile, ...)
AddSnapModelProfile(profile, ...)
```
... then it will create a policy as you describe ( (cmdlineA || cmdlineB) && model )

PR #25, opened 2020-03-19
Add LockAccessToSealedKeys API

chrisccoulson:
This API is called to prevent access to any keys sealed by this package
until the next TPM restart or TPM reset. It works with any key sealed by
this package, regardless of the PCR protection policy.

### DIFF
--- a/policy.go
+++ b/policy.go
@@ -836,17 +836,25 @@ func LockAccessToSealedKeys(tpm *TPMConnection) error {
 		return xerrors.Errorf("cannot obtain handles from TPM: %w", err)
 	}
 	if len(handles) == 0 || handles[0] != lockNVHandle {
-		// Not provisioned, so no keys to protect.
+		// Not provisioned, so no keys created by this package can be unsealed by this TPM
 		return nil
 	}
 	lock, err := tpm.CreateResourceContextFromTPM(lockNVHandle)
 	if err != nil {
 		return xerrors.Errorf("cannot obtain context for lock NV index: %w", err)
 	}
+	lockPublic, _, err := tpm.NVReadPublic(lock, session.IncludeAttrs(tpm2.AttrAudit))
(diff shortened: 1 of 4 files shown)


policy.go
```diff
@@ -822,11 +822,15 @@ func executePolicySession(tpm *tpm2.TPMContext, policySession tpm2.SessionContex
...
-// policy of sealed key objects created by this package contain a TPM2_PolicyNV assertion on this index, which will fail once the
-// read lock bit has been enabled.
-func lockAccessToSealedKeysUntilTPMReset(tpm *tpm2.TPMContext, session tpm2.SessionContext) error {
+// LockAccessToSealedKeys locks access to keys sealed by this package until the next TPM restart (equivalent to eg, system resume
+// from suspend-to-disk) or TPM reset (equivalent to booting after a system restart). This works for all keys sealed by this package
+// regardless of their PCR protection profile.
+//
+// On success, subsequent calls to SealedKeyObject.UnsealFromTPM will fail with a ErrSealedKeyAccessLocked error until the next TPM
+// restart or TPM reset.
+func LockAccessToSealedKeys(tpm *TPMConnection) error {
```
pedronis: the trivial case does not seem to be covered:

```
if len(handles) == 0 || handles[0] != lockNVHandle
```

also I'm not sure I understand the 2nd check,  will we get nil sometimes there? or something completely unrelated/different?

chrisccoulson: The second check is because of the way that TPM2_GetCapability works. From the documentation:

"The capability parameter determines the category of data returned. The property parameter selects the first value of the selected category to be returned. If there is no property that corresponds to the value of property, the next higher value is returned, if it exists"

So it could return a handle, but it might not necessarily be lockNVHandle.

But yeah, there should be a test for the case where the TPM isn't provisioned.

PR #34, opened 2020-03-26
Add ChangePIN API for setting a PIN

chrisccoulson:


pin.go
```diff
@@ -251,3 +252,63 @@ func performPinChange(tpm *tpm2.TPMContext, public *tpm2.NVPublic, authPolicies
...
+//
+// If the TPM's dictionary attack logic has been triggered, a ErrTPMLockout error will be returned.
+//
+// If the file at the specified path cannot be opened, then a wrapped *os.PathError error will be returned.
+//
+// If the supplied key data file fails validation checks, an InvalidKeyFileError error will be returned.
+//
+// If oldPIN is incorrect, then a ErrPINFail error will be returned and the TPM's dictionary attack counter will be incremented.
+func ChangePIN(tpm *TPMConnection, path string, oldPIN, newPIN string) error {
+	// Check if the TPM is in lockout mode
```
pedronis, line 267: it seems we have other places that do this check? worth a helper, maybe not yet?

pin.go
```diff
@@ -251,3 +252,63 @@ func performPinChange(tpm *tpm2.TPMContext, public *tpm2.NVPublic, authPolicies
...
+	if err != nil {
+		var kfErr keyFileError
+		if xerrors.As(err, &kfErr) {
+			return InvalidKeyFileError{err.Error()}
+		}
+		return xerrors.Errorf("cannot read and validate key data file: %w", err)
+	}
+
+	// Change the PIN
+	if err := performPinChange(tpm.TPMContext, pinIndexPublic, data.StaticPolicyData.PinIndexAuthPolicies, oldPIN, newPIN, tpm.HmacSession()); err != nil {
```
pedronis, line 295: this mutates tpm own state but not data, is that right?

pin.go
```diff
@@ -251,3 +252,63 @@ func performPinChange(tpm *tpm2.TPMContext, public *tpm2.NVPublic, authPolicies
...
+		if isAuthFailError(err, tpm2.CommandNVChangeAuth, 1) {
+			return ErrPINFail
+		}
+		return err
+	}
+
+	// Update the metadata and write a new key data file
```
pedronis, line 302: should we do the writing only if the hint has changed?

pin.go
chrisccoulson: I've added this check here for now as a stop-gap - I think I might remove both checks in a follow up in favour of just detecting the lockout error returned from the TPM instead.

pin.go
chrisccoulson: That's correct.

pin.go
chrisccoulson: That sounds ok to me.

PR #39, opened 2020-04-06
Add ActivateVolume helper APIs for activating LUKS volumes

chrisccoulson:
This adds 2 functions for activating LUKS volumes.

ActivateVolumeWithTPMSealedKey attempts to activate a volume using the TPM
protected secret. If a PIN is required to unseal the secret, it will be
requested using systemd-ask-password. Once unsealed, the secret will be
used to activate the LUKS volume using systemd-cryptsetup.

ActivateVolumeWithTPMSealedKey can fall back to activating with the
fallback recovery key if activating with the TPM protected secret fails.

ActivateVolumeWithRecoveryKey can be used to activate a volume with the
fallback recovery key directly without attempting first with the TPM
protected secret.

In both functions, the fallback recovery key is requested using
systemd-ask-password. When activation with the recovery key is successful,
the key is added to the user keyring in the kernel. The description of
the keyring entry will contain an indicator of the reason that the
fallback recovery key was used.

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -183,37 +183,38 @@ func decodeRecoveryKey(passphrase string) ([]byte, error) {
 	return key.Bytes(), nil
 }
 
-// RecoveryReason indicates the reason that a volume had to be activated with the fallback recovery key instead of the TPM sealed key.
-type RecoveryReason uint8
+// RecoveryKeyUsageReason indicates the reason that a volume had to be activated with the fallback recovery key instead of the TPM
+// sealed key.
+type RecoveryKeyUsageReason uint8
 
 const (
-	// RecoveryReasonUnexpectedError indicates that a volume had to be activated with the fallback recovery key because an unexpected
-	// error was encountered during activation with the TPM sealed key.
-	RecoveryReasonUnexpectedError RecoveryReason = iota + 1
+	// RecoveryKeyUsageReasonUnexpectedError indicates that a volume had to be activated with the fallback recovery key because an
+	// unexpected error was encountered during activation with the TPM sealed key.
+	RecoveryKeyUsageReasonUnexpectedError RecoveryKeyUsageReason = iota + 1
 
-	// RecoveryReasonRequested indicates that a volume was activated with the fallback recovery key via the ActivateVolumeWithRecoveryKey
-	// API.
-	RecoveryReasonRequested
+	// RecoveryKeyUsageReasonRequested indicates that a volume was activated with the fallback recovery key via the
+	// ActivateVolumeWithRecoveryKey API.
+	RecoveryKeyUsageReasonRequested
 
-	// RecoveryReasonTPMLockout indicates that a volume had to be activated with the fallback recovery key because the TPM is in
+	// RecoveryKeyUsageReasonTPMLockout indicates that a volume had to be activated with the fallback recovery key because the TPM is in
 	// dictionary attack lockout mode.
-	RecoveryReasonTPMLockout
+	RecoveryKeyUsageReasonTPMLockout
 
-	// RecoveryReasonTPMProvisioningError indicates that a volume had to be activated with the fallback recovery key because the TPM is
-	// not correctly provisioned.
-	RecoveryReasonTPMProvisioningError
+	// RecoveryKeyUsageReasonTPMProvisioningError indicates that a volume had to be activated with the fallback recovery key because the
+	// TPM is not correctly provisioned.
+	RecoveryKeyUsageReasonTPMProvisioningError
 
-	// RecoveryReasonInvalidKeyFile indicates that a volume had to be activated with the fallback recovery key because the TPM sealed key
-	// file is invalid. Note that attempts to resolve this by creating a new file with SealKeyToTPM may indicate that the TPM is also not
-	// correctly provisioned.
-	RecoveryReasonInvalidKeyFile
+	// RecoveryKeyUsageReasonInvalidKeyFile indicates that a volume had to be activated with the fallback recovery key because the TPM
+	// sealed key file is invalid. Note that attempts to resolve this by creating a new file with SealKeyToTPM may indicate that the TPM
+	// is also not correctly provisioned.
+	RecoveryKeyUsageReasonInvalidKeyFile
 
-	// RecoveryReasonPINFail indicates that a volume had to be activated with the fallback recovery key because the correct PIN was not
-	// provided.
-	RecoveryReasonPINFail
+	// RecoveryKeyUsageReasonPINFail indicates that a volume had to be activated with the fallback recovery key because the correct PIN
+	// was not provided.
+	RecoveryKeyUsageReasonPINFail
 )
 
-func activateWithRecoveryKey(volumeName, sourceDevicePath string, keyReader io.Reader, tries int, reason RecoveryReason, activateOptions []string) error {
+func activateWithRecoveryKey(volumeName, sourceDevicePath string, keyReader io.Reader, tries int, reason RecoveryKeyUsageReason, activateOptions []string) error {
 	if tries == 0 {
 		return errors.New("no recovery key tries permitted")
 	}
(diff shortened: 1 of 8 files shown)


crypt.go
```diff
@@ -0,0 +1,488 @@
...
+			reason = RecoveryReasonInvalidKeyFile
+		}
+		rErr := activateWithRecoveryKey(volumeName, sourceDevicePath, nil, options.RecoveryKeyTries, reason, activateOptions)
+		return rErr == nil, &ActivateWithTPMSealedKeyError{err, rErr}
+	}
+
+	return true, nil
+}
+
+type ActivateWithRecoveryKeyOptions struct {
```
pedronis, line 458: this needs doc comments as well.

crypt.go
```diff
@@ -0,0 +1,488 @@
...
+	if reader != nil {
+		scanner := bufio.NewScanner(reader)
+		switch {
+		case scanner.Scan():
+			return scanner.Text(), nil
+		case scanner.Err() != nil:
+			return "", xerrors.Errorf("cannot obtain %s from scanner: %w", description, scanner.Err())
+		}
+	}
+	return askPassword(sourceDevicePath, "Please enter the "+description+" for disk "+sourceDevicePath+":")
```
pedronis, line 159: the *Options should allow a way to override this function, something like:

```
AskPassword func(kind PasswordKind, sourceDevicePath string) (string, error)
```

where `PasswordKind`:

```
type PasswordKind int

const (
  PasswordKindPIN =  iota
  PasswordKindRecoveryKey
)
```

can be a follow up though

crypt.go
```diff
@@ -0,0 +1,488 @@
...
+	return key.Bytes(), nil
+}
+
+// RecoveryReason indicates the reason that a volume had to be activated with the fallback recovery key instead of the TPM sealed key.
+type RecoveryReason uint8
+
+const (
+	// RecoveryReasonUnexpectedError indicates that a volume had to be activated with the fallback recovery key because an unexpected
+	// error was encountered during activation with the TPM sealed key.
+	RecoveryReasonUnexpectedError RecoveryReason = iota + 1
```
pedronis, line 192: using just the Recovery prefix when recovery key is meant is slightly confusing, given we have many "recovery" concepts around already (systems, partitions etc)

crypt.go
chrisccoulson: I've added some documentation for this now.

crypt.go
```diff
@@ -0,0 +1,494 @@
...
+	return &execError{path: cmd.Path, err: err}
+}
+
+func isExecError(err error, path string) bool {
+	var e *execError
+	return xerrors.As(err, &e) && e.path == path
+}
+
+func activate(volumeName, sourceDevicePath string, key []byte, options []string) error {
+	keyFilePath, err := func() (string, error) {
```
cmatsuoka, line 73: This can be done in a follow-up, but I think it could be interesting to consider using a named pipe or similar mechanism to avoid creating temporary files containing the key.

crypt.go
pedronis: my suggestions are:

1. s/RecoveryReason/RecoverKeyRecourse/
2. if 1. sounds too weird, it might or might not, then be very explicit with something like s/RecoveryReason/RecoverKeyUsageReason/

errors.go
```diff
@@ -104,3 +116,33 @@ type InvalidKeyFileError struct {
...
+}
+
+// ActivateWithTPMSealedKeyError is returned from ActivateVolumeWithTPMSealedKey if activation with the TPM protected key failed.
+type ActivateWithTPMSealedKeyError struct {
+	// TPMErr details the error that occurred during activation with the TPM sealed key.
+	TPMErr error
+
+	// RecoveryErr details the error that occurred during activation with the fallback recovery key, if activation with the recovery key
+	// was also unsuccessful.
+	RecoveryErr error
```
pedronis, line 140: this should be RecoveryKeyErr or RecoveryKeyRecourseErr, see other discussion about just using plain "Recovery"

crypt.go
```diff
@@ -0,0 +1,494 @@
...
+		if err := f.Chmod(0600); err != nil {
+			return "", err
+		}
+		if _, err := f.Write(key); err != nil {
+			return "", xerrors.Errorf("cannot write key to file: %w", err)
+		}
+		return f.Name(), nil
+	}()
+	if err != nil {
+		return xerrors.Errorf("error saving key for systemd-cryptsetup: %w", err)
```
pedronis, line 88: cannot temporarily save key for systemd-cryptsetup:

crypt.go
```diff
@@ -0,0 +1,494 @@
...
+}
+
+func activateWithTPMKey(tpm *TPMConnection, volumeName, sourceDevicePath, keyPath string, pinReader io.Reader, pinTries int, lock bool, activateOptions []string) error {
+	var lockErr error
+	key, err := func() ([]byte, error) {
+		defer func() {
+			if !lock {
+				return
+			}
+			lockErr = LockAccessToSealedKeys(tpm)
```
pedronis, line 300: this code path doesn't seem tested

crypt.go
```diff
@@ -0,0 +1,494 @@
...
+	activateOptions, err := makeActivateOptions(options.ActivateOptions)
+	if err != nil {
+		return false, err
+	}
+
+	if err := activateWithTPMKey(tpm, volumeName, sourceDevicePath, keyPath, pinReader, options.PINTries, options.LockSealedKeyAccess, activateOptions); err != nil {
+		reason := RecoveryReasonUnexpectedError
+		switch {
+		case isLockAccessError(err):
+			return false, LockAccessToSealedKeysError(err.Error())
```
pedronis, line 436: fwiw, tests don't seem to reach here

crypt.go
```diff
@@ -0,0 +1,494 @@
...
+		case pinTries == 0:
+			pinTries = 1
+		}
+
+		var key []byte
+
+		for ; pinTries > 0; pinTries-- {
+			var pin string
+			if k.AuthMode2F() == AuthModePIN {
+				r := pinReader
```
pedronis, line 320: this block doesn't seem tested atm

crypt.go
```diff
@@ -0,0 +1,494 @@
...
+		case isLockAccessError(err):
+			return false, LockAccessToSealedKeysError(err.Error())
+		case xerrors.Is(err, ErrTPMLockout):
+			reason = RecoveryReasonTPMLockout
+		case xerrors.Is(err, ErrTPMProvisioning):
+			reason = RecoveryReasonTPMProvisioningError
+		case isInvalidKeyFileError(err):
+			reason = RecoveryReasonInvalidKeyFile
+		case xerrors.Is(err, requiresPinErr):
+			reason = RecoveryReasonPINFail
```
pedronis, line 444: this doesn't seem exercised

crypt.go
```diff
@@ -0,0 +1,494 @@
...
+		case xerrors.Is(err, ErrPINFail):
+			reason = RecoveryReasonPINFail
```
pedronis, line 446: same

crypt.go
chrisccoulson: I think RecoveryKeysUsageReason sounds better, even if it's longer.

crypt.go
chrisccoulson: I'm not sure how to test this yet - it's difficult to hit any of the error conditions in a test.

crypt.go
chrisccoulson: Same as above.

crypt.go
chrisccoulson: This requires https://github.com/snapcore/secboot/pull/34

crypt.go
chrisccoulson: This requires https://github.com/snapcore/secboot/pull/34

crypt.go
chrisccoulson: This requires https://github.com/snapcore/secboot/pull/34

crypt.go
pedronis: ok, I left some questions there last week

errors.go
pedronis: I suppose actually, RecoveryKeyUsageErr given the last naming decision

PR #48, opened 2020-04-15
Add an API to bind a PCR profile to a set of snap models

chrisccoulson:
This binds a PCR profile to the brand (via the brand-id and sign-key-sha3-384 fields), model and series properties of a set of model assertions and defines the format in which snap-bootstrap should perform the actual measurement of these properties.

### DIFF
--- a/snapmodel_policy.go
+++ b/snapmodel_policy.go
@@ -23,12 +23,35 @@ import (
 	"encoding/base64"
 	"encoding/binary"
 	"errors"
+	"fmt"
 
 	"github.com/canonical/go-tpm2"
 	"github.com/snapcore/snapd/asserts"
 
 	"golang.org/x/xerrors"
 )
+
+const (
+	modelGradeUnset uint8 = iota
+	modelGradeSecured
+	modelGradeSigned
+	modelGradeDangerous
+)
+
+func modelGradeEnum(grade asserts.ModelGrade) (uint8, error) {
+	switch grade {
+	case asserts.ModelGradeUnset:
+		return modelGradeUnset, nil
+	case asserts.ModelSecured:
+		return modelGradeSecured, nil
+	case asserts.ModelSigned:
+		return modelGradeSigned, nil
+	case asserts.ModelDangerous:
+		return modelGradeDangerous, nil
+	default:
+		return modelGradeUnset, fmt.Errorf("unknown grade: %v", grade)
+	}
+}
 
 // SnapModelProfileParams provides the parameters to AddSnapModelProfile.
 type SnapModelProfileParams struct {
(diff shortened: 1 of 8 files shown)


tpm_test.go
```diff
@@ -59,7 +59,7 @@ var (
 	testEncodedEkCertChain []byte
 )
 
-func decodeHexString(t *testing.T, s string) []byte {
+func decodeHexStringT(t *testing.T, s string) []byte {
```
pedronis, line 62: my suggestion (no blocker) in this sort of situation given that we expect this to succeed in most cases, and is not really code that in itself needs to be tested, is to have a single helper that panics on error, in the spirit basically of must* helpers in the stdlib (not saying it must be called must* in this case)

snapmodel_policy.go
```diff
@@ -0,0 +1,100 @@
...
+	// Models is the set of models to add to the PCR profile.
+	Models []*asserts.Model
+}
+
+// AddSnapModelProfile adds the snap model profile to the PCR protection profile, as measured by snap-bootstrap, in order to generate
+// a PCR policy that is bound to a specific set of device models. It is the responsibility of snap-bootstrap to verify the integrity
+// of the model that it has measured.
+//
+// The measurement of a model is a digest computed as follows (where H is the supplied params.PCRAlgorithm):
+//  digest1 = H(tpm2.HashAlgorithmSHA384 || sign-key-sha3-384 || brand-id)
```
pedronis, line 52: can you expand a bit why we include sign-key-sha3-384, is the same principle I suppose by which we measure the keys that are used to verify the images?

snapmodel_policy.go
```diff
@@ -0,0 +1,100 @@
...
+//  digest2 = H(digest1 || model)
```
pedronis, line 53: (series, brand-id, model) is supposed to be unique but this is really not controlled by anyone by the brands themselves, also an assertion can be revised (different assertion revision). As previously discussed because it controls some fundamental bits of snapd policy vs how it handles/what it allows wrt recovery seeds and encryption, I think we want also to measure  []byte(model.Grade())

snapmodel_policy.go
```diff
@@ -0,0 +1,100 @@
...
+//  digest3 = H(digest2 || series)
+// String fields are hashed without null terminators, the signing key digest algorithm identifier is encoded in little-endian
+// format, and the sign-key-sha3-384 field is hashed in decoded (binary) form. Separate extend operations are used because brand-id,
+// model and series are variable length.
+//
+// The PCR index that snap-bootstrap measures the model to can be specified via the PCRIndex field of params.
+//
+// The set of models to add to the PCRProtectionProfile is specified via the Models field of params.
+func AddSnapModelProfile(profile *PCRProtectionProfile, params *SnapModelProfileParams) error {
```
pedronis, line 62: this sets up the policy. But I suppose we want a corresponding helper to do the actual measurement:

MeasureSnapModel(tpm, PCRIndex,  model *asserts.Model)

right ?

snapmodel_policy_test.go
```diff
@@ -0,0 +1,401 @@
...
+}
+
+func (s *snapModelProfileTest) TestAddSnapModelProfile1(c *C) {
+	s.testAddSnapModelProfile(c, &testAddSnapModelProfileData{
+		params: &SnapModelProfileParams{
+			PCRAlgorithm: tpm2.HashAlgorithmSHA256,
+			PCRIndex:     12,
+			Models: []*asserts.Model{
+				assertstest.FakeAssertion(map[string]interface{}{
+					"type":         "model",
```
pedronis, line 70: to be clear these are all UC18 style models,  not UC20 ones,  UC20 ones look like this:

https://github.com/snapcore/models/blob/master/ubuntu-core-20-amd64.model
https://github.com/snapcore/snapd/blob/master/seed/seed20_test.go#L106

tpm_test.go
chrisccoulson: This is temporary - there will eventually be a single helper once the rest of the tests are using gopkg.in/check.v1.

snapmodel_policy.go
chrisccoulson: I've added the grade now, but converted it to a single byte enum to avoid having another variable length field to digest.

snapmodel_policy_test.go
chrisccoulson: I've updated the tests to use UC20 style models now.

snapmodel_policy.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+const (
+	modelGradeUnset uint8 = iota
+	modelGradeSecured
+	modelGradeSigned
+	modelGradeDangerous
+)
+
+func modelGradeEnum(grade asserts.ModelGrade) (uint8, error) {
```
pedronis, line 41: as discussed this should be implemented directly by asserts as ModelGrade.ToByte or similar, it's not a blocker to land this, but I expect the values used to diverge from what is used at the moment here. I will work on that ASAP.

chrisccoulson: Thanks. Also, it doesn't have to be a byte - if you think there's a chance that it might need more than a byte in the future, then any other fixed size is ok

pedronis: yes, maybe is better to consider future extensions/extensibility and go for 32 bits

PR #50, opened 2020-04-20
Expect ModelGrade.Code to be measured

pedronis:
asserts.ModelGrade is growing a Code method to give us a fixed size
bit representation convenient for measuring, expect the use of that

drive-by: make the series values in the tests more realistic

### DIFF
--- a/snapmodel_policy.go
+++ b/snapmodel_policy.go
@@ -57,7 +57,7 @@ type SnapModelProfileParams struct {
 //  digest2 = H(digest1 || model)
 //  digestModel = H(digest2 || series || grade)
 // The signing key digest algorithm is encoded in little-endian format, and the sign-key-sha3-384 field is hashed in decoded (binary)
-// form. The brand-id, model and series fields are hashed without null terminators. The grade field is encoded as the 32 bits from asserts.ModelGrade.Code.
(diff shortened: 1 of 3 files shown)


pedronis:
Requires https://github.com/snapcore/snapd/pull/8522

pedronis:
@chrisccoulson the snapd bits have landed and now this is green as well

snapmodel_policy.go
```diff
@@ -80,12 +57,8 @@ type SnapModelProfileParams struct {
 //  digest2 = H(digest1 || model)
 //  digestModel = H(digest2 || series || grade)
 // The signing key digest algorithm is encoded in little-endian format, and the sign-key-sha3-384 field is hashed in decoded (binary)
-// form. The brand-id, model and series fields are hashed without null terminators. The grade field is encoded as a single byte with
-// the following conversion:
-//  "unset":     0
-//  "secured":   1
-//  "signed":    2
-//  "dangerous": 3
+// form. The brand-id, model and series fields are hashed without null terminators. The grade field is encoded as the 32 bits from asserts.ModelGrade.Code.
```
chrisccoulson, line 60: I would be explicit that the 4-byte grade field is encoded in little endian.

pedronis: done

PR #54, opened 2020-04-22
Add APIs for initializing LUKS volumes

chrisccoulson:
This adds 3 functions:
- InitializeLUKS2Container for formatting a new LUKS2 container with the
default settings and a supplied label. The supplied key is the one that
should be protected by the TPM.
- AddRecoveryKeyToLUKS2Container for adding a 16-byte recovery key to
an existing container.
- ChangeLUKS2KeyUsingRecoveryKey for changing the default key, which
is intended to be used after unlocking with the recovery key if the TPM
sealed key file is invalid.


crypt.go
```diff
@@ -493,3 +508,136 @@ func ActivateVolumeWithRecoveryKey(volumeName, sourceDevicePath string, keyReade
...
+		devicePath)
+	cmd.Stdin = bytes.NewReader(key[:])
+	if output, err := cmd.CombinedOutput(); err != nil {
+		return osutil.OutputErr(output, err)
+	}
+
+	return setLUKS2KeyslotPreferred(devicePath, 0)
+}
+
+func addKeyToLUKS2Container(devicePath string, existingKey, key []byte, extraArgs []string) error {
```
pedronis, line 565: maybe extraArgs should be extraOptionArgs to make clear these come before the positional args

crypt.go
```diff
@@ -493,3 +508,136 @@ func ActivateVolumeWithRecoveryKey(volumeName, sourceDevicePath string, keyReade
...
+		// because the supplied input key has the same entropy (512-bits) as the derived key and therefore
+		// increased time or memory cost don't provide a security benefit (but does slow down unlocking).
+		"--pbkdf", "argon2i", "--pbkdf-force-iterations", "4", "--pbkdf-memory", "32",
+		// set LUKS2 label
+		"--label", label,
+		// device to format
+		devicePath)
+	cmd.Stdin = bytes.NewReader(key[:])
+	if output, err := cmd.CombinedOutput(); err != nil {
+		return osutil.OutputErr(output, err)
```
pedronis, line 559: should the failure paths be tested?

crypt.go
```diff
@@ -493,3 +508,136 @@ func ActivateVolumeWithRecoveryKey(volumeName, sourceDevicePath string, keyReade
...
+		"--new-keyfile-offset", strconv.Itoa(len(existingKey)), "--new-keyfile-size", strconv.Itoa(len(key))}
+	args = append(args, extraArgs...)
+	args = append(args,
+		// container to add key to
+		devicePath,
+		// read recovery key from temporary file at previously specified offset
+		keyFilePath)
+	cmd := exec.Command("cryptsetup", args...)
+	if output, err := cmd.CombinedOutput(); err != nil {
+		return osutil.OutputErr(output, err)
```
pedronis, line 591: same question

crypt.go
```diff
@@ -493,3 +508,136 @@ func ActivateVolumeWithRecoveryKey(volumeName, sourceDevicePath string, keyReade
...
+// be used to activate the LUKS2 container.
+func ChangeLUKS2KeyUsingRecoveryKey(devicePath string, recoveryKey [16]byte, key []byte) error {
+	if len(key) != 64 {
+		return fmt.Errorf("expected a key length of 512-bits (got %d)", len(key)*8)
+	}
+
+	cmd := exec.Command("cryptsetup", "luksKillSlot", "--key-file", "-", devicePath, "0")
+	cmd.Stdin = bytes.NewReader(recoveryKey[:])
+	if output, err := cmd.CombinedOutput(); err != nil {
+		return osutil.OutputErr(output, err)
```
pedronis, line 629: same question

chrisccoulson:
> Thanks Chris. I previously used a named pipe to pass the key to cryptsetup when adding the recovery key, but I see you consider a temporary file to be good enough in this case.

I think there was a reason for this, and I probably should have left a comment because I've forgotten what it was now. I'll have a play around with cryptsetup again.

PR #56, opened 2020-04-22
Add MeasureSnapSystemEpochToTPM

chrisccoulson:


snapmodel_policy.go
```diff
@@ -136,13 +139,31 @@ func MeasureSnapModelToTPM(tpm *TPMConnection, pcrIndex int, model *asserts.Mode
...
+			return xerrors.Errorf("cannot compute digest for algorithm %v: %w", s.Hash, err)
 		}
 
 		digests = append(digests, tpm2.TaggedHash{HashAlg: s.Hash, Digest: digest})
 	}
 
 	return tpm.PCRExtend(tpm.PCRHandleContext(pcrIndex), digests, tpm.HmacSession())
 }
+
+// MeasureSnapSystemEpochToTPM measures a digest of uint32(0) to the specified PCR for all supported PCR banks. See the documentation
```
pedronis, line 153: maybe:  ... a digest of the hardcoded system epoch (uint32(0)) ...

snapmodel_policy.go
```diff
@@ -136,13 +139,31 @@ func MeasureSnapModelToTPM(tpm *TPMConnection, pcrIndex int, model *asserts.Mode
...
+// for AddSnapModelProfile for more details.
+func MeasureSnapSystemEpochToTPM(tpm *TPMConnection, pcrIndex int) error {
+	return measureSnapPropertyToTPM(tpm, pcrIndex, func(alg tpm2.HashAlgorithmId) (tpm2.Digest, error) {
+		h := alg.NewHash()
+		binary.Write(h, binary.LittleEndian, uint32(0))
```
pedronis, line 158: maybe we do still want a:

const zeroSnapSystemEpoch uint32 = 0

to use here and AddSnapModelProfile.

Otherwise have these 3 lines as a helper:

zeroSnapSystemEpochDigest(tpm2.HashAlgorithm)  tpm2.Hash

? to call from here and from AddSnapModelProfile perhaps

PR #58, opened 2020-04-25
Support dual signed efi binaries

chrisccoulson:


secureboot_policy.go
```diff
@@ -865,42 +915,58 @@ func (g *secureBootPolicyGen) computeAndExtendVerificationMeasurement(paths []*s
...
+						KeyUsages:     []x509.ExtKeyUsage{x509.ExtKeyUsageAny}}
+					if _, err := sig.signer.Verify(opts); err == nil {
+						// This signer certificate is trusted by this authority
+						authority = &secureBootAuthority{signature: caSig, source: db}
+						break
+					}
+
+					if authority != nil {
+						break
+					}
```
pedronis, line 956: do we need this check at this level?  also wouldn't  using  break Label as the original code be simpler?  I am probably missing something

secureboot_policy.go
```diff
@@ -174,39 +177,39 @@ func (c *winCertificateAuthenticode) wCertificateType() uint16 {
 
 // decodeWinCertificate decodes the WIN_CERTIFICATE implementation from r. Currently supported types are WIN_CERT_TYPE_PKCS_SIGNED_DATA
 // and WIN_CERT_TYPE_EFI_GUID.
-func decodeWinCertificate(r io.Reader) (winCertificate, error) {
+func decodeWinCertificate(r io.Reader) (winCertificate, int, error) {
```
pedronis, line 180: this should probably have named return values now, it's not super clear what the int one is, as it is

secureboot_policy.go
chrisccoulson: I've added the label back now

secureboot_policy.go
chrisccoulson: I've named them now - the int is the total number of bytes read

PR #61, opened 2020-05-01
Use the PIN to derive additional key material

chrisccoulson:
The default configuration of having no PIN set protects data from
opportunistic attackers. Setting a PIN protects data from more
determined attackers. This is the case even if the PIN is fairly low
entropy, because we use the TPM to assert that the provided PIN is
correct and this benefits from the TPM's dictionary attack logic. It is
beyond the ability even for most determined attackers to compromise the
actual TPM hardware to defeat this mechanism.

However, this may not be sufficient to protect your data from 3 letter
agencies with the means to compromise the TPM hardware in a way that
allows them access to a shielded location without the use of a protected
capability (a function that operates on a shielded location) - the type
of thing that would only be possible by state actors. Setting a higher
entropy PIN or passhprase doesn't help here, because the TPM just asserts
that the supplied PIN or passphrase is correct - it doesn't provide any
additional key material required for unsealing the disk encryption key,
and the PIN is stored inside the TPM's NV memory too. All of the key
material required to unlock an encrypted volume lives on the device.

In the future, we may want to support a network unlocking feature, where a
trusted server can work with the existing TPM protected key object to
authorize disk unlocking as long as the device is physically attached to
a specific network. An implementation of this should require that the
trusted server contributes key material required to unlock an encrypted
volume, but this doesn't fit well with the current design.

Some use cases for network unlocking are:
- Preventing automatic unlocking of devices that are detached from a
specific network (ie, if they are stolen).
- Bypassing the requirement for a PIN in enterprise environments when
devices are physically attached to a corporate network (where a PIN is
required when the device isn't physically attached to it).

This commit changes how the PIN is used to protect a key in order to
improve this.

During SealKeyToTPM, rather than sealing the disk encryption key (DEK) to
the TPM, it now:
- Generates an intermediate key (IK).
- Encrypts DEK with IK and an initialization vector (DEKiv) to produce
DEKenc.
- Seals DEKenc to the TPM.
- Passes IK through a AF splitter to produce IKsplit.
- Saves DEKiv and IKsplit to the metadata of the sealed key object.

During unsealing when a PIN hasn't been set, the following happens:
- The TPM unseals DEKenc if the PCR values match the PCR policy.
- IKsplit is merged to recreate IK.
- DEKiv and IK are used to decrypt DEKenc to recover DEK.

When a PIN is set, the following happens:
- The PIN and a salt are passed through Argon2i to create a derived key
(DK).
- The upper 32-bits of DK are used for the authorization value of the
associated NV index on the TPM, which is updated accordingly.
- The lower 32-bits of DK are used with an initialization vector (IKiv) to
encrypt IKsplit.
- The encrypted IKsplit is written to the sealed key object metadata,
replacing the unencrypted version.
- The salt and IKiv are also written to the sealed key object metadata.

Unsealing with a PIN looks like this:
- The supplied PIN and a salt are passed through Argon2i to create a
derived key (DK).
- The TPM unseals DEKenc if the PCR values match the PCR policy and the
upper 32-bits of DK match the authorization value of the PIN NV index.
- The lower 32-bits of DK are used along with IKiv to decrypt IKsplit.
- The decrypted IKsplit is merged to recreate IK.
- DEKiv and IK are used to decrypt DEKenc to recover DEK.

This scheme retains the property that low-entropy PINs provide some
protection of data against most determined attackers, because it still
relies on the TPM asserting knowledge of the correct PIN before unsealing
the encrypted DEK, and this benefits from the TPM's dictionary attack
protection. Also, without a compromise of the TPM hardware, the most
recently set PIN is always required in order to unlock a volume, even if
an attacker has a backup of a sealed key data file with an IK encrypted
with a previous PIN.

This scheme should be more suitable for incorporating a network unlock
feature in the future, where IK can be protected with the public key of
a trusted device on the local network (an assertion signed by the network
server's private key would probably also be require to allow the encrypted
DEK to be unsealed from the TPM).

### DIFF
--- a/keydata.go
+++ b/keydata.go
@@ -22,10 +22,14 @@ package secboot
 import (
 	"bytes"
 	"crypto"
+	"crypto/aes"
+	"crypto/cipher"
 	"crypto/rsa"
 	"crypto/x509"
+	"encoding/binary"
 	"errors"
 	"fmt"
+	"hash"
 	"io"
 	"math/big"
 	"os"
@@ -35,6 +39,8 @@ import (
 	"github.com/snapcore/snapd/osutil/sys"
 
 	"golang.org/x/xerrors"
+
+	"maze.io/x/crypto/afis"
 )
 
 const (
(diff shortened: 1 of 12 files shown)


pin.go
```diff
@@ -253,6 +388,88 @@ func performPinChange(tpm *tpm2.TPMContext, public *tpm2.NVPublic, authPolicies
...
+		Salt:           salt[:],
+		Time:           timeCost,
+		Memory:         memoryCost,
+		Threads:        threads,
+		DerivedKeySize: derivedKeySize}, nil
+}
+
+// PINParams provides some additional parameters to ChangePIN.
+type PINParams struct {
+	// MaxMemoryCost is used to specify the maximum amount of memory in MB to use when performing key derivation with the PIN.
```
setharnold: MB or KB? :)

keydata.go
```diff
@@ -61,14 +67,134 @@ type keyPolicyUpdateData struct {
...
+		return nbytes, err
+	}
+
+	var s uint32
+	if err := binary.Read(r, binary.BigEndian, &s); err != nil {
+		return nbytes, err
+	}
+	nbytes += binary.Size(s)
+
+	d.Data = make([]byte, s)
```
setharnold, line 114: I'm concerned about these values being used to allocated memory before copying from the reader -- I think we ought to bound these to reasonable limits.

pedronis:
will this break using preexisting sealed keys?

chrisccoulson:
> will this break using preexisting sealed keys?

It will, although this is still a WIP. Depending on when it lands, it might be possible to bump the metadata version and keep the handling for unsealing old keys.

setharnold:
On Wed, May 13, 2020 at 01:05:47AM -0700, Chris Coulson wrote:
> > will this break using preexisting sealed keys?
> 
> It will, although this is still a WIP. Depending on when it lands, it
> might be possible to bump the metadata version and keep the handling for
> unsealing old keys.

Considering that this is unreleased, I'd like us to avoid building in too
much complexity for pre-existing systems. Security problems often hide in
code designed to handle corner cases.

I know it's a serious inconvenience to re-do a machine's FDE key, but our
inconvenience now may save us from maintaining twice as many images in
testing infrastructure, twice as many test cases to run, and possible
future security problems around handling the previous systems.

Thanks

PR #64, opened 2020-05-12
Don't create a transient EK in ConnectToDefaultTPM

chrisccoulson:
If the TPM doesn't have a persistent endorsement key, both
ConnectToDefaultTPM and SecureConnectToDefaultTPM try to create a
transient one. RSA key creation on hardware TPMs is quite slow (in
excess of 10 seconds), so don't do this inside ConnectToDefaultTPM.

### DIFF
--- a/tpm.go
+++ b/tpm.go
@@ -216,10 +216,10 @@ func (t *TPMConnection) init() error {
 	// Without verification against the EK certificate, ek isn't yet safe to use for secret sharing with the TPM.
 	ek, err := func() (tpm2.ResourceContext, error) {
 		ek, err := t.CreateResourceContextFromTPM(ekHandle)
-		if err == nil {
+		if err == nil || !secureMode {
 			return ek, nil
 		}
-		if !tpm2.IsResourceUnavailableError(err, ekHandle) || !secureMode {
+		if !tpm2.IsResourceUnavailableError(err, ekHandle) {
 			return nil, err
 		}
 		if ek, err := createTransientEk(t.TPMContext); err == nil {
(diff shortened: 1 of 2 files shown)


tpm.go
```diff
@@ -217,38 +219,39 @@ func (t *TPMConnection) init() error {
 		if err == nil {
 			return ek, nil
 		}
-		if !tpm2.IsResourceUnavailableError(err, ekHandle) {
+		if !tpm2.IsResourceUnavailableError(err, ekHandle) || !secureMode {
```
pedronis, line 222: if we are in !secureMode we return the error immediately but then we just ignore it afaict, it probably would be a more readable to turn the error in that case to nil here already

tpm.go
```diff
@@ -217,38 +219,39 @@ func (t *TPMConnection) init() error {
...
 			return nil, err
 		}
 		if ek, err := createTransientEk(t.TPMContext); err == nil {
 			return ek, nil
 		}
 		return nil, err
 	}()
-	if err != nil && len(t.verifiedEkCertChain) > 0 {
+	if err != nil && secureMode {
```
pedronis, line 230: then we don't need the && part here

tpm.go
```diff
@@ -217,38 +219,39 @@ func (t *TPMConnection) init() error {
...
 		// A lack of EK should be fatal in this context
 		return xerrors.Errorf("cannot obtain context for EK: %w", err)
 	}
 
-	ekIsPersistent := false
```
pedronis, line 233: even if we don't want to keep the flag and update it, I think a local function:

```
ekIsPersistent := func() bool {
     return ek != nil && ek.Handle() == ekHandle
}
```

could make the code slightly more readable

chrisccoulson:
I think I've addressed the comments now.

tpm.go
```diff
@@ -225,30 +227,32 @@ func (t *TPMConnection) init() error {
...
 		// object, then try to create a transient EK with the provided authorization and make another attempt at verification, in case
 		// the persistent object isn't a valid EK.
 		rc, err := func() (tpm2.ResourceContext, error) {
 			err := verifyEk(t.verifiedEkCertChain[0], ek)
 			if err == nil {
-				return ek, nil
+				return nil, nil
 			}
-			if ek.Handle().Type() == tpm2.HandleTypeTransient {
+			if ek.Handle() != ekHandle {
```
pedronis, line 255: this could be for consistency and maybe a bit more clarity:  !ekIsPersistent()

PR #65, opened 2020-05-12
Switch from RSA-PSS to ECDSA for signing keys

chrisccoulson:
We create 1 signing key during provisioning, and 2 during sealing. EC keys are a bit faster to create than RSA ones, so switch from RSA-PSS to ECDSA where we sign digests that have to be verified by the TPM.

### DIFF
--- a/keydata.go
+++ b/keydata.go
@@ -22,6 +22,7 @@ package secboot
 import (
 	"bytes"
 	"crypto"
+	"crypto/ecdsa"
 	"crypto/rsa"
 	"crypto/x509"
 	"errors"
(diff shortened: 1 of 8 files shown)


pedronis:
is this just temporary setup data or will this break using preexisting seal artifacts (keydata, ...)?

chrisccoulson:
> is this just temporary setup data or will this break using preexisting seal artifacts (keydata, ...)?

Yeah, this would break existing keys, unless we bumped the version and kept handling for both key types.

pedronis:
>  > is this just temporary setup data or will this break using preexisting seal artifacts (keydata, ...)?

> Yeah, this would break existing keys, unless we bumped the version and kept handling for both key types.

given that we are about to ship the beta, and although we aren't sure we will be able to keep it updating, but there will be a few snapd releases at this point between now and GA. I think we should try to keep old installations being able to open their disks even after snapd/kernel refreshes for a while at least.

chrisccoulson:
I've submitted another PR that will be a pre-requisite for this: https://github.com/snapcore/secboot/pull/66

chrisccoulson:
This has been updated now, and the changes are backwards compatible with current keyfiles. The target of the PR is v1-wip as opposed to master.

keydata.go
```diff
@@ -86,7 +94,7 @@ func (d *keyPolicyUpdateData) Unmarshal(r io.Reader) (nbytes int, err error) {
 	}
 
 	switch version {
-	case 0, 1:
+	case 0:
 		var raw keyPolicyUpdateDataRaw_v0
```
pedronis, line 98: this path and other corresponding paths in the file are not tested anymore. If I understand correctly we still need to have this working?

policy.go
```diff
@@ -137,15 +140,46 @@ func incrementDynamicPolicyCounter(tpm *tpm2.TPMContext, nvPublic *tpm2.NVPublic
...
 	binary.Write(h, binary.BigEndian, int32(0)) // expiration
 
 	// Sign the digest
-	sig, err := rsa.SignPSS(rand.Reader, key, signDigest.GetHash(), h.Sum(nil), &rsa.PSSOptions{SaltLength: rsa.PSSSaltLengthEqualsHash})
-	if err != nil {
-		return xerrors.Errorf("cannot sign authorization: %w", err)
+	var signature tpm2.Signature
+	switch k := key.(type) {
+	case *rsa.PrivateKey:
+		sig, err := rsa.SignPSS(rand.Reader, k, signDigest.GetHash(), h.Sum(nil), &rsa.PSSOptions{SaltLength: rsa.PSSSaltLengthEqualsHash})
```
pedronis, line 159: this path is not exercised anymore

policy.go
```diff
@@ -613,17 +648,34 @@ func computeDynamicPolicy(version uint32, alg tpm2.HashAlgorithmId, input *dynam
 	h.Write(authorizedPolicy)
 
 	// Sign the digest
-	sig, err := rsa.SignPSS(rand.Reader, input.key, input.signAlg.GetHash(), h.Sum(nil), &rsa.PSSOptions{SaltLength: rsa.PSSSaltLengthEqualsHash})
-	if err != nil {
-		return nil, xerrors.Errorf("cannot provide signature for initializing NV index: %w", err)
-	}
+	var signature tpm2.Signature
+	if version == 0 {
+		sig, err := rsa.SignPSS(rand.Reader, input.key.(*rsa.PrivateKey), input.signAlg.GetHash(), h.Sum(nil),
```
pedronis, line 653: same here

pedronis (commented):
this looks good to me but it seems we are not testing still using v 0  keyPolicyUpdateData ?

pedronis (approved):
sorry, one needs to run coverage with -coverpkg . to get the coverage contributed by the compat tests.

lgtm

PR #66, opened 2020-05-13
Make it easier to bump the metadata version

chrisccoulson:
This splits keyData, keyPolicyUpdateData, staticPolicyData and
dynamicPolicyData in to version agnostic parts and version-specific
parts. The version-specific parts are what is marshalled / unmarshalled.

### DIFF
--- a/policy.go
+++ b/policy.go
@@ -67,6 +67,16 @@ type policyOrDataNode struct {
 
 type policyOrDataTree []policyOrDataNode
 
+type dynamicPolicyDataRaw_v0 dynamicPolicyData
+
+func (d *dynamicPolicyDataRaw_v0) data() *dynamicPolicyData {
+	return (*dynamicPolicyData)(d)
+}
+
+func makeDynamicPolicyDataRaw_v0(data *dynamicPolicyData) *dynamicPolicyDataRaw_v0 {
+	return (*dynamicPolicyDataRaw_v0)(data)
+}
+
 type dynamicPolicyData struct {
 	PCRSelection              tpm2.PCRSelectionList
 	PCROrData                 policyOrDataTree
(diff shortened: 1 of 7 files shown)


policy.go
```diff
@@ -67,6 +67,16 @@ type policyOrDataNode struct {
 
 type policyOrDataTree []policyOrDataNode
 
+type dynamicPolicyDataRaw_v0 dynamicPolicyData
```
pedronis, line 70: the order of the definitions here is a bit odd,  I think maybe a comment could help about the relation of _v0 vs dynamicPolicyData atm

same for static

policy.go
```diff
@@ -67,6 +67,16 @@ type policyOrDataNode struct {
...
+
+func (d *dynamicPolicyDataRaw_v0) data() *dynamicPolicyData {
+	return (*dynamicPolicyData)(d)
+}
+
+func makeDynamicPolicyDataRaw_v0(data *dynamicPolicyData) *dynamicPolicyDataRaw_v0 {
```
pedronis, line 76: this name puzzled me a bit, but I suppose in general this wouldn't be a cast but produce a different *struct, right? so "make" makes sense

policy.go
```diff
@@ -565,7 +585,10 @@ func computePCRSelectionListFromValues(v tpm2.PCRValues) (out tpm2.PCRSelectionL
 
 // computeDynamicPolicy computes the part of an authorization policy associated with a sealed key object that can change and be
 // updated.
-func computeDynamicPolicy(alg tpm2.HashAlgorithmId, input *dynamicPolicyComputeParams) (*dynamicPolicyData, error) {
+func computeDynamicPolicy(version uint32, alg tpm2.HashAlgorithmId, input *dynamicPolicyComputeParams) (*dynamicPolicyData, error) {
+	if version != 0 {
```
pedronis, line 589: this probably merits a comment, because I suppose it is intentional that 0 is hardcoded

policy.go
chrisccoulson: Yeah, I did it this way because once the metadata format is changed, this will do something other than cast.

PR #68, opened 2020-05-14
Add ErrNoTPM2Device and TPMConnection.IsEnabled

chrisccoulson:
This makes use of tpm2.TPMContext.IsTPM2, newly added to go-tpm2 in order
to determine if the TPM device is actually a TPM2 device during connection.
If there is no TPM device or there is a TPM device but it is not a TPM2
device, then a new ErrNoTPM2Device error is returned.

Also added is TPMConnection.IsEnabled, which is used to determine whether
the TPM has been disabled by the platform firmware. A TPM can be disabled
by the platform firmware but still remain visible to the operating system.


tpm.go
```diff
@@ -356,10 +368,23 @@ var openDefaultTcti = func() (io.ReadWriteCloser, error) {
 func connectToDefaultTPM() (*tpm2.TPMContext, error) {
 	tcti, err := openDefaultTcti()
 	if err != nil {
+		if isPathError(err) {
+			return nil, nil
```
pedronis, line 372: returning a nil pointer and no error invites panics,  is there are reason to do that instead of returning the error of the level above already here? I don't see wrapping in the level above that would get confused atm

tpm.go
```diff
@@ -828,11 +853,16 @@ func EncodeEKCertificateChain(ekCert *x509.Certificate, parents []*x509.Certific
...
 // FetchAndSaveEKCertificateChain. It should not be used in any other scenario.
+//
+// If no TPM2 device is available, then a ErrNoTPM2Device error will be returned.
 func ConnectToDefaultTPM() (*TPMConnection, error) {
 	tpm, err := connectToDefaultTPM()
 	if err != nil {
 		return nil, err
 	}
+	if tpm == nil {
+		return nil, ErrNoTPM2Device
```
pedronis, line 864: I don't see test changes corresponding to this new error?

PR #70, opened 2020-06-16
Make PCRProtectionProfile.computePCRValues non recursive

chrisccoulson:
AddEFISecureBootPolicyProfile currently operates non-recursively. A
future PR will change the interaction between
AddEFISecureBootPolicyProfile and PCRProtectionProfile so that it
creates a profile using the methods of PCRProtectionProfile rather
than computing PCR values itself and adding them to the profile with
PCRProtectionProfile.AddPCRValue as it does now. This change will
preserve the property that computing a PCR profile for the secure
boot policy is non-recursive.

### DIFF
--- a/pcr_profile.go
+++ b/pcr_profile.go
@@ -75,26 +75,9 @@ type pcrProtectionProfileAddPCRValueInstr struct {
 	value tpm2.Digest
 }
 
-func (i *pcrProtectionProfileAddPCRValueInstr) apply(_ *tpm2.TPMContext, values pcrValuesList) (pcrValuesList, error) {
-	values.setValue(i.alg, i.pcr, i.value)
-	return values, nil
-}
-
 type pcrProtectionProfileAddPCRValueFromTPMInstr struct {
 	alg tpm2.HashAlgorithmId
 	pcr int
-}
-
-func (i *pcrProtectionProfileAddPCRValueFromTPMInstr) apply(tpm *tpm2.TPMContext, values pcrValuesList) (pcrValuesList, error) {
-	if tpm == nil {
-		return nil, fmt.Errorf("cannot read current value of PCR %d from bank %v: no TPM context", i.pcr, i.alg)
-	}
-	_, v, err := tpm.PCRRead(tpm2.PCRSelectionList{{Hash: i.alg, Select: []int{i.pcr}}})
-	if err != nil {
-		return nil, xerrors.Errorf("cannot read current value of PCR %d from bank %v: %w", i.pcr, i.alg, err)
-	}
-	values.setValue(i.alg, i.pcr, v[i.alg][i.pcr])
-	return values, nil
 }
 
 type pcrProtectionProfileExtendPCRInstr struct {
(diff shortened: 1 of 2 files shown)


pcr_profile.go
```diff
@@ -179,28 +144,109 @@ func (p *PCRProtectionProfile) AddProfileOR(profiles ...*PCRProtectionProfile) *
...
+// pcrProtectionProfileIterator provides a mechanism to perform a depth first traversal of instructions in a PCRProtectionProfile.
+type pcrProtectionProfileIterator struct {
+	instrs [][]pcrProtectionProfileInstr
+}
+
+func (iter *pcrProtectionProfileIterator) descendInToProfiles(profiles ...*PCRProtectionProfile) {
+	instrs := make([][]pcrProtectionProfileInstr, 0, len(profiles)+len(iter.instrs))
+	for _, p := range profiles {
+		pi := make([]pcrProtectionProfileInstr, len(p.instrs))
+		copy(pi, p.instrs)
```
pedronis, line 156: I'm not sure I understand the need of the copy,  pi is a slice which in itself is by value?

pcr_profile.go
```diff
@@ -179,28 +144,109 @@ func (p *PCRProtectionProfile) AddProfileOR(profiles ...*PCRProtectionProfile) *
...
+// pcrProtectionProfileComputeContext records state used when computing PCR values for a PCRProtectionProfile
+type pcrProtectionProfileComputeContext struct {
+	parent *pcrProtectionProfileComputeContext
+	values pcrValuesList
+}
+
+// computePCRValues computes a list of different PCR value combinations from this PCRProtectionProfile.
+func (p *PCRProtectionProfile) computePCRValues(tpm *tpm2.TPMContext) (pcrValuesList, error) {
+	// Create 2 contexts. The root context is not associated with any profile and exists just to collect PCR values from the root
+	// profile.
```
pedronis, line 203: this is confusing, the fact that the root context is not associated with the root profile but collect values from the root profile could use some clarification? it that related that we have a End for it without a starting Or? Maybe having some methods on Context could also help a bit with clarity

pcr_profile.go
chrisccoulson: Yeah, the copy is unnecessary in this case.

pcr_profile.go
chrisccoulson: I've removed this root context, cleaned up the whole of computePCRValues and added some more documentation to hopefully make it a bit more understandable.

pcr_profile.go
```diff
@@ -75,60 +75,25 @@ type pcrProtectionProfileAddPCRValueInstr struct {
...
-		v, err := p.computePCRValues(tpm, values.copy())
-		if err != nil {
-			// TODO: More context
-			return nil, err
-		}
-		out = append(out, v...)
-	}
-	return out, nil
-}
+type pcrProtectionProfileEndProfileInstr struct{}
```
pedronis, line 93: maybe also add a comment to this one that it's used as a marker?

pcr_profile.go
```diff
@@ -179,28 +144,149 @@ func (p *PCRProtectionProfile) AddProfileOR(profiles ...*PCRProtectionProfile) *
...
+// pcrProtectionProfileComputeContext records state used when computing PCR values for a PCRProtectionProfile
+type pcrProtectionProfileComputeContext struct {
+	parent *pcrProtectionProfileComputeContext
+	values pcrValuesList
+}
+
+// handleBranch is called when encountering a branch in a profile, and returns a slice of new *pcrProtectionProfileComputeContext
+// instances (one for each sub-branch). At the end of each sub-branch, finishBranch must be called on the associated
+// *pcrProtectionProfileComputeContext.
+func (c *pcrProtectionProfileComputeContext) handleBranch(n int) (out []*pcrProtectionProfileComputeContext) {
```
pedronis, line 203: shouldn't this be called handleBranches or handleBranching ? otherwise it not so clear that is setting state for n OR branches

PR #72, opened 2020-06-17
Stringify PCRProtectionProfile, which is helpful for debugging

chrisccoulson:


pcr_profile.go
```diff
@@ -192,6 +192,62 @@ func (p *PCRProtectionProfile) traverseInstructions() *pcrProtectionProfileItera
 	return i
 }
 
+type pcrProtectionProfileStringifyBranchContext struct {
+	index int
+	total int
+}
+
+func (p *PCRProtectionProfile) String() string {
```
pedronis, line 200: I think one happy test for this would be good even if it's only for debugging

PR #75, opened 2020-06-18
Add EFI boot manager code and boot attempts PCR profile

chrisccoulson:
This adds support for creating PCR profiles for the EFI boot manager
code and boot attempts PCR (PCR4), in order to generate a PCR policy
that restricts access to a specific set of EFI binaries.

This imports a copy of debug/pe to avoid a hard dependency on go 1.14.

The new API only works with the FileEFIImage implementation of EFIImage. SnapFileEFIImage will panic for now because computing the PE image hash requires knowledge of the file size.


pedronis:
@chrisccoulson this is quite large, could the internal/pe bits be split out to their own prereq PR?

(sorry I initially posted this to the wrong PR)

efibootmanager_policy.go
```diff
@@ -0,0 +1,335 @@
...
+	case *pe.OptionalHeader64:
+		isPe32Plus = true
+		sizeOfHeaders = int64(oh.SizeOfHeaders)
+		dd = oh.DataDirectory[0:oh.NumberOfRvaAndSizes]
+	default:
+		return nil, errors.New("PE binary doesn't contain an optional header")
+	}
+
+	// 1) Load the image header in to memory.
+	hr := io.NewSectionReader(r, 0, sizeOfHeaders)
```
pedronis, line 85: should we check that sizeOfHeaders is not unreasonably large?

efibootmanager_policy.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	// 10) Using the Offset field in the referenced SectionHeader structure as a key, arrange the table's elements in ascending order.
+	// In other words, sort the section headers in ascending order according to the disk-file offset of the sections.
+	sort.Slice(sections, func(i, j int) bool { return sections[i].Offset < sections[j].Offset })
+
+	for _, section := range sections {
+		// 11) Walk through the sorted table, load the corresponding section into memory, and hash the entire section. Use the
+		// Size field in the SectionHeader structure to determine the amount of data to hash.
+		sr := io.NewSectionReader(r, int64(section.Offset), int64(section.Size))
```
pedronis, line 152: should be check that section.Size is not unreasonably large? or chunk?

efibootmanager_policy.go
```diff
@@ -0,0 +1,335 @@
...
+		var certSize int64
+		if certTable != nil {
+			certSize = int64(certTable.Size)
+		}
+
+		if fileSize < (sumOfBytesHashed + certSize) {
+			return nil, errors.New("image too short")
+		}
+
+		b = make([]byte, fileSize-sumOfBytesHashed-certSize)
```
pedronis, line 184: same question about reasonable size?

efibootmanager_policy.go
```diff
@@ -0,0 +1,335 @@
...
+	branches []*PCRProtectionProfile
+}
+
+func (n *bootManagerCodePolicyGenBranch) branch() *bootManagerCodePolicyGenBranch {
+	b := &bootManagerCodePolicyGenBranch{profile: NewPCRProtectionProfile()}
+	n.branches = append(n.branches, b.profile)
+	return b
+}
+
+type bmLoadEventAndBranch struct {
```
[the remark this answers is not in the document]
chrisccoulson: I've added a comment now

efibootmanager_policy.go
```diff
@@ -0,0 +1,335 @@
...
+		loadEvents = loadEvents[1:]
+
+		digest, err := computePeImageDigest(params.PCRAlgorithm, e.event.Image)
+		if err != nil {
+			return err
+		}
+		e.branch.profile.ExtendPCR(params.PCRAlgorithm, bootManagerCodePCR, digest)
+
+		if len(e.event.Next) == 1 {
+			e.event = e.event.Next[0]
```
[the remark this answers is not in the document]
chrisccoulson: I've addressed this now.

efibootmanager_policy.go
chrisccoulson: I've add a helper to hash data 1 page at a time now

efibootmanager_policy.go
chrisccoulson: I've add a helper to hash data 1 page at a time now

efibootmanager_policy.go
chrisccoulson: I've add a helper to hash data 1 page at a time now

efibootmanager_policy.go
```diff
@@ -0,0 +1,335 @@
...
+		return nil, xerrors.Errorf("cannot decode PE binary: %w", err)
+	}
+
+	var isPe32Plus bool
+	var sizeOfHeaders int64
+	var dd []pe.DataDirectory
+	switch oh := p.OptionalHeader.(type) {
+	case *pe.OptionalHeader32:
+		sizeOfHeaders = int64(oh.SizeOfHeaders)
+		dd = oh.DataDirectory[0:oh.NumberOfRvaAndSizes]
```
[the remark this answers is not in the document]
chrisccoulson: The binaries we provide that will be parsed by this code should never hit this code path, so I didn't produce a binary to test it. The code is here more for correctness, in that it is one of the 2 COFF header formats.

efibootmanager_policy.go
```diff
@@ -0,0 +1,335 @@
...
+			nextLoadEvents = nil
+		}
+	}
+
+	for _, b := range allBranches {
+		if len(b.branches) == 0 {
+			continue
+		}
+		b.profile.AddProfileOR(b.branches...)
+	}
```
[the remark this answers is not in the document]
chrisccoulson: It happens on the first iteration of this loop, where b == root.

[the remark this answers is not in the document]
pedronis: ah, I somehow read {&root} much earlier as {}.  Anyway after that I stared at the loop some more, maybe it merits a comment that the order of processing allBranches doesn't matter because we are building a tree of OR instructions, not computing something.

efibootmanager_policy.go
pedronis: thank you

PR #77, opened 2020-06-19
Embed a copy of debug/pe from go 1.14.2

chrisccoulson:
Versions of debug/pe before go 1.14 do not correctly handle variable
length optional headers (where the data directory is less than 16 entries)
and therefore don't correctly decode kernel EFI images. We had a
workaround for this in secboot to allow us to decode the optional header,
but the subsequent section headers are still decoded incorrectly because
debug/pe thinks there is no optional header. Whilst we don't require the
section headers right now, a future PR will do, and the current code only
works by luck (as errors encountered when incorrectly decoding the section
table can cause pe.NewFile to fail anyway).

Embedding debug/pe avoids a hard dependency on go 1.14.


internal/pe/README
```diff
@@ -0,0 +1,7 @@
+This is a copy of debug/pe from go 1.14.2 (96745b980cfde139e8611772e2bc0c59a8e6cdf7)
```
pedronis, line 1: when we do this kind of thing (we had to in the past for some things in snapd itself) we tend to call the directory something like pe14

PR #78, opened 2020-06-19
Run the TPM simulator from the test binary's main()

chrisccoulson:
This is in preparation for creating internal/testutil, which can be
consumed by multiple test packages and will contain a helper for working
with the TPM simulator. This also removes the separate TPM interface /
platform port options - the platform TCP port is always the TPM interface
TCP port + 1.


tpm_test.go
```diff
@@ -476,7 +476,7 @@ func TestConnectToDefaultTPM(t *testing.T) {
 
 func TestSecureConnectToDefaultTPM(t *testing.T) {
 	SetOpenDefaultTctiFn(func() (io.ReadWriteCloser, error) {
-		return tpm2.OpenMssim(*mssimHost, *mssimTpmPort, *mssimPlatformPort)
+		return tpm2.OpenMssim("", *mssimPort, *mssimPort + 1)
```
cmatsuoka, line 479: Is the port, port+1 for the command and platform server ports the default for tpm2-simulator? (Because I see that only mssimPort is used when starting the simulator)

tpm_test.go
```diff
@@ -804,10 +823,21 @@ func TestMain(m *testing.M) {
...
-				fmt.Fprintf(os.Stderr, "Failed to open mssim connection: %v", err)
-				return 1
+			var tcti *tpm2.TctiMssim
+			// Give the simulator 5 seconds to start up
+		Loop:
+			for i := 0; ; i++ {
+				var err error
+				tcti, err = tpm2.OpenMssim("", *mssimPort, *mssimPort+1)
+				switch {
+				case err != nil && i == 4:
```
cmatsuoka, line 833: Is there a real situation where the simulator takes several seconds to start, or we're just playing it safe?

tpm_test.go
chrisccoulson: The TPM simulator opens the TPM interface on the port provided on the commandline (tcp port 2321 by default), and the platform port is always started on the next port number (tcp port 2322 by default). The fact that go-tpm2 requires both ports to be provided is probably a bug - it should just take a single port number and assume that the platform port will use the next port number.

tpm_test.go
chrisccoulson: Not several seconds, but this code does race with the simulator startup without having a retry here.

tpm_test.go
```diff
@@ -792,6 +792,25 @@ func TestMain(m *testing.M) {
 	rand.Seed(time.Now().UnixNano())
 	os.Exit(func() int {
 		if *useMssim {
+			mssimPath := ""
+			for _, p := range[]string{"tpm2-simulator", "tpm2-simulator-chrisccoulson.tpm2-simulator"} {
```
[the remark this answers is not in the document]
chrisccoulson: The TPM simulator is the TCG reference implementation, but I believe that libswtpm also embeds the reference implementation too. The simulator and swtpm serve different purposes - swtpm can be used to provide what looks like a real TPM interface to guests with all of the expected platform integration and associated limitations, whereas the simulator can't - it's really only meant for testing, and exposes a side-channel (the platform channel) for doing things you can't do with swtpm or a real TPM, such as resetting it.

chrisccoulson:
> Thanks Chris. I left some questions but overall it looks good. The simulator shutdown code looks a bit hard to follow in a late night debug session, perhaps you could consider having helpers for simulator launch/shutdown in a follow-up? Simulator-based tests would be interesting for snapd and helpers would be useful there.

The shutdown code is a bit complicated to handle the case where we can't cleanly shutdown the simulator from the test side. A few things need to happen to shut it down from the test side - we need to be able to establish connections to both the TPM interface and platform ports, we need to be able to send it a stop command, and then we need to be able to close the TCP connections on the test side, at which point the simulator closes its end and exits. If any of these fail, we just resort to killing the process, but this is undesirable in the normal case because the kernel will keep TCP connections around in the TIME_WAIT state for a couple of minutes afterwards, which prevents you from being able to start the simulator again for a short while.

The plan is to move this code to a helper in a follow-up.

PR #79, opened 2020-06-19
Don't panic when processing a profile with an empty branch

chrisccoulson:
Calling PCRProtectionProfile.AddProfileOR with no branches is pointless,
but it probably shouldn't result in a panic when computing PCR digests.
Handle this case gracefully by just omitting it from the profile.

### DIFF
--- a/pcr_profile.go
+++ b/pcr_profile.go
@@ -169,20 +169,29 @@ func (iter *pcrProtectionProfileIterator) next() pcrProtectionProfileInstr {
 	if len(iter.instrs) == 0 {
 		panic("no more instructions")
 	}
-	if len(iter.instrs[0]) == 0 {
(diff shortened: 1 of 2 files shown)


pcr_profile.go
```diff
@@ -169,20 +169,29 @@ func (iter *pcrProtectionProfileIterator) next() pcrProtectionProfileInstr {
 	if len(iter.instrs) == 0 {
```
pedronis, line 169: the callers detect this situation implicitly and avoid this panic, but I start to think a more explicit api would make it clearer that they do need to deal with this:

maybe the return of this should be (instr pcrProtectionProfileInstr, exhausted bool) instead of the panic

chrisccoulson: Yeah, I can probably look at this again in a follow up.

PR #81, opened 2020-07-01
Add a quirk when precomputing PCR values for UEFI signature database updates

chrisccoulson:
Section 8.2.2 of the UEFI spec, v2.8 states:

"For variables with the GUID EFI_IMAGE_SECURITY_DATABASE_GUID (i.e. where
the data buffer is formatted as EFI_SIGNATURE_LIST), the driver shall not
perform an append of EFI_SIGNATURE_DATA values that are already part of
the existing variable value"

When creating a PCR profile from precomputed digests before a signature
database is applied, we consider two signatures to be the same if the
entire EFI_SIGNATURE_DATA structure is the same. But firmware on Dell
machines and Intel NUCs that I have tested consider two signatures to be
the same if the SignatureData fields are the same, even if they have
different SignatureOwner values, which isn't how I'd interpret the
specification. EDK behaves how I interpreted the specification (two
signatures are the same only if their entire EFI_SIGNATURE_DATA structures
match).

As we don't know which firmware implementations behave in which way, and
as the PCR profile created from these pre-computed digests is only a
transient profile (the PCR profile should be recreated from digests
computed from the real signature database once all updates are applied),
create a PCR profile containing PCR values precomputed using both
de-duplication methods when there are signature database updates to apply.


secureboot_policy.go
```diff
@@ -711,7 +730,12 @@ type authenticodeSignerAndIntermediates struct {
 // secureBootPolicyGen is the main structure involved with computing secure boot policy PCR digests. It is essentially just
 // a container for EFISecureBootPolicyProfileParams - per-branch context is maintained in secureBootPolicyGenBranch instead.
 type secureBootPolicyGen struct {
-	*EFISecureBootPolicyProfileParams
+	pcrAlgorithm  tpm2.HashAlgorithmId
+	loadSequences []*EFIImageLoadEvent
+
+	events                     []*tcglog.Event
+	initialOSVerificationEvent *secureBootVerificationEvent
+	sigDbUpdates               []*secureBootDbUpdate
```
pedronis, line 738: no strong opinion, but could this also carry the quirk mode?

then the branches can get it via  branch.gen.... ?

PR #82, opened 2020-07-02
Use a named pipe for passing keys to cryptsetup where stdin can't be used

chrisccoulson:
systemd-cryptsetup can't read the key from stdin, so we passed it via a
temporary file. Whilst cryptsetup can read a key from stdin, it can
only read a single key and so isn't suitable for providing both keys in
the luksAddKey case. In this case, we passed both the existing key and
new key via a temporary file, and this couldn't be a named pipe because
it is required to be seekable.

Switch to using a named pipe for passing the key to systemd-cryptsetup.
For "cryptsetup luksAddKey", pass the existing key via a named pipe and
pass the new key on stdin. The "cryptsetup luksFormat" case is unchanged -
the initial key is passed to cryptsetup via stdin.


crypt.go
```diff
@@ -135,6 +139,23 @@ func activate(volumeName, sourceDevicePath string, key []byte, options []string)
...
 	}()
+
+	f, err := os.OpenFile(fifoPath, os.O_WRONLY, 0)
+	if err != nil {
+		// If we fail to open the write end, the read end will be blocked in open()
+		cmd.Process.Kill()
+		return xerrors.Errorf("cannot open FIFO for passing key to systemd-cryptsetup: %w", err)
+	}
+
+	if _, err := f.Write(key); err != nil {
```
cmatsuoka, line 150: I wonder if we can have a short write here, resulting in a situation where we have the right key, but it's only partially sent to cryptsetup.

crypt.go
```diff
@@ -135,6 +139,23 @@ func activate(volumeName, sourceDevicePath string, key []byte, options []string)
...
+		f.Close()
+		// The read end is open and blocked inside read(). Closing our write end will result in the
+		// read end returning 0 bytes (EOF) and exitting cleanly.
+		cmd.Wait()
+		return xerrors.Errorf("cannot pass key to systemd-cryptsetup: %w", err)
+	}
+
+	f.Close()
```
cmatsuoka, line 158: Should we check if this Close() call fails?

crypt.go
```diff
@@ -563,36 +584,54 @@ func InitializeLUKS2Container(devicePath, label string, key []byte) error {
...
 
-	return nil
+	f, err := os.OpenFile(fifoPath, os.O_WRONLY, 0)
+	if err != nil {
+		// If we fail to open the write end, the read end will be blocked in open()
+		cmd.Process.Kill()
+		return xerrors.Errorf("cannot open FIFO for passing existing key to cryptsetup: %w", err)
+	}
+
+	if _, err := f.Write(existingKey); err != nil {
```
cmatsuoka, line 622: Same question about short writes here.

crypt.go
chrisccoulson: This can't happen - the kernel guarantees that writes to pipes of less than one page are atomic - if you write() up to one page to a pipe that's opened with O_NONBLOCK, then either all bytes are written or no bytes are written (Without O_NONBLOCK, the write will block if there isn't enough room to write all of the data)

crypt.go
chrisccoulson: I did think about this, but I'm not sure what to do with the error. Perhaps just killing cryptsetup, as it will hang if the close fails anyway?

cmatsuoka: Yes, I think it's the only thing we can do.

chrisccoulson: I've merged it as-is for now. I had another think about it and I'm not really sure what would be the appropriate thing to do with the error. As close() does no I/O, there's nothing that could cause the syscall to fail in the kernel (it can only fail if the fd is invalid or already closed). As long as the close syscall executes, the cryptsetup process is guaranteed to continue and perform the requested action. I need to have a look at the go code to see if os.File.Close() can fail for any reason on the go side, and whether it can fail after the syscall completes successfully - if it can, then that would mean that both AddRecoveryKeyToLUKS2Container and ChangeLUKS2KeyUsingRecoveryKey could both return an error in some circumstances whilst still successfully making changes to the LUKS header.

PR #86, opened 2020-08-05
Test that changes are backwards compatible with supported metadata versions

chrisccoulson:
This adds an initial set of tests to make sure that future changes are
backwards compatible with key files created with the v0 version of the
on-disk metadata. The tests operate on a checked in key file and TPM
simulator state, created with a secboot version that creates key files
with the v0 metadata version.

In the future, additional tests will be added for every revision of the
on-disk metadata format version.

The tests cover operations that are expected to continue working on key
files created with older secboot versions, such as
UpdateKeyPCRProtectionPolicy, ChangePIN and SealedKeyObject.UnsealFromTPM.


internal/compattest/compattest_test.go
```diff
@@ -0,0 +1,238 @@
...
+	c.Check(secboot.UpdateKeyPCRProtectionPolicy(s.TPM, s.absPath("key"), s.absPath("pud"), pcrProfile), IsNil)
+}
+
+func (s *compatTestSuiteBase) testUpdateKeyPCRProtectionPolicyRevokes(c *C, pcrProfile *secboot.PCRProtectionProfile, pcrEventsFile string) {
+	tmpDir := c.MkDir()
+	keyFile, err := os.Open(s.absPath("key"))
+	c.Assert(err, IsNil)
+	defer keyFile.Close()
+
+	keyFileBackup, err := os.Create(filepath.Join(tmpDir, "key"))
```
cmatsuoka, line 193: What is the purpose of creating this keyfile backup? It seems that it's not used in this test.

internal/compattest/compattest_test.go
```diff
@@ -0,0 +1,238 @@
...
+	key, err := k.UnsealFromTPM(s.TPM, "")
+	c.Check(err, IsNil)
+
+	expectedKey, err := ioutil.ReadFile(s.absPath("clearKey"))
+	c.Assert(err, IsNil)
+
+	c.Check(key, DeepEquals, expectedKey)
+}
+
+func (s *compatTestSuiteBase) testUnsealAfterLock(c *C, pcrEventsFile string) {
```
cmatsuoka, line 228: This is a good test to have :+1:

internal/compattest/compattest_test.go
chrisccoulson: It's loaded a few lines down:

https://github.com/snapcore/secboot/pull/86/files/962267b8cad3a3d0c4019290b60572bee4c1c590#diff-1c58d1fcc056a4f05e5d823f6b670648R204

cmatsuoka: Ah right. Thanks.

PR #92, opened 2020-08-19
Don't use a NV index for PIN integration

chrisccoulson:
The current implementation uses a NV index for PIN integration using
the TPM2_PolicySecret assertion. In the original technical plan, PIN
integration was going to be achieved using a separate TPM object (and
TPM2_PolicySecret), although this was changed to a NV index instead.

There are a couple of reasons for this approach:
- The original plan didn't feature the use of the TPM2_PolicyAuthorize
assertion for allowing the PCR policy to be updated without having to
reseal - instead, updating the PCR policy was going to be achieved by
resealing (ie, creating a new sealed key object). Having a separate
object for PIN integration avoids the need to know the PIN in order to
perform a resealing operation.
- I initially thought that the sealed key object containing the disk
encryption key required the userWithAuth attribute in order to use the
object's auth value in its own authorization policy, and we can't set that
attribute on the sealed key object becuase it would mean it could be
unsealed without satisfying the authorization policy.

This last point isn't correct though - a sealed key object's auth value
can be used in its own authorization policy using the TPM2_PolicyAuthValue
assertion without requiring the userWithAuth attribute (it can't be used
in the authorization policy for another object with TPM2_PolicySecret
though).

The current implementation makes use of the TPM2_PolicyAuthorize assertion
so that the PCR policy can be updated without resealing (using
UpdateKeyPCRProtectionPolicy).

This refactors the PIN integration to no longer depend on a NV index,
although the NV index remains because it serves one other purpose (as an
optional counter for revoking PCR policies). The PIN is now implemented
by setting the auth value of the sealed key object with
TPM2_ObjectChangeAuth. As changing the PIN now involves the recreation
of the private part of the sealed key object, the key data file is now also
serialized through the same AF-splitter used by LUKS for storing binary
keyslot data.

### DIFF
--- a/keydata.go
+++ b/keydata.go
@@ -384,7 +384,7 @@ func (d *keyData) load(tpm *tpm2.TPMContext, session tpm2.SessionContext) (tpm2.
 }
 
 // validate performs some correctness checking on the provided keyData and keyPolicyUpdateData. On success, it returns the validated
-// public area for the dynamic policy counter.
+// public area for the PCR policy counter.
 func (d *keyData) validate(tpm *tpm2.TPMContext, policyUpdateData *keyPolicyUpdateData, session tpm2.SessionContext) (*tpm2.NVPublic, error) {
 	if d.version > currentMetadataVersion {
 		return nil, keyFileError{errors.New("invalid metadata version")}
@@ -442,23 +442,23 @@ func (d *keyData) validate(tpm *tpm2.TPMContext, policyUpdateData *keyPolicyUpda
 		return nil, xerrors.Errorf("cannot compute lock NV index name: %w", err)
 	}
 
-	// Obtain a ResourceContext for the dynamic authorization policy counter. Go-tpm2 calls TPM2_NV_ReadPublic twice here. The second
-	// time is with a session, and there is also verification that the returned public area is for the specified handle so that we know
-	// that the returned ResourceContext corresponds to an actual entity on the TPM at the specified handle. This index is used for
-	// dynamic authorization policy revocation, and also for PIN integration with v0 metadata only.
-	policyCounterHandle := d.staticPolicyData.policyCounterHandle
-	if (policyCounterHandle != tpm2.HandleNull || d.version == 0) && policyCounterHandle.Type() != tpm2.HandleTypeNVIndex {
-		return nil, keyFileError{errors.New("dynamic authorization policy counter handle is invalid")}
+	// Obtain a ResourceContext for the PCR policy counter. Go-tpm2 calls TPM2_NV_ReadPublic twice here. The second time is with a
+	// session, and there is also verification that the returned public area is for the specified handle so that we know that the
+	// returned ResourceContext corresponds to an actual entity on the TPM at the specified handle. This index is used for PCR policy
+	// revocation, and also for PIN integration with v0 metadata only.
+	pcrPolicyCounterHandle := d.staticPolicyData.pcrPolicyCounterHandle
+	if (pcrPolicyCounterHandle != tpm2.HandleNull || d.version == 0) && pcrPolicyCounterHandle.Type() != tpm2.HandleTypeNVIndex {
+		return nil, keyFileError{errors.New("PCR policy counter handle is invalid")}
 	}
 
-	var policyCounter tpm2.ResourceContext
-	if policyCounterHandle != tpm2.HandleNull {
-		policyCounter, err = tpm.CreateResourceContextFromTPM(policyCounterHandle, session.IncludeAttrs(tpm2.AttrAudit))
+	var pcrPolicyCounter tpm2.ResourceContext
+	if pcrPolicyCounterHandle != tpm2.HandleNull {
+		pcrPolicyCounter, err = tpm.CreateResourceContextFromTPM(pcrPolicyCounterHandle, session.IncludeAttrs(tpm2.AttrAudit))
 		if err != nil {
-			if tpm2.IsResourceUnavailableError(err, policyCounterHandle) {
-				return nil, keyFileError{errors.New("dynamic authorization policy counter is unavailable")}
+			if tpm2.IsResourceUnavailableError(err, pcrPolicyCounterHandle) {
+				return nil, keyFileError{errors.New("PCR policy counter is unavailable")}
 			}
-			return nil, xerrors.Errorf("cannot create context for dynamic authorization policy counter: %w", err)
+			return nil, xerrors.Errorf("cannot create context for PCR policy counter: %w", err)
 		}
 	}
 
@@ -478,11 +478,11 @@ func (d *keyData) validate(tpm *tpm2.TPMContext, policyUpdateData *keyPolicyUpda
 		return nil, keyFileError{xerrors.Errorf("cannot determine if static authorization policy matches sealed key object: %w", err)}
 	}
 
-	dynamicPolicyRef := d.staticPolicyData.dynamicPolicyRef
+	pcrPolicyRef := d.staticPolicyData.pcrPolicyRef
 
-	trial.PolicyAuthorize(dynamicPolicyRef, authKeyName)
+	trial.PolicyAuthorize(pcrPolicyRef, authKeyName)
 	if d.version == 0 {
-		trial.PolicySecret(policyCounter.Name(), nil)
+		trial.PolicySecret(pcrPolicyCounter.Name(), nil)
 	} else {
 		// v1 metadata and later
 		trial.PolicyAuthValue()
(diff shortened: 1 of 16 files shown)


chrisccoulson:
@setharnold Hi, you recently took a look at a WIP PR (#61) which changes how PIN integration works. I've recently re-implemented that in another branch which is based on some preliminary changes that are in this PR. Would you mind taking a look at this one too?

keydata.go
```diff
@@ -52,6 +55,92 @@ const (
...
+
+func (d *afSplitDataRaw) Unmarshal(r io.Reader) (nbytes int, err error) {
+	var h afSplitDataRawHdr
+	n, err := tpm2.UnmarshalFromReader(r, &h)
+	nbytes += n
+	if err != nil {
+		return nbytes, xerrors.Errorf("cannot unmarshal header: %w", err)
+	}
+
+	data := make([]byte, h.Size)
```
setharnold, line 81: Are there any reasonable limits that can be placed on this allocation? Allocating four gigabytes may be a surprising operation. Allocating zero bytes here may also lead to a divide-by-zero later on.

keydata.go
```diff
@@ -52,6 +55,92 @@ const (
...
+		return nil, errors.New("invalid number of stripes")
+	}
+	if !d.hashAlg.Supported() {
+		return nil, errors.New("unsupported digest algorithm")
+	}
+	return afis.MergeHash(d.data, int(d.stripes), func() hash.Hash { return d.hashAlg.NewHash() })
+}
+
+func makeAfSplitData(data []byte, minSz int, hashAlg tpm2.HashAlgorithmId) (*afSplitData, error) {
+	stripes := uint32((minSz / len(data)) + 1)
```
setharnold, line 126: Here's the potential divide-by-zero case

keydata.go
```diff
@@ -52,6 +55,92 @@ const (
...
+	nbytes += n
+	if err != nil {
+		return nbytes, xerrors.Errorf("cannot unmarshal header: %w", err)
+	}
+
+	data := make([]byte, h.Size)
+	n, err = io.ReadFull(r, data)
+	nbytes += n
+	if err != nil {
+		return nbytes, xerrors.Errorf("cannot read data: %w", err)
```
setharnold, line 85: What happens to the memory used by `data` in the event this fails?

keydata.go
```diff
@@ -216,6 +327,30 @@ func (d *keyData) Unmarshal(r io.Reader) (nbytes int, err error) {
 			authModeHint:      raw.AuthModeHint,
 			staticPolicyData:  raw.StaticPolicyData.data(),
 			dynamicPolicyData: raw.DynamicPolicyData.data()}
+	case 1:
+		var splitData afSplitDataRaw
+		n, err := tpm2.UnmarshalFromReader(r, &splitData)
+		nbytes += n
+		if err != nil {
+			return nbytes, xerrors.Errorf("cannot unmarshal split data: %w", err)
```
setharnold, line 335: Should these error cases `return` `0` or `nbytes`? The `d.version` code in `keydata.go` appeared to `return 0, ...` on error.

keydata.go
```diff
@@ -175,22 +273,35 @@ func (d *keyData) Marshal(w io.Writer) (nbytes int, err error) {
...
+			AuthModeHint:      d.authModeHint,
+			StaticPolicyData:  makeStaticPolicyDataRaw_v1(d.staticPolicyData),
+			DynamicPolicyData: makeDynamicPolicyDataRaw_v0(d.dynamicPolicyData)}
+		if _, err := tpm2.MarshalToWriter(&tmpW, raw); err != nil {
+			return 0, xerrors.Errorf("cannot marshal data: %w", err)
+		}
+		splitData, err := makeAfSplitData(tmpW.Bytes(), 128*1024, tpm2.HashAlgorithmSHA256)
 		if err != nil {
-			return nbytes, xerrors.Errorf("cannot marshal raw data: %w", err)
+			return 0, xerrors.Errorf("cannot split data: %w", err)
```
setharnold, line 298: Should these error cases `return` `0` or `nbytes`? The `version` code in `keydata.go` appeared to `return nbytes, ...` on error.

policy.go
```diff
@@ -69,61 +67,131 @@ type policyOrDataTree []policyOrDataNode
...
+	var authPolicies tpm2.DigestList
+
+	trial, err := tpm2.ComputeAuthPolicy(alg)
+	if err != nil {
+		return nil, err
+	}
+	trial.PolicyNvWritten(false)
+	authPolicies = append(authPolicies, trial.GetDigest())
+
+	trial, _ = tpm2.ComputeAuthPolicy(alg)
```
setharnold, line 179: Why is the error return discarded in this case?

policy.go
```diff
@@ -494,28 +644,55 @@ func ensureSufficientORDigests(digests tpm2.DigestList) tpm2.DigestList {
...
+		var err error
+		counterName, err = counterPublic.Name()
+		if err != nil {
+			return nil, xerrors.Errorf("cannot compute name of counter: %w", err)
+		}
+	}
+
+	h := alg.NewHash()
+	h.Write([]byte("DYNAMIC-PCR"))
+	h.Write(counterName)
```
setharnold, line 664: Do we need a closing sentinel of some sort to prevent length extension attacks on this field?

unseal_test.go
```diff
@@ -244,7 +250,22 @@ func TestUnsealErrorHandling(t *testing.T) {
...
 		err := run(t, tpm, func(keyFile, policyUpdateFile string) {
-			if err := UpdateKeyPCRProtectionPolicy(tpm, keyFile, policyUpdateFile, getTestPCRProfile()); err != nil {
+			src, err := os.Open(keyFile)
+			if err != nil {
+				t.Fatalf("Open failed: %v", err)
+			}
+			defer src.Close()
+
+			newKeyFile := filepath.Dir(keyFile) + "/newkeydata"
+			dst, err := os.OpenFile(newKeyFile, os.O_WRONLY|os.O_CREATE, 0644)
```
setharnold, line 260: Is this code strictly for test suite use? I suspect so. If not, this should probably use a more-`mkstemp(3)`-ish approach, including `O_EXCL` and `O_NOFOLLOW` to reduce the risk of unexpected hardlinks and symlinks.

setharnold (commented):
With the caveat that I don't really know much about the semantics of the changes, here's my thoughts on what I've read. Thanks

keydata.go
chrisccoulson: There are issues like this in other places too. I've pushed a fix to go-tpm2 to ensure that sizes that are larger than the amount of data remaining are rejected, but that requires an update the vendored go-tpm2.

Zero bytes here will trigger a EOF in keyData.Unmarshal. The call to makeAfSplitData is called during marshalling instead.

keydata.go
chrisccoulson: This shouldn't be possible - it's called during marshalling from keyData.Marshal, and the data length should always be non-zero:

https://github.com/snapcore/secboot/blob/fc07b3864d05363e8d253157164d89c81094f0fb/keydata.go#L296

keydata.go
chrisccoulson: It should eventually be reclaimed by go's garbage collector.

keydata.go
chrisccoulson: The top-level API's in go-tpm2 return the number of bytes read, even in the case of an error (see https://godoc.org/github.com/canonical/go-tpm2/mu#UnmarshalFromReader), which is why nbytes is returned here. The case where 0 is returned is probably a mistake, but I've just pushed some changes to go-tpm2 to avoid the need for this logic to be here at all:

https://github.com/canonical/go-tpm2/commit/bdbb15b0f6937b4f68708bb533482c144c5071e2

keydata.go
chrisccoulson: Yeah, that should probably be nbytes, but this change in go-tpm2 will make this unnecessary:

https://github.com/canonical/go-tpm2/commit/bdbb15b0f6937b4f68708bb533482c144c5071e2

policy.go
chrisccoulson: The only way this can fail is if the algorithm isn't supported by the go runtime, but there's an earlier call to tpm2.ComputeAuthPolicy in this function with the same algorithm where the error is checked. If that succeeded then there isn't a way for this call to fail.

unseal_test.go
chrisccoulson: Yeah, this is just test code.

policy.go
chrisccoulson: We shouldn't do - the field is always the same length (it's a digest preceded by 2 bytes to indicate the algorithm).

keydata.go
setharnold: Oh nice, this feels like a really good cleanup, it removes something that has confused me more than once! Thanks

keydata.go
```diff
@@ -53,6 +56,92 @@ const (
 	AuthModePIN
 )
 
+type afSplitDataRawHdr struct {
+	Stripes uint32
+	HashAlg tpm2.HashAlgorithmId
+	Size    uint32
+}
+
+type afSplitDataRaw struct {
```
pedronis, line 65: this needs a doc comment and an explanation of what is going on (also it should probably been its own PR)

unseal.go
```diff
@@ -140,11 +140,16 @@ func (k *SealedKeyObject) UnsealFromTPM(tpm *TPMConnection, pin string) ([]byte,
 		return nil, err
 	}
 
+	// Required for metadata version > 0
```
pedronis, line 143: it this is a irrelevant for version 0 or needs to be guarded?  The comment should be more of the form

// For metadata version > 0 we set the pin as auth value , that's how we in that case auth is in ....

pin.go
```diff
@@ -79,153 +76,13 @@ func computePinNVIndexPostInitAuthPolicies(alg tpm2.HashAlgorithmId, updateKeyNa
...
-
-	initKeyPublic := createPublicAreaForECDSAKey(&initKey.PublicKey)
-	initKeyName, err := initKeyPublic.Name()
-	if err != nil {
-		return nil, nil, xerrors.Errorf("cannot compute name of signing key for initializing NV index: %w", err)
-	}
-
-	nameAlg := tpm2.HashAlgorithmSHA256
-
-	// The NV index requires 5 policies:
```
pedronis, line 111: we should keep this comment somewhere close to v0 metadata otherwise it's hard to follow version == 0 auth code, part of it is kept implemented by computeV0PinNVIndexPostInitAuthPolicies? but that is used only validate so it's a bit too indirect

policy.go
```diff
@@ -72,61 +69,128 @@ type policyOrDataTree []policyOrDataNode
...
 
-// incrementDynamicPolicyCounter will increment the NV counter index associated with nvPublic. This is designed to operate on a
-// NV index created by createPinNVIndex. The authorization policy digests returned from createPinNVIndex must be supplied via the
-// nvAuthPolicies argument.
+// computePcrPolicyCounterAuthPolicies computes the authorization policy digests passed to TPM2_PolicyOR for a PCR
+// policy counter that can be updated with the key associated with updateKeyName.
+func computePcrPolicyCounterAuthPolicies(alg tpm2.HashAlgorithmId, updateKeyName tpm2.Name) (tpm2.DigestList, error) {
+	// The NV index requires 2 policies:
+	// - A policy to initialize the index with no authorization
+	// - A policy for updating the index to revoke old PCR policies, using a signed assertion.
```
pedronis, line 168: these are much simpler than the v0 policies,   a comment here like  

// These are simpler than v0 because ...

would be useful

policy.go
```diff
@@ -192,12 +256,20 @@ func incrementDynamicPolicyCounter(tpm *tpm2.TPMContext, nvPublic *tpm2.NVPublic
 	defer tpm.FlushContext(keyLoaded)
 
 	// Execute the policy assertions
-	if err := tpm.PolicyCommandCode(policySession, tpm2.CommandNVIncrement); err != nil {
-		return xerrors.Errorf("cannot execute assertion to increment counter: %w", err)
-	}
-	if err := tpm.PolicyNvWritten(policySession, true); err != nil {
-		return xerrors.Errorf("cannot execute assertion to increment counter: %w", err)
+	if version == 0 {
+		if err := tpm.PolicyCommandCode(policySession, tpm2.CommandNVIncrement); err != nil {
```
pedronis, line 260: this should have a comment pointing to the place where we keep the v0 policies comment

policy.go
```diff
@@ -213,34 +285,108 @@ func incrementDynamicPolicyCounter(tpm *tpm2.TPMContext, nvPublic *tpm2.NVPublic
...
 	index, err := tpm2.CreateNVIndexResourceContextFromPublic(nvPublic)
 	if err != nil {
 		return 0, xerrors.Errorf("cannot create context for NV index: %w", err)
 	}
 
-	policySession, err := tpm.StartAuthSession(nil, nil, tpm2.SessionTypePolicy, nil, nvPublic.NameAlg)
+	var authSession tpm2.SessionContext
+	var extraSession tpm2.SessionContext
+	if version == 0 {
+		authSession, err = tpm.StartAuthSession(nil, nil, tpm2.SessionTypePolicy, nil, nvPublic.NameAlg)
```
pedronis, line 301: same

policy.go
```diff
@@ -784,95 +960,118 @@ func executePolicySession(tpm *tpm2.TPMContext, policySession tpm2.SessionContex
...
+			return staticPolicyDataError{errors.New("public area of dynamic authorization policy signing key is invalid")}
 		}
-		return xerrors.Errorf("cannot load public area for dynamic authorization policy signature verification key: %w", err)
+		return xerrors.Errorf("cannot load public area for dynamic authorization policy signing key: %w", err)
 	}
 	defer tpm.FlushContext(authorizeKey)
 
+	var pcrPolicyRef tpm2.Nonce
+	if version > 0 {
+		pcrPolicyRef = computePcrPolicyRefFromCounterContext(policyCounter)
```
pedronis, line 1040: this is probably a good place to repeat the reasons/purpose of policyRef

policy.go
```diff
@@ -784,95 +960,118 @@ func executePolicySession(tpm *tpm2.TPMContext, policySession tpm2.SessionContex
...
-	pinIndex.SetAuthValue([]byte(pin))
-	if _, _, err := tpm.PolicySecret(pinIndex, policySession, nil, nil, 0, hmacSession); err != nil {
-		return xerrors.Errorf("cannot execute PolicySecret assertion: %w", err)
+	if version == 0 {
+		policyCounter.SetAuthValue([]byte(pin))
+		if _, _, err := tpm.PolicySecret(policyCounter, policySession, nil, nil, 0, hmacSession); err != nil {
+			return xerrors.Errorf("cannot execute PolicySecret assertion: %w", err)
+		}
+	} else {
+		if err := tpm.PolicyAuthValue(policySession); err != nil {
```
pedronis, line 1072: this corresponds to https://github.com/snapcore/secboot/pull/92/files#diff-9c53128bbf22cd8275aacdbf0ae82035R726 ?

pedronis (commented):
did a pass, some comments and questions.  I don't have a clear suggestion but the code, especially in policy.go in fairly hard to follow, I suppose it's relates to the fact that the auth logic is so different because the working principle is quite different. Maybe splitting out v0 code into helpers like pin.go does a bit more would help, maybe...

keydata.go
chrisccoulson: I've added a comment for this now.

unseal.go
chrisccoulson: I've updated this. It doesn't need to be guarded for version 0 because without the TPM2_PolicyAuthValue assertion, the pin doesn't get included in the session HMAC and so this is a noop (see https://github.com/canonical/go-tpm2/blob/master/cmds_ea.go#L366, https://github.com/canonical/go-tpm2/blob/master/auth.go#L300 and https://github.com/canonical/go-tpm2/blob/master/auth.go#L142

pin.go
chrisccoulson: Would it be ok to just reference the comment for computeV0PinNVIndexPostInitAuthPolicies from other places?

policy.go
chrisccoulson: I've added this.

policy.go
chrisccoulson: I've added a comment pointing to the description above computeV0PinNVIndexPostInitAuthPolicies.

policy.go
chrisccoulson: I've added a comment pointing to the description above computeV0PinNVIndexPostInitAuthPolicies, with some additional context about the policy being required to read the counter without knowing the auth value.

policy.go
chrisccoulson: I've added this.

policy.go
chrisccoulson: I've added a comment here.

setharnold:
What motivated the change from rsapss to ecdsa?

Thanks

chrisccoulson:
> What motivated the change from rsapss to ecdsa?
> 
> Thanks

This was done in https://github.com/snapcore/secboot/pull/65 - the motivation being that it's quicker to create the keys and they're significantly smaller. There's a follow up PR which stores the key used to sign PCR policy updates inside the TPM sealed key instead of storing it on disk inside the encrypted volume, now that they key is much smaller.

pin.go
pedronis: given you have been very explicit about that now, seems ok

policy.go
```diff
@@ -529,28 +685,60 @@ func ensureSufficientORDigests(digests tpm2.DigestList) tpm2.DigestList {
...
+func computePcrPolicyRefFromCounterContext(context tpm2.ResourceContext) tpm2.Nonce {
+	var name tpm2.Name
+	if context != nil {
+		name = context.Name()
+	}
+
+	return computePcrPolicyRefFromCounterName(name)
+}
+
 // computeStaticPolicy computes the part of an authorization policy that is bound to a sealed key object and never changes.
```
pedronis, line 717: I didn't remark this before, but I think it would make sense to describe what these static policies are concretely

policy.go
```diff
@@ -615,13 +803,6 @@ func computePolicyORData(alg tpm2.HashAlgorithmId, trial *tpm2.TrialAuthPolicy,
 // computeDynamicPolicy computes the part of an authorization policy associated with a sealed key object that can change and be
 // updated.
```
pedronis, line 804: could also have a short description that is the OR of the pcr values plus a check on the optional counter being <=

pedronis (approved):
my main comment is that as some other compute*Policy functions have descriptions of the policies they setup computeStaticPolicy should have some as well. computeDynamicPolicy could also use a bit of explanation...

pedronis:
@chrisccoulson thanks for the added comments

PR #93, opened 2020-08-24
Update the vendored version of go-tpm2

chrisccoulson:
This simplifies the CustomMarshaller interface so that implementations
don't need to keep track of the number of bytes read or written.


pedronis (approved):
lgtm

PR #95, opened 2020-08-25
Move some functions to SealedKeyObject

chrisccoulson:
SealedKeyObject was originally intended to provide some read-only interaction with a sealed key file to avoid having to decode it more than once. There are some additional functions in the global namespace that modify a keyfile (ChangePIN and UpdateKeyPCRProtectionPolicy).

In order to create a tpm2 KeyData platform implementation, the functions that interact with TPM sealed keys need to be updated to work with bytes without having to care about how the bytes are stored. The current functions have a hard dependency on file storage because file paths are exposed via the APIs.

As a prerequisite to removing the hard dependency on file storage, the existing global functions that modify a keyfile should be converted to methods of SealedKeyObject. Future PRs will then work towards removing the file dependency, and adding in some helper functions for reading / writing existing key files.

For additional context of where things are going, see https://github.com/snapcore/secboot/pull/154, https://github.com/snapcore/secboot/pull/158, https://github.com/snapcore/secboot/pull/159 and https://github.com/snapcore/secboot/pull/161, which get to the point of being able to unlock volumes
using the new secboot.ActivateVolumeWithKeyData API with existing TPM key files - the next step
after this is to make the code in tpm2/ a proper KeyData implementation.

(I'll rebase this on https://github.com/snapcore/secboot/pull/160 once it lands so that the tests pass)

### DIFF
--- a/internal/compattest/compattest_test.go
+++ b/internal/compattest/compattest_test.go
@@ -183,19 +183,24 @@ func (s *compatTestSuiteBase) TestChangePIN(c *C) {
 	c.Assert(err, IsNil)
 	c.Check(k.AuthMode2F(), Equals, secboot.AuthModeNone)
 
-	c.Check(secboot_tpm2.ChangePIN(s.TPM, s.absPath("key"), "", testPIN), IsNil)
+	c.Check(k.ChangePIN(s.TPM, "", testPIN), IsNil)
+	c.Check(k.AuthMode2F(), Equals, secboot.AuthModePassphrase)
 	k, err = secboot_tpm2.ReadSealedKeyObject(s.absPath("key"))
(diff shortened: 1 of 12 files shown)


pedronis:
I'm marking this blocked, the change is innocent enough but is not super clear to me whether you imagine mixing TPM concerns with LUKS concerns or yet something else. ATM we are mixing reading/writing files vs the actual manipulation of the bytes as meaningful for the TPM. There is an approach in which we mix deciding to write files or something else per key storage backend with the low-level bytes manipulations and another where we keep those orthogonal.

pedronis (commented):
this looks alright to me considering the description but the tests are currently failing

pedronis (approved):
thanks

PR #96, opened 2020-09-01
Use LUKS tokens to identify keyslots

chrisccoulson:
This makes several changes to the code that manages LUKS containers:

- A new internal package is created that exports APIs for low-level cryptsetup operations, and is consumed by the code in crypt.go.
- The new internal package contains a LUKS metadata decoder for decoding the LUKS metadata without having to depend on parsing the output of "cryptsetup luksDump". The cryptsetup binary isn't shipped in the initramfs anyway which would make it difficult to consume the metadata during early boot, and we might want this in the future in order to export tokens containing embedded key metadata.
- Adds tokens to identify the master and recovery keyslots.
- Renames AddRecoveryKeyToLUKS2Container to SetLUKS2ContainerRecoveryKey and updates its behaviour accordingly - the new function allows a single recovery key to be set using any other key, and removes the previous recovery key if it was set.
- Renames ChangeLUKS2KeyUsingRecoveryKey to SetLUKS2ContainerMasterKey and updates its behaviour accordingly - the new function allows a single "master" key (where "master" in this context is the key that is protected by a hardware-backed keystore) to be set using any other key, and removes the previous master key if it was set.
- Makes transactions that span multiple cryptsetup operations atomic by implementing the cryptsetup locking in the calling process and then calling cryptsetup with its own locking disabled. This will make it safer for snapd to perform maintenance operations without clashing with an administrator that might be using cryptsetup to make changes to the same container. An example might be calling SetLUKS2ContainerMasterKey after a recovery scenario - here we decode the current metadata, find the ID of the token corresponding to the existing keyslot, identify the existing keyslot ID from the token, add a new keyslot, decode the metadata again, identify the ID of the newly added keyslot, import a new token for the new keyslot, delete the old token using its ID and then remove the old keyslot using its ID. If an administrator uses cryptsetup separately to add another keyslot at just the wrong moment, then we might identify and tag the wrong keyslot with our token.

### DIFF
--- a/internal/luks2/luks2.go
+++ b/internal/luks2/luks2.go
@@ -625,37 +625,37 @@ func getHash(alg string) crypto.Hash {
 	}
 }
 
-func decodeAndCheckHeader(r io.ReadSeeker, offset int64, primary bool) (*binaryHdr, error) {
+func decodeAndCheckHeader(r io.ReadSeeker, offset int64, primary bool) (*binaryHdr, []byte, error) {
 	if _, err := r.Seek(offset, io.SeekStart); err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	var hdr binaryHdr
 	if err := binary.Read(r, binary.BigEndian, &hdr); err != nil {
-		return nil, xerrors.Errorf("cannot read header: %w", err)
+		return nil, nil, xerrors.Errorf("cannot read header: %w", err)
 	}
 	switch {
 	case primary && bytes.Equal(hdr.Magic[:], []byte("LUKS\xba\xbe")):
 	case !primary && bytes.Equal(hdr.Magic[:], []byte("SKUL\xba\xbe")):
 	default:
-		return nil, errors.New("invalid magic")
+		return nil, nil, errors.New("invalid magic")
 	}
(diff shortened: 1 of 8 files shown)


internal/luks2/luks2.go
```diff
@@ -0,0 +1,968 @@
...
+	}
+
+	if !bytes.Equal(h.Sum(nil), hdr.Csum[0:csumHash.Size()]) {
+		return nil, nil, errors.New("invalid header checksum")
+	}
+
+	return &hdr, jsonBuffer.Bytes(), nil
+}
+
+func DecodeHdr(path string) (*HdrInfo, error) {
```
pedronis, line 697: To make this landable, I think this would need to be submitted first with its own tests and with pointers to upstream code/docs as its own PR. Then the rest of luks2 as its own PR, then finally its usage from crypt.go etc.

pedronis (changes requested):
this is too big right now and under explained vs upstream, I made a suggestion on how to proceed to accelerate trying to land this, it involves splitting this at least in 3 PRs.

chrisccoulson:
This can be closed as it's been superseded by https://github.com/snapcore/secboot/pull/197

PR #97, opened 2020-09-07
Store the private part of the policy auth key inside the sealed key object

chrisccoulson:
Now that an elliptic curve key is used to authorize PCR policy updates and
this is significantly smaller than the previous RSA key, store the private
part of the key inside the sealed key object and have it returned from
SealedKeyObject.UnsealFromTPM rather than storing it inside a file inside
the encrypted volume.

This removes the need to store an additional file in order to use
UpdateKeyPCRProtectionPolicy.

ActivateVolumeWithTPMSealedKey now adds the private part of the key to
the root user's keyring. Callers to UpdateKeyPCRProtectionPolicy will
need to retrieve the key data from there using the keyctl syscall.

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -304,7 +304,7 @@ func activateWithRecoveryKey(volumeName, sourceDevicePath string, keyReader io.R
 			continue
 		}
 
-		if _, err := unix.AddKey("user", fmt.Sprintf("%s:%s:reason=%d", filepath.Base(os.Args[0]), volumeName, reason), key[:], userKeyring); err != nil {
+		if _, err := unix.AddKey("user", fmt.Sprintf("secboot-recover:%s?reason=%d", sourceDevicePath, reason), key[:], userKeyring); err != nil {
 			lastErr = xerrors.Errorf("cannot add recovery key to user keyring: %w", err)
 		}
 		break
@@ -408,7 +408,7 @@ func activateWithTPMKey(tpm *TPMConnection, volumeName, sourceDevicePath, keyPat
 
 	// Ignore errors - we've activated the volume and so we shouldn't return an error at this point unless we
 	// close the volume again.
-	unix.AddKey("user", fmt.Sprintf("%s:%s:auth", filepath.Base(os.Args[0]), volumeName), authPrivateKey, userKeyring)
+	unix.AddKey("user", fmt.Sprintf("secboot-tpmauth:%s", sourceDevicePath), authPrivateKey, userKeyring)
 
 	return nil
 }
@@ -462,7 +462,7 @@ type ActivateWithTPMSealedKeyOptions struct {
 // how many attempts should be made to activate the volume with the recovery key before failing. If this is set to 0, then no attempts
 // will be made to activate the encrypted volume with the fallback recovery key. If activation with the recovery key is successful,
 // the recovery key will be added to the root user keyring in the kernel with a description of the format
-// "<argv[0]>:<volumeName>:reason=<reason>" where reason is an integer that describes the recovery reason - see the
+// "secboot-recover:<sourceDevicePath>?reason=<reason>" where reason is an integer that describes the recovery reason - see the
 // RecoveryKeyUsageReason type.
 //
 // If either the PINTries or RecoveryKeyTries fields of options are less than zero, an error will be returned. If the ActivateOptions
@@ -478,7 +478,7 @@ type ActivateWithTPMSealedKeyOptions struct {
 //
 // If the volume is successfully activated with the TPM sealed key and the TPM sealed key has a version of greater than 1, the
 // private part of the key used for authorizing PCR policy updates with UpdateKeyPCRProtectionPolicy is added to the root user keyring
-// in the kernel with a description of the format "<argv[0]>:<volumeName>:auth".
+// in the kernel with a description of the format "secboot-tpmauth:<sourceDevicePath>".
 //
 // If the volume is successfully activated, either with the TPM sealed key or the fallback recovery key, this function returns true.
 // If it is not successfully activated, then this function returns false.
@@ -542,7 +542,7 @@ type ActivateWithRecoveryKeyOptions struct {
 // The ActivateOptions field of options can be used to specify additional options to pass to systemd-cryptsetup.
 //
 // If activation with the recovery key is successful, the recovery key will be added to the root user keyring in the kernel with a
-// description of the format "<argv[0]>:<volumeName>:reason=2".
+// description of the format "secboot-recover:<sourceDevicePath>?reason=2".
 //
 // If the Tries field of options is less than zero, an error will be returned. If the ActivateOptions field of options contains the
 // "tries=" option, then an error will be returned. This option cannot be used with this function.
--- a/crypt.go
+++ b/crypt.go
@@ -319,7 +319,7 @@ func activateWithRecoveryKey(volumeName, sourceDevicePath string, keyReader io.R
 		// Possessor Set Attribute / Possessor Link / Possessor Search / Possessor Write / Possessor Read / Possessor View / User View.
 		// Possessor permissions only apply to a process with a searchable link to the key from one of its own keyrings - just having the
 		// same UID is not sufficient. Read permission is required to read the contents of the key (view permission only permits viewing
-		// of the description and other public metadata that isn't the key payload).//
+		// of the description and other public metadata that isn't the key payload).
 		//
 		// Note that by default, systemd starts services with a private session keyring which does not contain a link to the user keyring.
 		// Therefore these services cannot access the contents of keys in the root user's user keyring if those keys only permit
@@ -431,7 +431,7 @@ func activateWithTPMKey(tpm *TPMConnection, volumeName, sourceDevicePath, keyPat
 	// Possessor Set Attribute / Possessor Link / Possessor Search / Possessor Write / Possessor Read / Possessor View / User View.
 	// Possessor permissions only apply to a process with a searchable link to the key from one of its own keyrings - just having the
 	// same UID is not sufficient. Read permission is required to read the contents of the key (view permission only permits viewing
-	// of the description and other public metadata that isn't the key payload).//
+	// of the description and other public metadata that isn't the key payload).
 	//
 	// Note that by default, systemd starts services with a private session keyring which does not contain a link to the user keyring.
 	// Therefore these services cannot access the contents of keys in the root user's user keyring if those keys only permit
--- a/crypt.go
+++ b/crypt.go
@@ -313,8 +313,8 @@ func activateWithRecoveryKey(volumeName, sourceDevicePath string, keyReader io.R
 	return lastErr
 }
 
-func unsealKeyFromTPM(tpm *TPMConnection, k *SealedKeyObject, pin string) ([]byte, error) {
-	key, err := k.UnsealFromTPM(tpm, pin)
+func unsealKeyFromTPM(tpm *TPMConnection, k *SealedKeyObject, pin string) ([]byte, []byte, error) {
+	sealedKey, authPrivateKey, err := k.UnsealFromTPM(tpm, pin)
 	if err == ErrTPMProvisioning {
 		// ErrTPMProvisioning in this context might indicate that there isn't a valid persistent SRK. Have a go at creating one now and then
 		// retrying the unseal operation - if the previous SRK was evicted, the TPM owner hasn't changed and the storage hierarchy still
@@ -322,10 +322,10 @@ func unsealKeyFromTPM(tpm *TPMConnection, k *SealedKeyObject, pin string) ([]byt
 		// storage hierarchy has a non-null authorization value, ProvionTPM will fail. If the TPM owner has changed, ProvisionTPM might
 		// succeed, but UnsealFromTPM will fail with InvalidKeyFileError when retried.
 		if pErr := ProvisionTPM(tpm, ProvisionModeWithoutLockout, nil); pErr == nil {
-			key, err = k.UnsealFromTPM(tpm, pin)
+			sealedKey, authPrivateKey, err = k.UnsealFromTPM(tpm, pin)
 		}
 	}
-	return key, err
+	return sealedKey, authPrivateKey, err
 }
 
 var requiresPinErr = errors.New("no PIN tries permitted when a PIN is required")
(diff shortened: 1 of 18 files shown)


unseal.go
```diff
@@ -68,16 +68,17 @@ import (
 // condition can also occur as the result of an incorrectly provisioned TPM, which will be detected during a subsequent call to
 // SealKeyToTPM.
 //
-// On success, the unsealed cleartext key is returned.
-func (k *SealedKeyObject) UnsealFromTPM(tpm *TPMConnection, pin string) ([]byte, error) {
+// On success, the unsealed cleartext key is returned as the first return value, and the private part of the key used for
+// authorizing PCR policy updates with UpdateKeyPCRProtectionPolicy is returned as the second return value.
+func (k *SealedKeyObject) UnsealFromTPM(tpm *TPMConnection, pin string) ([]byte, []byte, error) {
```
pedronis, line 73: now probably this should have named return arguments for clarity

seal.go
```diff
@@ -155,10 +150,15 @@ type KeyCreationParams struct {
 //
 // The key will be protected with a PCR policy computed from the PCRProtectionProfile supplied via the PCRProfile field of the params
 // argument.
-func SealKeyToTPM(tpm *TPMConnection, key []byte, keyPath, policyUpdatePath string, params *KeyCreationParams) error {
+//
+// On success, this function returns the private part of the key used for authorizing PCR policy updates with
+// UpdateKeyPCRProtectionPolicy. This key doesn't need to be stored anywhere, and certainly mustn't be stored outside of the encrypted
+// volume protected with this sealed key file. The key is stored encrypted inside this sealed key file and returned from future calls
+// to SealedKeyObject.UnsealFromTPM.
+func SealKeyToTPM(tpm *TPMConnection, key []byte, keyPath string, params *KeyCreationParams) ([]byte, error) {
```
pedronis, line 158: even if there's less ambiguity it's probably a good idea to have named return arguments also here

seal.go
```diff
@@ -272,32 +256,25 @@ func SealKeyToTPM(tpm *TPMConnection, key []byte, keyPath, policyUpdatePath stri
...
-	}
-	creationInfo := h.Sum(nil)
+	sensitive := tpm2.SensitiveCreate{Data: sealedData}
 
 	// Now create the sealed key object. The command is integrity protected so if the object at the handle we expect the SRK to reside
 	// at has a different name (ie, if we're connected via a resource manager and somebody swapped the object with another one), this
 	// command will fail. We take advantage of parameter encryption here too.
-	priv, pub, creationData, _, creationTicket, err :=
-		tpm.Create(srk, &sensitive, template, creationInfo, nil, session.IncludeAttrs(tpm2.AttrCommandEncrypt))
+	priv, pub, _, _, _, err := tpm.Create(srk, &sensitive, template, nil, nil, session.IncludeAttrs(tpm2.AttrCommandEncrypt))
```
pedronis, line 275: can you clarify why we can ignore creationTicket and creationData now?

keydata.go
```diff
@@ -590,60 +532,25 @@ func (d *keyData) validate(tpm *tpm2.TPMContext, policyUpdateData *keyPolicyUpda
...
-		return nil, keyFileError{errors.New("mismatched metadata versions")}
-	}
-
-	// Verify that the private data structure is bound to the key data structure.
-	h := keyPublic.NameAlg.NewHash()
-	if _, err := tpm2.MarshalToWriter(h, policyUpdateData.creationData); err != nil {
-		panic(fmt.Sprintf("cannot marshal creation data: %v", err))
-	}
-
-	if _, _, err := tpm.CertifyCreation(nil, keyContext, nil, h.Sum(nil), nil, policyUpdateData.creationTicket, nil,
```
pedronis, line 608: are we losing integrity/security without this when we still use v0 update files?

pedronis (changes requested):
The changes look straightforward enough all things considered, some questions though. The main sticking point is that we need more support and regain some control for the caller in crypt.go. It seems I didn't look closely enough at the similar code we have for the recovery key  usage case...

keydata.go
chrisccoulson: This was only there for integrity checking of the file containing the authorization key, in case it grew to contain additional fields. It's not required for checking the actual authorization key - there's already a check to ensure that the public part of it matches the authorization policy of the sealed key, and go's crypto library performs validation that the public and private parts are consistent when it is loaded. Since the file is being removed, there's not much point keeping this additional complexity around.

seal.go
chrisccoulson: We were only using the creation ticket and creation data to cryptographically bind the file containing the authorization key to the sealed key object for the purposes of verifying later on that they were consistent in keyData.validate, but as that has been removed then it's ok to ignore these now.

keydata.go
pedronis: thanks for the clarification on this

crypt.go
```diff
@@ -401,10 +402,14 @@ func activateWithTPMKey(tpm *TPMConnection, volumeName, sourceDevicePath, keyPat
...
 	}
 
-	if err := activate(volumeName, sourceDevicePath, key, activateOptions); err != nil {
+	if err := activate(volumeName, sourceDevicePath, sealedKey, activateOptions); err != nil {
 		return xerrors.Errorf("cannot activate volume: %w", err)
 	}
 
+	// Ignore errors - we've activated the volume and so we shouldn't return an error at this point unless we
+	// close the volume again.
+	unix.AddKey("user", fmt.Sprintf("%s:%s:auth", filepath.Base(os.Args[0]), volumeName), authPrivateKey, userKeyring)
```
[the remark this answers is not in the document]
pedronis: oops, didn't quote \< properly

unseal.go
chrisccoulson: I've added named return arguments now, but I've also added a TPMPolicyAuthKey type which probably makes this unnecessary.

crypt.go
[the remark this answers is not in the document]
chrisccoulson: I think it makes sense for having a function to retrieve data added to the kernel keyring, although I'm not sure that the API needs to be quite so complicated. I've added a "GetActivationDataFromKernel(string)" which just takes a device path and returns a piece of data with a type dependent on how the volume was activated - all of the parsing of the key description happens internally. I'm not sure what to call the function - I didn't want to use "key" in it because it returns more data than just a key - the returned type indicates how the device was unlocked and the type of key, and in the recovery case it also returns the recovery reason.

What would be the use case for having the ability to override writing the key to the kernel?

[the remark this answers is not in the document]
pedronis: As I said, at the moment is not even possible to log problems or control how the key is named in the kernel. Also while most snaps don't have keyring access some might (a few interfaces grant that), so it's unclear whether putting keys there is a generally viable strategy, because keyctl is not very mediation friendly atm AFAIR.

crypt.go
```diff
@@ -304,7 +304,7 @@ func activateWithRecoveryKey(volumeName, sourceDevicePath string, keyReader io.R
 			continue
 		}
 
-		if _, err := unix.AddKey("user", fmt.Sprintf("%s:%s:reason=%d", filepath.Base(os.Args[0]), volumeName, reason), key[:], userKeyring); err != nil {
+		if _, err := unix.AddKey("user", fmt.Sprintf("secboot-recover:%s?reason=%d", sourceDevicePath, reason), key[:], userKeyring); err != nil {
```
pedronis: I'm happy this moved away from os.Args[0], but I still think we want a way to let the caller control these prefixes. I still think we need to have a callback mechanism for this.

crypt.go
```diff
@@ -304,7 +305,7 @@ func activateWithRecoveryKey(volumeName, sourceDevicePath string, keyReader io.R
 			continue
 		}
 
-		if _, err := unix.AddKey("user", fmt.Sprintf("%s:%s:reason=%d", filepath.Base(os.Args[0]), volumeName, reason), key[:], userKeyring); err != nil {
+		if _, err := unix.AddKey("user", fmt.Sprintf("secboot:%s?type=recovery&reason=%d", sourceDevicePath, reason), key[:], userKeyring); err != nil {
```
pedronis, line 308: this looks much better, but the caller still needs the options of:

* choosing the prefix part of this (secboot vs something more caller specific)
* replace this entirely with something else

crypt.go
```diff
@@ -401,10 +403,14 @@ func activateWithTPMKey(tpm *TPMConnection, volumeName, sourceDevicePath, keyPat
...
 	}
 
-	if err := activate(volumeName, sourceDevicePath, key, activateOptions); err != nil {
+	if err := activate(volumeName, sourceDevicePath, sealedKey, activateOptions); err != nil {
 		return xerrors.Errorf("cannot activate volume: %w", err)
 	}
 
+	// Ignore errors - we've activated the volume and so we shouldn't return an error at this point unless we
+	// close the volume again.
+	unix.AddKey("user", fmt.Sprintf("secboot:%s?type=tpm", sourceDevicePath), authPrivateKey, userKeyring)
```
pedronis, line 412: same considerations

crypt.go
```diff
@@ -550,6 +559,114 @@ func ActivateVolumeWithRecoveryKey(volumeName, sourceDevicePath string, keyReade
...
+				userKeys = append(userKeys, int(binary.LittleEndian.Uint32(payload)))
+				payload = payload[4:]
+			}
+			break
+		}
+
+		sz = n
+	}
+
+	re := regexp.MustCompile(`^user;[[:digit:]]+;[[:digit:]]+;[[:xdigit:]]+;secboot:([^\?]+)\??(.*)`)
```
pedronis, line 603: if we let the caller control the prefix as we should, this should at least take that in. It can default to "" => "secboot"

pedronis (changes requested):
this is better naming wise and structuring wise for what is put in the keyring but the issues essentially around prefix/naming, logging and control remain

crypt.go
```diff
@@ -521,6 +562,9 @@ type ActivateWithRecoveryKeyOptions struct {
 
 	// ActivateOptions provides a mechanism to pass additional options to systemd-cryptsetup.
 	ActivateOptions []string
+
+	// KeyringPrefix is the prefix used for the description of any kernel keys created during activation.
+	KeyringPrefix string
```
pedronis, line 567: thanks for this

crypt.go
```diff
@@ -547,7 +591,123 @@ func ActivateVolumeWithRecoveryKey(volumeName, sourceDevicePath string, keyReade
...
+}
+
+// GetActivationDataFromKernel retrieves data that was added to the current user's user keyring by ActivateVolumeWithTPMSealedKey or
+// ActivateVolumeWithRecoveryKey for the specified source block device, using the prefix that was passed to either of those functions.
+// The block device path must match the path passed to one of the ActivateVolume functions. The type of data returned is dependent on
+// how the volume was activated - see the documentation for each function, If no data is found for the specified device, a
+// ErrNoActivationData error is returned.
+//
+// If remove is true, this function will unlink the key from the user's user keyring.
+func GetActivationDataFromKernel(prefix, sourceDevicePath string, remove bool) (ActivationData, error) {
```
pedronis, line 613: this should probably return the kind of key/situtation and the key id but not read the key itself such that the caller can decide to read the key and unlink or link it to a different keyring before unlinking...

crypt.go
```diff
@@ -547,7 +591,123 @@ func ActivateVolumeWithRecoveryKey(volumeName, sourceDevicePath string, keyReade
...
+		payload := make([]byte, sz)
+		_, err = unix.KeyctlBuffer(unix.KEYCTL_READ, id, payload, 0)
+		if err != nil {
+			return nil, xerrors.Errorf("cannot read key payload: %w", err)
+		}
+
+		if remove {
+			// XXX: What should we do if unlinking fails?
+			unix.KeyctlInt(unix.KEYCTL_UNLINK, id, userKeyring, 0, 0)
+		}
```
pedronis, line 668: this bit of the code seem simple enough for the caller to deal with it as price for more control

crypt.go
```diff
@@ -401,10 +422,24 @@ func activateWithTPMKey(tpm *TPMConnection, volumeName, sourceDevicePath, keyPat
...
+	// Possessor permissions only apply to a process with a searchable link to the key from one of its own keyrings - just having the
+	// same UID is not sufficient. Read permission is required to read the contents of the key (view permission only permits viewing
+	// of the description and other public metadata that isn't the key payload).
+	//
+	// Note that by default, systemd starts services with a private session keyring which does not contain a link to the user keyring.
+	// Therefore these services cannot access the contents of keys in the root user's user keyring if those keys only permit
+	// possessor-read.
+	//
+	// Ignore errors - we've activated the volume and so we shouldn't return an error at this point unless we close the volume again.
+	unix.AddKey("user", fmt.Sprintf("%s:%s?type=tpm", keyringPrefixOrDefault(keyringPrefix), sourceDevicePath), authPrivateKey, userKeyring)
```
pedronis, line 441: should "tpm" here be more "tpm-policy-auth" ?

pedronis (commented):
thanks for the changes,  some more considerations on the API of  GetActivationDataFromKernel

seal.go
```diff
@@ -213,31 +213,15 @@ func SealKeyToTPM(tpm *TPMConnection, key []byte, keyPath, policyUpdatePath stri
...
-			os.Remove(policyUpdatePath)
-		}()
-	}
-
 	// Create an asymmetric key for signing authorization policy updates, and authorizing dynamic authorization policy revocations.
-	authKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	goAuthKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
 	if err != nil {
-		return xerrors.Errorf("cannot generate RSA key pair for signing dynamic authorization policies: %w", err)
+		return nil, xerrors.Errorf("cannot generate RSA key pair for signing dynamic authorization policies: %w", err)
```
pedronis, line 219: I didn't notice this in the previous PR, this still mentions RSA

crypt.go
pedronis: we discussed this and it seems the pattern would be to transfer the key by creating a new key in the other keyring,  there are complexities though around using the process keyring so we might not be able to do that for a while anway.

crypt.go
pedronis: given the work in #111 is not clear whether this should be more generic like "reseal" or be specific, we can discuss there

pedronis (approved):
let's merge this and go from there

PR #98, opened 2020-09-07
Switch to use Container.RandomAccessFile's result Size

pedronis:
make the EFIImage interface follow that as well

### DIFF
--- a/efi.go
+++ b/efi.go
@@ -88,6 +88,7 @@ func (p FileEFIImage) Open() (interface {
 	}
 	fi, err := f.Stat()
 	if err != nil {
+		f.Close()
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


efi.go
```diff
@@ -104,13 +80,17 @@ func (p FileEFIImage) String() string {
...
+	Size() int64
 }, error) {
 	f, err := os.Open(string(p))
 	if err != nil {
 		return nil, err
 	}
-	return &fileEFIImageHandle{f}, nil
+	fi, err := f.Stat()
+	if err != nil {
+		return nil, err
```
chrisccoulson, line 91: The open file handle should probably be closed here.

chrisccoulson:
Thanks, this looks mostly ok - I've left a comment about a file handle leak in an error path though.

PR #99, opened 2020-09-07
Be explicit via interface SnapModel which data from the model is measured

pedronis:


mvo5 (approved):
Thanks for this PR, looks fine.

snapmodel_policy.go
```diff
@@ -38,7 +38,16 @@ func computeSnapSystemEpochDigest(alg tpm2.HashAlgorithmId, epoch uint32) tpm2.D
 	return h.Sum(nil)
 }
 
-func computeSnapModelDigest(alg tpm2.HashAlgorithmId, model *asserts.Model) (tpm2.Digest, error) {
+// Model exposes the details of a device model that are to be measured.
+type Model interface {
```
chrisccoulson, line 42: As this is in the secboot namespace, it might be worth prefixing this with "Snap" (so, SnapModel) to match the other parts of the API (MeasureSnapModelToTPM, SnapModelProfileParams etc).

chrisccoulson (approved):
Thanks, I've just left a minor comment

PR #102, opened 2020-09-10
Use the PIN to derive additional key material

chrisccoulson:
The default configuration of having no PIN set protects data from
opportunistic attackers. Setting a PIN protects data from more
determined attackers. This is the case even if the PIN is fairly low
entropy, because we use the TPM to assert that the provided PIN is
correct and this benefits from the TPM's dictionary attack logic. It is
beyond the ability even for most determined attackers to compromise the
actual TPM hardware to defeat this mechanism.

However, this may not be sufficient to protect your data from 3 letter
agencies with the means to compromise the TPM hardware in a way that
allows them access to a shielded location without the use of a protected
capability (a function that operates on a shielded location) - the type
of thing that would only be possible by state actors. Setting a higher
entropy PIN or passhprase doesn't help here, because the TPM just asserts
that the supplied PIN or passphrase is correct - it doesn't provide any
additional key material required for unsealing the disk encryption key.
The PIN is stored inside the sealed key file, encrypted to the TPM's
storage hierarchy. All of the key material required to unlock an encrypted
volume lives on the device.

In the future, we may want to support a network unlocking feature, where a
trusted server can work with the existing TPM protected key object to
authorize disk unlocking as long as the device is physically attached to
a specific network. An implementation of this should require that the
trusted server contributes key material required to unlock an encrypted
volume, but this doesn't fit well with the current design.

Some use cases for network unlocking are:
- Preventing automatic unlocking of devices that are detached from a
specific network (ie, if they are stolen).
- Bypassing the requirement for a PIN in enterprise environments when
devices are physically attached to a corporate network (where a PIN is
required when the device isn't physically attached to it).

This commit changes how the PIN is used to protect a key in order to
improve this.

During SealKeyToTPM, rather than sealing the disk encryption key (DEK) to
the TPM, it now:
- Generates an intermediate key (IK).
- Encrypts DEK with IK and an initialization vector (DEKiv) to produce
DEKenc.
- Seals DEKenc to the TPM.
- Saves DEKiv and IK to the metadata of the sealed key object.

During unsealing when a PIN hasn't been set, the following happens:
- The TPM unseals DEKenc if the PCR values match the PCR policy.
- DEKiv and IK are used to decrypt DEKenc to recover DEK.

When a PIN is set, the following happens:
- The PIN and a salt are passed through Argon2i to create a derived key
(DK).
- The upper 32-bits of DK are used for the authorization value of the
sealed key object, which is updated accordingly.
- The lower 32-bits of DK are used with an initialization vector (IKiv) to
encrypt IK to produce IKenc.
- IKenc is written to the sealed key object metadata, replacing the
unencrypted version.
- The salt and IKiv are also written to the sealed key object metadata.

Unsealing with a PIN looks like this:
- The supplied PIN and a salt are passed through Argon2i to create a
derived key (DK).
- The TPM unseals DEKenc if the PCR values match the PCR policy and the
upper 32-bits of DK match the authorization value of the sealed key
object.
- The lower 32-bits of DK are used along with IKiv to decrypt IKenc in
order to recover IK.
- DEKiv and IK are used to decrypt DEKenc to recover DEK.

This scheme retains the property that low-entropy PINs provide some
protection of data against most determined attackers, because it still
relies on the TPM asserting knowledge of the correct PIN before unsealing
the encrypted DEK, and this benefits from the TPM's dictionary attack
protection.

This scheme should be more suitable for incorporating a network unlock
feature in the future, where IK and the the TPM auth value can be
protected with the public key of a trusted device on the local network.

### DIFF
--- a/crypt_test.go
+++ b/crypt_test.go
@@ -387,7 +387,7 @@ func (s *cryptTPMSuite) testActivateVolumeWithTPMSealedKeyAndPIN(c *C, data *tes
 func (s *cryptTPMSuite) TestActivateVolumeWithTPMSealedKeyAndPIN1(c *C) {
 	// Test with a single PIN attempt.
 	testPIN := "1234"
-	c.Assert(ChangePIN(s.TPM, s.keyFile, "", testPIN), IsNil)
+	c.Assert(ChangePIN(s.TPM, s.keyFile, &testPINParams, "", testPIN), IsNil)
 	s.testActivateVolumeWithTPMSealedKeyAndPIN(c, &testActivateVolumeWithTPMSealedKeyAndPINData{
 		pins:     []string{testPIN},
 		pinTries: 1,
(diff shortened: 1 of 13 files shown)


chrisccoulson:
@setharnold - this is an updated version of a PR you looked at a while ago (https://github.com/snapcore/secboot/pull/61)

chrisccoulson:
This can be closed now as it was superseded by https://github.com/snapcore/secboot/pull/188

PR #104, opened 2020-09-11
Remove ProvisionStatus and refactor provision API

chrisccoulson:
ProvisionStatus is a problematic API - AttrValidSRK and AttrValidEK
indicate that objects with the expected public areas exist at the expected
handles, but can't indicate whether those objects were created with the
correct templates and are actually valid objects. These attributes can't
be used to make a decision about whether ProvisionTPM should be called or
not, so just remove the ProvisionStatus API entirely.

Rename ProvisionTPM to EnsureProvisioned (and make it a method of
TPMConnection), and introduce a new error (ErrTPMProvisioningRequiresLockout)
which is returned from EnsureProvisioned if it is called with
ProvisionModeWithoutLockout but use of the lockout hierarchy is required
to fully provision the TPM. This new error is an indication that
EnsureProvisioned needs to be called again with a different mode in order
to fully provision it.

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -321,7 +321,7 @@ func unsealKeyFromTPM(tpm *TPMConnection, k *SealedKeyObject, pin string) ([]byt
 		// has a null authorization value, then this will allow us to unseal the key without requiring any type of manual recovery. If the
 		// storage hierarchy has a non-null authorization value, ProvionTPM will fail. If the TPM owner has changed, ProvisionTPM might
 		// succeed, but UnsealFromTPM will fail with InvalidKeyFileError when retried.
-		if pErr := ProvisionTPM(tpm, ProvisionModeWithoutLockout, nil); pErr == nil {
+		if pErr := tpm.EnsureProvisioned(ProvisionModeWithoutLockout, nil); pErr == nil || pErr == ErrTPMProvisioningRequiresLockout {
 			key, err = k.UnsealFromTPM(tpm, pin)
 		}
 	}
--- a/provisioning_test.go
+++ b/provisioning_test.go
@@ -250,6 +250,11 @@ func TestProvisionErrorHandling(t *testing.T) {
 			t.Fatalf("HierarchyChangeAuth failed: %v", err)
 		}
 	}
+	disableOwnerClear := func(t *testing.T) {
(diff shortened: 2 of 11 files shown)


provisioning_test.go
```diff
@@ -475,21 +508,30 @@ func TestRecreateSRK(t *testing.T) {
...
 				t.Errorf("EvictControl failed: %v", err)
 			}
 
-			if err := ProvisionTPM(tpm, data.mode, lockoutAuth); err != nil {
-				t.Fatalf("ProvisionTPM failed: %v", err)
+			if err := tpm.EnsureProvisioned(data.mode, lockoutAuth); err != nil {
+				t.Fatalf("EnsureProvisioned failed: %v", err)
+			}
+
+			srk, err = tpm.CreateResourceContextFromTPM(tcg.SRKHandle)
```
pedronis, line 529: this is an additional test now?

crypt.go
```diff
@@ -321,7 +321,7 @@ func unsealKeyFromTPM(tpm *TPMConnection, k *SealedKeyObject, pin string) ([]byt
 		// has a null authorization value, then this will allow us to unseal the key without requiring any type of manual recovery. If the
 		// storage hierarchy has a non-null authorization value, ProvionTPM will fail. If the TPM owner has changed, ProvisionTPM might
 		// succeed, but UnsealFromTPM will fail with InvalidKeyFileError when retried.
-		if pErr := ProvisionTPM(tpm, ProvisionModeWithoutLockout, nil); pErr == nil {
```
pedronis, line 324: the comment should maybe be expanded to cover the new || condition

pedronis (approved):
thanks, couple small comments

crypt.go
chrisccoulson: I've actually changed this in https://github.com/snapcore/secboot/pull/112 to just ignore the error instead, as there's not really any harm in attempting to unseal again even if provisioning fails with an error.

provisioning_test.go
chrisccoulson: It is - it verifies that the TPM hasn't been cleared by making sure that the new primary key is identical to the original one.

PR #106, opened 2020-09-15
Reimplement the support for LockAccessToSealedKeys to reduce fragility

chrisccoulson:
Access to sealed keys can be denied until the next TPM reset or restart
by calling LockAccessToSealedKeys. The traditional way of achieving this
is by extending another value to a PCR that is included in the PCR
selection for a sealed key's authorization policy after the key has been
unsealed, which makes its authorization policy non-satisfiable until the
next reset or restart. The issue with this is that it makes the feature
dependent on the PCR profile.

The current implementation of LockAccessToSealedKeys uses a NV index at a
well known handle and takes advantage of a couple of properties of NV
index read locks:
- Once enabled, they can only be disabled by a TPM reset or restart.
- Enabling a read lock sets the index's TPMA_NV_READLOCKED attribute which
changes its name.

Sealed keys created by this package have an authorization policy that
asserts that this NV index exists at its well known handle. Calling
LockAccessToSealedKeys enables the read lock for this index, which changes
its name and makes this assertion fail.

The problem with this approach though is that the TPM owner (ie, anyone
who knows the authorization value for the storage hierarchy, which is
empty by default) can undefine and redefine a NV index in order to clear
the read lock attribute, thus re-enabling access to sealed keys. One
approach to prevent this from happening relies on the fact that writing to
a NV index for the first time sets the TPMA_NV_WRITTEN attribute which
also changes its name. If the index is defined with the
TPMA_NV_POLICY_WRITE attribute and an authorization policy that can only
be satisfied by an assertion signed with an ephemeral key (using
TPM2_PolicySigned), and sealed keys have authorization policies that can
only be satisfied by the presence of the initialized index, then it is not
possible for an adversary without the private part of the ephemeral key to
undefine and redefine an identical NV index in order to remove the
TPMA_NV_READLOCKED attribute and reenable access to sealed keys. This is
the approach currently taken by secboot, but this has its own problem - if
the NV index is accidentally undefined, then this would make all sealed
keys permanently unrecoverable. Ideally, provisioning the TPM without
clearing it should be able to remedy any accidental changes and restore
access to valid sealed keys in all cases except where the storage primary
seed has been changed (ie, the TPM has been cleared).

This changes the approach used for protecting the NV index used by
LockAccessToSealedKeys. The new approach makes use of 2 NV indices and has
the property that the NV indices have names that are common to all devices
and can always be recreated, but they have authorization policies that
enforce a creation sequence that means that they can only be created in
the locked state. The way this works is as follows:
- A bootstrap index is created at handle 2 with TPMA_NV_AUTHREAD and
TPMA_NV_AUTHWRITE attributes and an empty authorization policy.
- An empty write is performed to initialize the bootstrap index.
- An index is created at handle 1 with TPMA_NV_AUTHREAD,
TPMA_NV_POLICY_WRITE and TPMA_NV_READ_STCLEAR attributes, and an
authorization policy that can only be satisfied by the presence of the
initialized bootstrap index.
- An empty write is performed to initialize index 1 using a policy session.
- The bootstrap index is undefined and another index is created at handle
2 with TPMA_NV_AUTHREAD and TPMA_NV_POLICY_WRITE attributes, and an
authorization policy that can only be satisfied by the presence of the
initialized index at handle 1 with its read lock enabled.
- The read lock for the index at handle 1 is enabled.
- An empty write is performed to initialize index 2 using a policy session.

Authorization policies for sealed keys must assert the presence of both of
these NV indices using the names of the initialized indices without the
read lock enabled. Enabling the read lock on index 1 will disable access
to these keys.

It isn't possible to recreate these indices in their unlocked state.
Although they can be recovered if they are undefined accidentally, they
will not return to their unlocked state until the next TPM reset or
restart.

An adversary can't undefine and redefine either index in order to
reenable access to sealed keys. Eg, say an adversary performs the
following actions:
- They undefine and redefine index 1. The new index isn't read locked, but
sealed keys can't be unsealed until the new index has been written to. But
it has an authorization policy that requires index 2 to be undefined and
replaced with the bootstrap index.
- They undefine index 2 and create the boostrap index in its place.
- They perform an empty write to index 1 to initialize it. Index 1 now has
the expected name, but sealed keys still can't be accessed because index 2
no longer exists.
- They undefine the bootstrap index again and redefine index 2. However,
sealed keys still can't be unsealed until the new index has been written
to. But it has an authorization policy that requires the read lock to be
enabled on index 1.
- They enable the read lock on index 1.
- They perform an empty write to index 2 to initialize it. Index 2 now has
the expected name, but sealed keys still can't be accessed because index 1
is read locked. This can ony be undone by performing a TPM reset or
restart.

### DIFF
--- a/crypt_test.go
+++ b/crypt_test.go
@@ -235,48 +235,41 @@ func (ctb *cryptTestBase) checkRecoveryKeyKeyringEntry(c *C, reason RecoveryKeyU
 	c.Check(buf, DeepEquals, ctb.recoveryKey)
 }
 
-type cryptTPMTestBase struct {
+type cryptTPMSimulatorSuite struct {
+	testutil.TPMSimulatorTestBase
 	cryptTestBase
 
 	keyFile string
 }
 
-func (ctb *cryptTPMTestBase) setUpTestBase(c *C, ttb *testutil.TPMTestBase) {
-	ctb.cryptTestBase.setUpTestBase(c, &ttb.BaseTest)
+var _ = Suite(&cryptTPMSimulatorSuite{})
 
-	c.Assert(ProvisionTPM(ttb.TPM, ProvisionModeFull, nil), IsNil)
+func (s *cryptTPMSimulatorSuite) SetUpSuite(c *C) {
+	s.cryptTestBase.setUpSuiteBase(c)
+}
+
+func (s *cryptTPMSimulatorSuite) SetUpTest(c *C) {
+	s.TPMSimulatorTestBase.SetUpTest(c)
+	s.cryptTestBase.setUpTestBase(c, &s.BaseTest)
+
+	c.Assert(ProvisionTPM(s.TPM, ProvisionModeFull, nil), IsNil)
+	s.ResetTPMSimulator(c)
 
 	dir := c.MkDir()
-	ctb.keyFile = dir + "/keydata"
+	s.keyFile = dir + "/keydata"
 
 	pcrPolicyCounterHandle := tpm2.Handle(0x0181fff0)
-	c.Assert(SealKeyToTPM(ttb.TPM, ctb.tpmKey, ctb.keyFile, "", &KeyCreationParams{PCRProfile: getTestPCRProfile(), PCRPolicyCounterHandle: pcrPolicyCounterHandle}), IsNil)
-	pinIndex, err := ttb.TPM.CreateResourceContextFromTPM(pcrPolicyCounterHandle)
+	c.Assert(SealKeyToTPM(s.TPM, s.tpmKey, s.keyFile, "", &KeyCreationParams{PCRProfile: getTestPCRProfile(), PCRPolicyCounterHandle: pcrPolicyCounterHandle}), IsNil)
+	pinIndex, err := s.TPM.CreateResourceContextFromTPM(pcrPolicyCounterHandle)
 	c.Assert(err, IsNil)
-	ttb.AddCleanupNVSpace(c, ttb.TPM.OwnerHandleContext(), pinIndex)
+	s.AddCleanupNVSpace(c, s.TPM.OwnerHandleContext(), pinIndex)
 
-	c.Assert(ioutil.WriteFile(ctb.expectedTpmKeyFile, ctb.tpmKey, 0644), IsNil)
+	c.Assert(ioutil.WriteFile(s.expectedTpmKeyFile, s.tpmKey, 0644), IsNil)
 
 	// Some tests may increment the DA lockout counter
-	ttb.AddCleanup(func() {
-		c.Check(ttb.TPM.DictionaryAttackLockReset(ttb.TPM.LockoutHandleContext(), nil), IsNil)
+	s.AddCleanup(func() {
+		c.Check(s.TPM.DictionaryAttackLockReset(s.TPM.LockoutHandleContext(), nil), IsNil)
 	})
-}
-
-type cryptTPMSuite struct {
-	testutil.TPMTestBase
-	cryptTPMTestBase
-}
-
-var _ = Suite(&cryptTPMSuite{})
-
-func (s *cryptTPMSuite) SetUpSuite(c *C) {
-	s.cryptTPMTestBase.setUpSuiteBase(c)
-}
-
-func (s *cryptTPMSuite) SetUpTest(c *C) {
-	s.TPMTestBase.SetUpTest(c)
-	s.cryptTPMTestBase.setUpTestBase(c, &s.TPMTestBase)
 }
 
 type testActivateVolumeWithTPMSealedKeyNo2FAData struct {
@@ -287,7 +280,7 @@ type testActivateVolumeWithTPMSealedKeyNo2FAData struct {
 	activateOptions  []string
 }
 
-func (s *cryptTPMSuite) testActivateVolumeWithTPMSealedKeyNo2FA(c *C, data *testActivateVolumeWithTPMSealedKeyNo2FAData) {
+func (s *cryptTPMSimulatorSuite) testActivateVolumeWithTPMSealedKeyNo2FA(c *C, data *testActivateVolumeWithTPMSealedKeyNo2FAData) {
 	options := ActivateWithTPMSealedKeyOptions{PINTries: data.pinTries, RecoveryKeyTries: data.recoveryKeyTries, ActivateOptions: data.activateOptions}
 	success, err := ActivateVolumeWithTPMSealedKey(s.TPM, data.volumeName, data.sourceDevicePath, s.keyFile, nil, &options)
 	c.Check(success, Equals, true)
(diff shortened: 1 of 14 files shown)


dodys (approved):
LGTM

policy.go
```diff
@@ -399,218 +393,252 @@ func createPcrPolicyCounter(tpm *tpm2.TPMContext, handle tpm2.Handle, updateKeyN
...
+// indices without the read lock enabled. Enabling the read lock on index 1 by calling LockAccessToSealedKeys will disable access to
+// these keys.
+//
+// It isn't possible to recreate these indices in their unlocked state. Although they can be recovered if they are undefined
+// accidentally, they will not return to their unlocked state until the next TPM reset or restart.
+//
+// An adversary can't undefine and redefine either index in order to reenable access to sealed keys. Eg, say an adversary performs the
+// following actions:
+// - They undefine and redefine index 1. The new index isn't read locked, but sealed keys can't be unsealed until the new index has
+//   been written to. But it has an authorization policy that requires index 2 to be undefined and replaced with the bootstrap index.
```
pedronis, line 520: how does the binding of policy to indexes work?  from reading this sentence it sounds like it works in such a way that policy survives undefine/redefine of the index?

chrisccoulson: I'm not sure I understand the question, but if you're referring to the policy of sealed keys, then they are bound to the indices using the names of them, which are dependent on the indices own authorization policy and public attributes. If these indices are undefined, it's possible to recover and initialize them with the same authorization policy as before, but they won't have attributes (and therefore, names) that permit unsealing of keys until after a restart.

policy.go
```diff
@@ -399,218 +393,252 @@ func createPcrPolicyCounter(tpm *tpm2.TPMContext, handle tpm2.Handle, updateKeyN
...
+// - They undefine index 2 and create the boostrap index in its place.
+// - They perform an empty write to index 1 to initialize it. Index 1 now has the expected name, but sealed keys still can't be
+//   accessed because index 2 no longer exists.
+// - They undefine the bootstrap index again and redefine index 2. However, sealed keys still can't be unsealed until the new index
+//   has been written to. But it has an authorization policy that requires the read lock to be enabled on index 1.
+// - They enable the read lock on index 1.
+// - They perform an empty write to index 2 to initialize it. Index 2 now has the expected name, but sealed keys still can't be
+//   accessed because index 1 is read locked. This can ony be undone by performing a TPM reset or restart.
```
pedronis, line 528: is there are a test that matches this hypothetical attack? also this explanation should probably emphasize names even more.  I start to be a bit worried about these long explanation before functions that are kind of then hard to match up with code because the code is sometimes spread over helpers in sometimes asymmetric ways.

policy.go
pedronis: thx

crypt_test.go
```diff
@@ -239,49 +239,58 @@ func (ctb *cryptTestBase) checkRecoveryActivationData(c *C, prefix, path string,
...
 
 	keyFile        string
 	authPrivateKey TPMPolicyAuthKey
 }
 
-func (ctb *cryptTPMTestBase) setUpTestBase(c *C, ttb *testutil.TPMTestBase) {
-	ctb.cryptTestBase.setUpTestBase(c, &ttb.BaseTest)
+var _ = Suite(&cryptTPMSimulatorSuite{})
+
+func (s *cryptTPMSimulatorSuite) SetUpSuite(c *C) {
```
pedronis, line 252: can't these noisy changes be done separately and similar other ones? It's generally a bad idea to mix lots of small cosmetic changes and important changes together.

pedronis (changes requested):
The mixing of what looks like small cosmetic changes with spread out important changes plus backward compat support is making very hard to review this PR.

policy.go
chrisccoulson: I've added a test case for this.

pedronis:
@chrisccoulson can you expand on this argument a bit more?

> The issue with this is that it makes the feature
dependent on the PCR profile.

What's your real worry here:

* that you need to know which PCRs the policy depends on to do the scrambling, that sounds annoying but solvable to me?

* that we will have policies that don't depend on PCRs at all? is that possible? it would mean no check on the software that is running but trusting it, that sounds strange at face value

I'm asking this because however clever this is, it's quite a bit of delicate code to carry forward

chrisccoulson:
> @chrisccoulson can you expand on this argument a bit more?
> 
> > The issue with this is that it makes the feature
> > dependent on the PCR profile.
> 
> What's your real worry here:
> 
> * that one needs to know which PCRs the policy depends on to do the scrambling, that sounds annoying but solvable to me?
> * that we will have policies that don't depend on PCRs at all? is that possible? it would mean no check on the software that is running but trusting it, that sounds strange at face value
> 
> I'm asking this because however clever this is, it's quite a bit of delicate code to carry forward. I'll also notice that this is the 2nd implementation of this, will we need a 3rd?

The mechanism is not designed to protect the key that you need to unseal to boot the current OS, but to protect keys that you don't need and may not even be aware of in order to boot the current OS - eg, say you boot from external media on a device that has an internal encrypted drive, or you insert and boot from an additional internal drive on a device that already has an encrypted install. The feature is designed to protect the contents of an encrypted drive belonging to someone else from being accessed by an adversary that just boots the same OS from another drive on a device, but where the OS provided by the adversary is configured to permit some form of shell access with their own credentials.

In order to do this, it needs to work universally and it needs to work without being dependent on accessing key data for the keys it is meant to protect. Given that there could be products with slightly different PCR profiles (eg, non-EFI), it's not really possible to make the feature work universally and without being dependent on accessing key data if it relies on PCRs.

pedronis:
@chrisccoulson thanks for the explanation,  that probably should be incorporated in the comment for ensureLockNVIndices.

This also has conflicts from my changes to unify the ActivateVolume*Options.

We should merge master back into v1-wip as well.

policy.go
```diff
@@ -638,43 +682,102 @@ func readAndValidateLockNVIndexPublic(tpm *tpm2.TPMContext, index tpm2.ResourceC
...
+		return errors.New("found indices with unexpected names")
+	}
+
+	return nil
+}
+
+// validateLockNVIndices checks that the NV indices at the global handles used for locking access to sealed keys are valid lock
+// indices, and returns an error if they aren't. If the indices correspond to the legacy mechanism, the public area of the lock index
+// is returned, else no public area is returned.
+func validateLockNVIndices(tpm *tpm2.TPMContext, session tpm2.SessionContext) (*tpm2.NVPublic, error) {
```
pedronis, line 754: I understand it somehow minimized the diff, because OTOH the code base is not very internally modular ATM, but it's very awkward as a signature to carry into the future that this returns the legacy lock index public area or not.

I have an idea for an immediate follow-up, that would solve maybe that and also explore patterns to make the code a bit more internally modular.

pedronis (commented):
I'll try an immediate follow-up refactor to try to solve some awkwardness here, and also increase a bit my confidence into my own understanding of this.

policy.go
pedronis: @chrisccoulson I proposed the follow-up  https://github.com/snapcore/secboot/pull/118, it's in draft because some things are best moved once this has conflict fixed and the doc comment for ensureLockNVIndices updated. See last commit there: 8c62bf263ed3b451

Issue #107, opened 2020-09-16
UC20 can't be installed on Dell Embedded Box PC 3000 due to TPM error


woodrow-shen:
When trying to install UC20 on Dell Embedded Box PC 3000 it encounters the TPM error messages “cannot identify initial OS launch verification event: boot manager image load event occurred without a preceding verification event".

woodrow-shen:
Thanks @chrisccoulson and @pedronis I will verify again.

PR #108, opened 2020-09-16
Make AddEFISecureBootPolicyProfile less strict

chrisccoulson:
AddEFISecureBootPolicyProfile tries to determine the
EV_EFI_VARIABLE_AUTHORITY event in PCR7 associated with the verification
of shim, and whether that event was also used for verification of UEFI
drivers (in this case, the event needs to be preserved in the PCR profile,
even if we create a profile where shim is verified with another
authority). But the code for doing this is overcomplicated and too
strict - it finds the first EV_EFI_BOOT_SERVICES_APPLICATION in PCR4
(which is associated with shim), then assumes that the preceeding
EV_EFI_VARIABLE_AUTHORITY event in PCR7 is the one associated with the
verification of it. If this event in PCR7 occurs before the transition to
OS-present (ie, before the EV_SEPARATOR events in PCRs 0-6) then it
assumes that it was also used for verification of a UEFI driver.

This breaks on firmware that measures the EV_EFI_VARIABLE_AUTHORITY and
EV_EFI_BOOT_SERVICES_APPLICATION events in a different order.

Instead, simplify this by preserving all EV_EFI_VARIABLE_AUTHORITY events
in the generated PCR profile if they occur before the transition to
OS-present. Also handle the case where the EV_SEPARATOR event for PCR7 is
measured as part of the transition to OS-present as opposed to being
measured before the hand off to BDS (most implementations measure this
before the handoff to BDS, which is before any UEFI drivers or system
preparation applications are executed).

Fixes https://github.com/snapcore/secboot/issues/107


secureboot_policy.go
```diff
@@ -830,21 +789,16 @@ func (b *secureBootPolicyGenBranch) processPreOSEvents(events []*tcglog.Event, i
...
 		}
-	}
-
-	if len(events) == 0 {
-		return nil
-	}
 
-	if !initialOSVerificationEvent.measuredInPreOS {
-		return nil
+		if osPresent && seenSecureBootPCRSeparator {
```
pedronis, line 797: I'm sure I am missing something but reading the comment:

// Processing of the list of events stops when transitioning from pre-OS to OS-present (ie, a EV_SEPARATOR event is encountered in a
// PCR < 7).

I would have expected  this to be just the osPresent bit or an || ?   if this is correct as is, then the comment above need some expanding/clarification.

pedronis (commented):
one comment/question

secureboot_policy.go
chrisccoulson: I've updated this to try to make it clearer.

pedronis: thanks for the updated comment

pedronis (approved):
+1

PR #110, opened 2020-09-21
Let optionally provide an authorisation signing key from outside when sealing a … 

pedronis:


seal.go
```diff
@@ -155,6 +159,9 @@ type KeyCreationParams struct {
 // UpdateKeyPCRProtectionPolicy. This key doesn't need to be stored anywhere, and certainly mustn't be stored outside of the encrypted
 // volume protected with this sealed key file. The key is stored encrypted inside this sealed key file and returned from future calls
 // to SealedKeyObject.UnsealFromTPM.
+//
+// The authorization key can also be chosen and provided by setting
+// AuthKey in the params argument.
```
chrisccoulson, line 164: The comment / doc lines don't need to wrap quite so early.

chrisccoulson (commented):
Whilst I don't mind too much about having this API, I'm not sure how useful it is to provide an external key. If you have separate key objects then they have PCR profiles that have to be managed separately (even if they have the same profile), and so I'm not sure if there is a benefit to sharing keys, unless I'm missing something.

Defining the key internally means we can define it with properties we know will work on every device. Go's crypto/elliptic package has an interface to describe elliptic curves, with a few curve implementations provided. The TPM defines elliptic curves using curve ID's, and there is code in secboot to map between Go's curve implementations and TPM ID's where those mappings exist. The NIST P-256 curve is currently hardcoded in secboot because it is the only curve that is mandatory on PC client TPMs that has a corresponding Go implementation, and so this avoids having to do any feature detection on the TPM. This API makes it possible for a caller to specify alternative curves that are optional on PC client TPMs, and whilst this might appear to work correctly (SealKeyToTPM succeeds), it may produce an authorization key that can't be loaded in to the TPM because of a missing elliptic curve implementation.

pedronis:
> Whilst I don't mind too much about having this API, I'm not sure how useful it is to provide an external key. If you have separate key objects then they have PCR profiles that have to be managed separately (even if they have the same profile), and so I'm not sure if there is a benefit to sharing keys, unless I'm missing something.

"manage separately" is true in some sense and not in others. In the end snapd needs to reseal all objects around the same times. There is some convenience if not fundamental win in having one authorization to deal with all of them. Anyway at this point we need to keep our options open to experiment with how to implement things.

I can add some kind of check that the curve is appropriate, the intention is not to choose other curves.

seal.go
pedronis: I never bothered you with this but the current comments in secboot don't match the wrapping conventions of either snapd or the go std lib:  https://github.com/golang/go/blob/master/src/crypto/crypto.go#L165   The lines here tend to be way too long.  While there is no strong preference for short code lines, reasonable wrapping of comments is typical. It's actually hard for me to follow your long line conventions in my setup.

seal.go
```diff
@@ -155,12 +162,19 @@ type KeyCreationParams struct {
...
 func SealKeyToTPM(tpm *TPMConnection, key []byte, keyPath string, params *KeyCreationParams) (authKey TPMPolicyAuthKey, err error) {
 	// params is mandatory.
 	if params == nil {
 		return nil, errors.New("no KeyCreationParams provided")
 	}
 
+	if params.AuthKey != nil && params.AuthKey.Curve != elliptic.P256() {
+		return nil, errors.New("provided AuthKey must be from elliptictic.P256, no other curve is supported")
```
mvo5, line 175: Thanks for adding this

seal_test.go
```diff
@@ -298,4 +316,18 @@ func TestSealKeyToTPMErrorHandling(t *testing.T) {
 			t.Errorf("Unexpected error: %v", err)
 		}
 	})
+
+	t.Run("WrongCurve", func(t *testing.T) {
```
mvo5, line 320: Thanks for adding this test

mvo5 (approved):
This looks good to me, this is one step towards allowing us to work on the ubuntu-save implementation.

PR #114, opened 2020-09-23
Store keydata in luks metadata

chrisccoulson:
This adds the ability to store the TPM sealed key object inside the LUKS metadata.

ReadSealedKeyObject has been renamed to ReadSealedKeyObjectFromFile, and handles decoding a SealedKeyObject from a detached file.

It adds the following new APIs:
- SealLUKS2MasterKeyToTPM which behaves like SealKeyToTPM but instead writes the sealed key data to a LUKS token. Note that currently only one TPM sealed key is supported per device (as is, only one detached master key and only one recovery key). In a future PR, I will propose a "reference" argument to these APIs in order to create and identify more than one keyslot/token of the same type.
- ReadSealedKeyObjectFromLUKS2 which decodes a SealedKeyObject from the metadata area of a LUKS volume.
- ActivateLUKS2Volume which activates a LUKS volume by recovering a key from a supplied keystore using the LUKS metadata. It is designed to be keystore agnostic (it doesn't have to be used with a TPM).

ActivateLUKS2Volume relies on a new abstraction (LUKS2KeyHandler), which decodes a LUKS token and returns a KeyHandle. There should be one of these for each supported key type, and there is currently one implementation (TPM). In the future, we can have an API to allow other packages to register handlers.

There is a new abstraction from recovering keys from a keystore device such as a TPM. This is based around a couple of interfaces:
- KeyHandle, which is an abstraction of a key, and provides an interface to determine if the key requires a PIN (so asking for a PIN doesn't need to be handled by the implementation). There is a TPM implementation of this which just wraps a *SealedKeyObject, although KeyHandle implementations don't have to contain an actual sealed key - just some data about how to recover a key from a KeyStore - eg, imagine a toy-TPM implementation that just stores a key inside a NV index on the TPM - the KeyHandle could just contain a handle so that the KeyStore can read the corresponding key from that TPM index.
- KeyStore, which is an abstraction of a device from which keys can be recovered by providing a KeyHandle. There is a TPM implementation of this which just wraps a *TPMConnection, and which unseals a key using SealedKeyObject.UnsealFromTPM.

LUKS2KeyHandler also provides an interface to encode a type-specific payload to add to the kernel keyring for use by the OS after booting, along with a mechanism to decode this which will be used by GetActivationDataFromKernel.


chrisccoulson:
This is stacked on top of https://github.com/snapcore/secboot/pull/96 and also has https://github.com/snapcore/secboot/pull/95 merged in, so the diff looks a lot bigger than it actually is.

crypt_keystore.go
```diff
@@ -0,0 +1,166 @@
...
+	return &KeyRecoverError{Code: code, err: err}
+}
+
+func (e *KeyRecoverError) Error() string {
+	var prefix string
+	switch e.Code {
+	case KeyRecoverUnexpectedError:
+		prefix = "unexpected error"
+	case KeyRecoverKeyStoreTemporarilyUnavailableError:
+		prefix = "the keystore is temporarily unavailable for recovering keys"
```
pedronis, line 72: can these errors escape to the user? it's a bit unclear whether surfacing the concept of keystore externally in errors makes sense

chrisccoulson:
This can be closed as it's been superseded by https://github.com/snapcore/secboot/pull/198 and https://github.com/snapcore/secboot/pull/199

PR #118, opened 2020-09-29
Express internally new vs legacy approach to keys' locking via an interface

pedronis:
A keysLockSolution let's execute or try the authorization policy facet corresponding to the strategy and indexes in use for sealed keys locking.

Move sealed keys' locking code to keyslock.go.

At the cost of little bit of abstraction this avoids the awkward return value based on the legacy approach and the corresponding ifs. (I also personally like that the code fragments for the relevant policy bits are in their parallel functions).

This is also quite an important aspect of our approach and there's enough code to have it in its own file at this point.


pedronis:
given #119 this is not meaningful anymore, most of the code depending on legacyLockIndexName is gone

PR #119, opened 2020-10-14
Use a PCR fence for preventing access to keys by the OS

chrisccoulson:
The original implementation of LockAccessToSealedKeys relied on toggling the read lock on a single, non-recreatable NV index at a well known location. The goal of this was to make this feature completely independent of PCR profiles, which made sense in the original plan where we were only including a secure boot chain of trust in the PCR policy. If we have products with kernel images that have slightly different properties with regards to PCR allocation and what they measure to the TPM, an API that is agnostic of PCR profiles becomes important - if an adversary boots an OS from external media on a device that has an encrypted partition, we can guarantee that all keys protected by the device's TPM are inaccessible before the initramfs hands over to the external OS.

That design had a weakness - if the NV index were to be accidentally undefined, all keys protected by the TPM would be lost. Ideally, calling EnsureProvisioned() should always restore access to keys except in the case where the TPM has been cleared, but that wasn't possible with the original implementation.

https://github.com/snapcore/secboot/pull/106 implemented the feature in a different way, using 2 NV indices that could be recreated but only in their locked state until the next reboot. The issue with this is that the TPM2_PolicyNV assertion is immediate which opens up a hole that allows the mechanism to be bypassed using a carefully crafted sequence of commands if either index can be created individually with the correct attributes at any point in time. I don't think that it's possible to fix this, which leaves us with a couple of options:

- Go back to the previous mechanism, with its downsides.
- Abandon the idea of having a mechanism that is independent of PCR profiles.

Having thought about this a bit - given that we now protect a key to a specific set of trusted bootloader artefacts and kernel images rather than just the secure boot chain of trust, I'm not sure that this feature actually needs to be PCR profile agnostic anymore:

- We want to protect a key to a specific set of trusted kernel images, using a platform-specific mechanism (PCR4 in the EFI world).
- Those trusted kernel images have specific properties that define their PCR usage and what they measure (we define that the UC20 kernel measures the command line and snap model to PCR12).
- The runtime OS builds a PCR profile that protects a key to those trusted kernel images, and other properties that we know those kernel images measure.

If you have 2 kernels from different products but the same chain of trust - kernel A and kernel B - that have slightly different PCR usage, and you boot kernel B on a device that contains a key that is only recoverable when booting with kernel A, it no longer matters that kernel B doesn't know how to lock access to those keys because the policy will already prevent access from an OS booted with kernel B.

This switches LockAccessToSealedKeys to a PCR-fence style approach, with the intention that the initramfs will insert this fence in to any PCRs it touches during the boot before handing off to the rest of the OS (currently only PCR12 for UC20). This ends up allowing a lot of code to be deleted, even if the old-style locking approach remains supported for backwards compatibility.

Some things I'm not sure about yet:
- LockAccessToSealedKeys doesn't seem appropriately named anymore, given that it doesn't explicitly lock access to keys - it relies on keys having the correct PCR profile.
- ActivateVolumeWithTPMSealedKey has an option to call LockAccessToSealedKeys before returning, which I temporarily commented out whilst working on this and it seems that this feature isn't tested. I've left it commented out for now - adding it back in requires changing ActivateVolumeOptions. Given that snap-bootstrap already calls LockAccessToSealedKeys itself, I wonder whether the extra complexity here is even necessary?


pedronis:
> Some things I'm not sure about yet:

> * LockAccessToSealedKeys doesn't seem appropriately named anymore, given that it doesn't explicitly lock access to keys - it relies on keys having the correct PCR profile.

I agree, it needs name either based on what it does internally, or preferbaly maybe something more generic about blocking policies using the argument PCRs

> * ActivateVolumeWithTPMSealedKey has an option to call LockAccessToSealedKeys before returning, which I temporarily commented out whilst working on this and it seems that this feature isn't tested. I've left it commented out for now - adding it back in requires changing ActivateVolumeOptions. Given that snap-bootstrap already calls LockAccessToSealedKeys itself, I wonder whether the extra complexity here is even necessary?

I would just drop that option

crypt.go
```diff
@@ -375,7 +375,8 @@ func activateWithTPMKey(tpm *TPMConnection, volumeName, sourceDevicePath, keyPat
 			if !lock {
 				return
 			}
-			lockErr = LockAccessToSealedKeys(tpm)
+			// TODO: Figure out what to do with this
+			//lockErr = LockAccessToSealedKeys(tpm)
```
pedronis, line 379: as I said I would drop this,  it would simplify this code and with the new function it's much easier to just to just do this immediately after the relevant activates for the consumer of the library

policy.go
```diff
@@ -1258,33 +840,51 @@ func executePolicySession(tpm *tpm2.TPMContext, policySession tpm2.SessionContex
 //
 // On success, subsequent calls to SealedKeyObject.UnsealFromTPM will fail with a ErrSealedKeyAccessLocked error until the next TPM
 // restart or TPM reset.
-func LockAccessToSealedKeys(tpm *TPMConnection) error {
+//
+// TODO: Update this documentation and maybe rename the function, as it doesn't make much sense now.
+func LockAccessToSealedKeys(tpm *TPMConnection, pcrs []int) error {
```
pedronis, line 845: maybe something along "BlockPCRProtectionsPolicies" or similar ?

policy.go
```diff
@@ -1258,33 +840,51 @@ func executePolicySession(tpm *tpm2.TPMContext, policySession tpm2.SessionContex
...
+		seq, err := tpm.HashSequenceStart(nil, tpm2.HashAlgorithmNull)
+		if err != nil {
+			return xerrors.Errorf("cannot being hash sequence: %w", err)
+		}
+		if _, err := tpm.EventSequenceExecute(tpm.PCRHandleContext(pcr), seq, fence, session, nil); err != nil {
+			return xerrors.Errorf("cannot execute hash sequence: %w", err)
+		}
+	}
+
+	// Handle legacy locking for v0 key files
```
pedronis, line 862: do we strictly need to keep this bit? is it because the old v0 keys likely weren't using the binary digests? the scrambling of the PCRs will affect all keys new and old right?

pedronis (commented):
lgtm overall, did a pass, some comments/suggestions/questions

policy.go
chrisccoulson: Given that we know all of the keys created by this package use the same PCR profile right now, it probably is fine to drop this.

policy.go
chrisccoulson: Yeah, I think this is ok.

pedronis (approved):
thank you

PR #121, opened 2020-10-19
Add the ability to seal keys with shared authorization policies

chrisccoulson:
This introduces the ability to create keys that are related and share
authorization policies, by adding a new RelatedSealedKey field to
KeyCreationParams. If this is initialized to an existing key, then
SealKeyToTPM will seal a new key with an identical authorization
policy after performing validation on the original key data, as
opposed to computing a new authorization policy. Keys that share a
policy also share a PCR policy counter, if one is defined.

This also requires the private part of the signing key for authorizing
policy updates in the form returned by SealKeyToTPM and UnsealFromTPM
(so that it can be included in the new sealed key object), which is
provided by another new field in KeyCreationParams (RelatedAuthKey).

I'm not sure of the naming yet - "related" is probably the best thing
I can think of at the moment.

### DIFF
--- a/seal.go
+++ b/seal.go
@@ -132,6 +132,15 @@ type KeyCreationParams struct {
 	// If set a key from elliptic.P256 must be used,
 	// if not set one is generated.
 	AuthKey *ecdsa.PrivateKey
+
+	// RelatedSealedKey permits the creation of a new sealed key that shares the same authorization
+	// policy and TPM resources as the specified key. If this is supplied, then RelatedAuthKey
+	// must be provided, and PCRProfile, PCRPolicyCounterHandle and AuthKey must not be provided.
+	RelatedSealedKey *SealedKeyObject
+
+	// RelatedAuthKey is the private part of the dynamic authorization policy key associated with
+	// RelatedSealedKey.
+	RelatedAuthKey TPMPolicyAuthKey
 }
 
 // SealKeyToTPM seals the supplied disk encryption key to the storage hierarchy of the TPM. The sealed key object and associated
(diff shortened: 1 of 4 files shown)


seal.go
```diff
@@ -210,65 +233,126 @@ func SealKeyToTPM(tpm *TPMConnection, key []byte, keyPath string, params *KeyCre
...
+		}
+		dynamicPolicyData, err = computeSealedKeyDynamicAuthPolicy(tpm.TPMContext, currentMetadataVersion, template.NameAlg,
+			authPublicKey.NameAlg, goAuthKey, pcrPolicyCounterPub, nil, pcrProfile, session)
+		if err != nil {
+			return nil, xerrors.Errorf("cannot compute dynamic authorization policy: %w", err)
+		}
+
+		// Increment the PCR policy counter for the first time.
+		if pcrPolicyCounterPub != nil {
+			if err := incrementPcrPolicyCounter(tpm.TPMContext, currentMetadataVersion, pcrPolicyCounterPub, nil, goAuthKey, authPublicKey,
```
pedronis, line 347: we are incrementing the counter before the tpm.Create now, not sure it makes a big difference but wanted to point it out. Anyway in the refactoring to a single to wrap many keys it can probably be changed back more easily

seal_test.go
```diff
@@ -54,26 +54,33 @@ func TestSealKeyToTPM(t *testing.T) {
 	key := make([]byte, 64)
 	rand.Read(key)
 
-	run := func(t *testing.T, tpm *TPMConnection, params *KeyCreationParams) (authKeyBytes []byte) {
+	runNoCleanup := func(t *testing.T, tpm *TPMConnection, params *KeyCreationParams) (authKeyBytes []byte, path string, cleanup func()) {
```
pedronis, line 57: I think runDeferredCleanup might be a slightly clearer name, given that this returns a cleanup after all

pedronis (commented):
this looks reasonable notwithstanding our discussion about changing the API, couple small considerations

seal.go
chrisccoulson: It doesn't matter where it is because the counter gets undefined on failure, but it is easier to move it back to the end of the function in the case where related keys are created in one call, so I'll do that.

seal.go
```diff
@@ -197,18 +208,7 @@ func SealKeyToTPM(tpm *TPMConnection, key []byte, keyPath string, params *KeyCre
...
-		return nil, xerrors.Errorf("cannot create key data file: %w", err)
-	}
-	defer func() {
-		keyFile.Close()
-		if succeeded {
-			return
-		}
-		os.Remove(keyPath)
-	}()
+	// Compuate metadata.
```
pedronis, line 211: typo: compute

seal.go
```diff
@@ -314,7 +339,43 @@ func SealKeyToTPM(tpm *TPMConnection, key []byte, keyPath string, params *KeyCre
 	}
 
 	succeeded = true
-	return goAuthKey.D.Bytes(), nil
+	return
```
pedronis, line 342: this function is a bit too long to use implicit return arguments,  I would have explicit:

return authKey, err  

here

pedronis (approved):
lgtm, small comments

PR #122, opened 2020-10-19
Add UpdateKeyPCRProtectionPolicyMultiple

chrisccoulson:
This introduces UpdateKeyPCRProtectionPolicyMultiple, which can update the
PCR policy for multiple related keys (those that share an identical
authorization policy).

Note that the new API follows the style of the existing API (which accepts
file paths) - I'd prefer to change this in the future so that the actual
decoded SealedKeyObject is passed to these instead.

This also adds some simple tests for UpdateKeyPCRProtectionPolicy,
which were entirely missing previously.

### DIFF
--- a/seal.go
+++ b/seal.go
@@ -380,6 +380,10 @@ func SealKeyToTPM(tpm *TPMConnection, key []byte, keyPath string, params *KeyCre
 }
 
 func updateKeyPCRProtectionPolicyCommon(tpm *tpm2.TPMContext, keyPaths []string, authData interface{}, pcrProfile *PCRProtectionProfile, session tpm2.SessionContext) error {
+	if len(keyPaths) == 0 {
+		return errors.New("no key files supplied")
+	}
+
 	var datas []*keyData
 	// Open the primary data file
 	keyFile, err := os.Open(keyPaths[0])
--- a/seal.go
+++ b/seal.go
@@ -20,6 +20,7 @@
 package secboot
 
 import (
+	"bytes"
 	"crypto"
 	"crypto/ecdsa"
 	"crypto/elliptic"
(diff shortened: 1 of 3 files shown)


seal.go
```diff
@@ -378,48 +379,77 @@ func SealKeyToTPM(tpm *TPMConnection, key []byte, keyPath string, params *KeyCre
...
+		data, _, _, err := decodeAndValidateKeyData(tpm, keyFile, nil, session)
+		if err != nil {
+			if isKeyFileError(err) {
+				return InvalidKeyFileError{err.Error() + " (" + p + ")"}
+			}
+			// FIXME: Turn the missing lock NV index in to ErrProvisioning
+			return xerrors.Errorf("cannot read and validate related key data file: %w", err)
+		}
+		// The data is valid and the sealed key shares the same dynamic authorization policy
+		// signing key. Also verify that it has the same static authorization policy.
```
pedronis, line 419: This comment is not super clear to me, at least the order it mentions things is a bit odd. We don't have an explicit check about the signing key, OTOH below we do check the "static authorization policy." which implies the former? I would have expected the emphasis of the comment to be reversed, or I'm missing something

pedronis (commented):
did a first pass, one comment, one question

seal.go
```diff
@@ -378,48 +379,77 @@ func SealKeyToTPM(tpm *TPMConnection, key []byte, keyPath string, params *KeyCre
 	return SealKeyToTPMMultiple(tpm, []*SealKeyRequest{{Key: key, Path: keyPath}}, params)
 }
 
-func updateKeyPCRProtectionPolicyCommon(tpm *tpm2.TPMContext, keyPath string, authData interface{}, pcrProfile *PCRProtectionProfile, session tpm2.SessionContext) error {
-	// Open the key data file
-	keyFile, err := os.Open(keyPath)
+func updateKeyPCRProtectionPolicyCommon(tpm *tpm2.TPMContext, keyPaths []string, authData interface{}, pcrProfile *PCRProtectionProfile, session tpm2.SessionContext) error {
+	var datas []*keyData
```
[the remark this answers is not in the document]
chrisccoulson: I've added a check now.

seal.go
chrisccoulson: The comment is wrong - I wrote it and then changed the code without updating the comment. I've fixed it now.

pedronis (approved):
thanks for the comment update, now it is clearer and matches my understanding

PR #123, opened 2020-10-22
Add helper for activating a LUKS2 volume with explicitly provided key

bboozzoo:
Add a helper for activating a volume with an explicitly provided key.

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -560,10 +560,8 @@ func ActivateVolumeWithRecoveryKey(volumeName, sourceDevicePath string, keyReade
 // If the ActivateOptions field of options contains the "tries=" option, then an
 // error will be returned. This option cannot be used with this function.
 func ActivateVolumeWithKey(volumeName, sourceDevicePath string, key []byte, options *ActivateVolumeOptions) error {
-	if options.RecoveryKeyTries < 0 {
-		return errors.New("invalid RecoveryKeyTries")
-	}
-
+	// do not be more strict about checking options to allow reusing it
+	// across different calls
 	activateOptions, err := makeActivateOptions(options.ActivateOptions)
 	if err != nil {
 		return err
--- a/crypt.go
+++ b/crypt.go
@@ -550,6 +550,28 @@ func ActivateVolumeWithRecoveryKey(volumeName, sourceDevicePath string, keyReade
 	return activateWithRecoveryKey(volumeName, sourceDevicePath, keyReader, options.RecoveryKeyTries, RecoveryKeyUsageReasonRequested, activateOptions, options.KeyringPrefix)
 }
 
+// ActivateVolumeWithKey attempts to activate the LUKS encrypted volume at
+// sourceDevicePath and create a mapping with the name volumeName, using the
+// provided key. This makes use of systemd-cryptsetup.
+//
+// The ActivateOptions field of options can be used to specify additional
+// options to pass to systemd-cryptsetup. All other fields are ignored.
+//
+// If the ActivateOptions field of options contains the "tries=" option, then an
+// error will be returned. This option cannot be used with this function.
+func ActivateVolumeWithKey(volumeName, sourceDevicePath string, key []byte, options *ActivateVolumeOptions) error {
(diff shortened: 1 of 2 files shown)


crypt.go
```diff
@@ -550,6 +550,28 @@ func ActivateVolumeWithRecoveryKey(volumeName, sourceDevicePath string, keyReade
...
+// sourceDevicePath and create a mapping with the name volumeName, using the
+// provided key. This makes use of systemd-cryptsetup.
+//
+// The ActivateOptions field of options can be used to specify additional
+// options to pass to systemd-cryptsetup. All other fields are ignored.
+//
+// If the ActivateOptions field of options contains the "tries=" option, then an
+// error will be returned. This option cannot be used with this function.
+func ActivateVolumeWithKey(volumeName, sourceDevicePath string, key []byte, options *ActivateVolumeOptions) error {
+	if options.RecoveryKeyTries < 0 {
```
pedronis, line 563: this maybe merits a comment like:

// do not be more strict on ActivateVolumeOptions *Tries options to allow reusing the same options across
// different flavors

pedronis (approved):
thanks

crypt.go
bboozzoo: Actually, I should have removed the check completely.

crypt.go
```diff
@@ -550,6 +550,26 @@ func ActivateVolumeWithRecoveryKey(volumeName, sourceDevicePath string, keyReade
 	return activateWithRecoveryKey(volumeName, sourceDevicePath, keyReader, options.RecoveryKeyTries, RecoveryKeyUsageReasonRequested, activateOptions, options.KeyringPrefix)
 }
 
+// ActivateVolumeWithKey attempts to activate the LUKS encrypted volume at
+// sourceDevicePath and create a mapping with the name volumeName, using the
+// provided key. This makes use of systemd-cryptsetup.
+//
+// The ActivateOptions field of options can be used to specify additional
+// options to pass to systemd-cryptsetup. All other fields are ignored.
```
chrisccoulson, line 558: I wonder whether it would be worth just passing a []string for the options instead?

Incidentally, I don't think anything is using these options anyway, and I wonder if there is actually a use-case for it. From looking at the options handled by systemd-cryptsetup, I'm not sure if any of them are actually useful to surface here.

crypt_test.go
```diff
@@ -1212,6 +1212,100 @@ func (s *cryptSuite) TestActivateVolumeWithRecoveryKeyErrorHandling8(c *C) {
...
+
+	expectedKeyDataPath := filepath.Join(c.MkDir(), "expected-key")
+	mockSdCryptsetup := snapd_testutil.MockCommand(c, c.MkDir()+"/systemd-cryptsetup", fmt.Sprintf(`
+set -x
+key=$(xxd -p < "$4")
+if [ ! -f "%[1]s" ]; then echo "test error: expected key data not mocked"; exit 123; fi
+expected="$(xxd -p < "%[1]s")"
+if [ "$key" != "$expected" ]; then
+    exit 5
+fi
```
chrisccoulson, line 1234: This suite already has a mock for systemd-cryptsetup, with 2 expected keys set up by the suite - tpmKey and recoveryKey, although I have a refactoring in another branch that renames tpmKey but it should be ok to use here despite the name. Is it possible to use the existing mock rather than creating a new one?

chrisccoulson (commented):
Thanks, this looks mostly ok - is it possible to use the existing systemd-cryptsetup mock in the tests though? (it's ok if it's not).

crypt.go
bboozzoo: I think that for consistency with other Activate* calls we should stick to passing ActivateVolumeOptions, but that's just my non-binding opinion.

pedronis: I agree with @bboozzoo point at least until this is moved somewhere else. 

@chrisccoulson your comment about options not being useful was it specific to use the key directly or also for the other usages we have atm by the other ActivateVolume* functions?

chrisccoulson (approved):

PR #125, opened 2020-10-26
Add ActivateVolumeWithMultipleTPMSealedKeys

chrisccoulson:
This adds the ability to activate a volume with multiple TPM sealed key objects. The existing API is not just a wrapper around the new API.

I'm not sure what to do about returning errors from this function - I've added a new ActivateWithMultipleTPMSealedKeysError which has the ability to access each of the individual errors, but I'm not sure that it's actually worth surfacing all of this information.

I've renamed RecoveryKeyUsageReason to KeyErrorCode, as its usage is now more associated with why a key failed rather than why the recovery key was used. Where multiple keys are supplied and they all fail, they each have their own error code.

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -568,6 +568,10 @@ type ActivateVolumeOptions struct {
 // If the volume is successfully activated, either with a TPM sealed key or the fallback recovery key, this function returns true.
 // If it is not successfully activated, then this function returns false.
 func ActivateVolumeWithMultipleTPMSealedKeys(tpm *TPMConnection, volumeName, sourceDevicePath string, keyPaths []string, passphraseReader io.Reader, options *ActivateVolumeOptions) (bool, error) {
+	if len(keyPaths) == 0 {
+		return false, errors.New("no key files provided")
+	}
+
 	if options.PassphraseTries < 0 {
 		return false, errors.New("invalid PassphraseTries")
 	}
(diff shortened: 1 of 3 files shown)


pedronis (commented):
this looks reasonable,  error handling can maybe be simplified a bit, also the multi-key error formatting doesn't quite fit with snapd usual patterns.  @chrisccoulson I will push the necessary tweaks myself.

pedronis (approved):
thanks, I also applied the tweaks from my own feedback

crypt.go
```diff
@@ -584,12 +583,7 @@ func ActivateVolumeWithMultipleTPMSealedKeys(tpm *TPMConnection, volumeName, sou
 		return false, err
 	}
 
-	if err := activateWithTPMKeys(tpm, volumeName, sourceDevicePath, keyPaths, passphraseReader, options.PassphraseTries, activateOptions, options.KeyringPrefix); err != nil {
-		errs, ok := err.(activateWithTPMKeysErrors)
-		if !ok {
-			return false, xerrors.Errorf("unexpected error: %w", err)
-		}
-
+	if success, errs := activateWithTPMKeys(tpm, volumeName, sourceDevicePath, keyPaths, passphraseReader, options.PassphraseTries, activateOptions, options.KeyringPrefix); !success {
```
mvo5, line 586: I think the change here is good but the function return still does not feel super natural/go-ish. But no idea what else could be done to improve so +1 :)

mvo5 (approved):
Thank you!

bboozzoo (approved):
LGTM

PR #127, opened 2020-11-05
Update the KDF parameters for the primary key

chrisccoulson:
This updates the KDF parameters for the primary key to not
hardcode minimum values for iterations and memory cost, and instead
specify a benchmark time of 200ms.

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -945,7 +945,7 @@ func InitializeLUKS2Container(devicePath, label string, key []byte, options *Ini
 		// entropy of at least 32 bytes, and increased cost doesn't provide a security benefit because
 		// this key and these settings are already more secure than the recovery key. Increased cost
 		// here only slows down unlocking.
-		"--pbkdf", "argon2i", "--iter-time", "200",
+		"--pbkdf", "argon2i", "--iter-time", "100",
 		// set LUKS2 label
 		"--label", label,
 	}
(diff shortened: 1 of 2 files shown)


xnox:
Should we change classic ubuntu to 200ms too? or should we stay with 2000ms for the classic ubuntu?

mvo5 (approved):
Thank you

chrisccoulson:
> Should we change classic ubuntu to 200ms too? or should we stay with 2000ms for the classic ubuntu?

Oh no, the only reason we change it for UC20 is because we don't have a low entropy passphrase - it definitely shouldn't be changed on classic. The cost parameters were already tuned to their minimum values on UC20 based on the fact that the key we seal is the same entropy as the derived key (64-bytes). This tweak actually increases the cost a bit because the TPM's primary seed is only 32-bytes.

xnox:
> > Should we change classic ubuntu to 200ms too? or should we stay with 2000ms for the classic ubuntu?
> 
> Oh no, the only reason we change it for UC20 is because we don't have a low entropy passphrase - it definitely shouldn't be changed on classic. The cost parameters were already tuned to their minimum values on UC20 based on the fact that the key we seal is the same entropy as the derived key (64-bytes). This tweak actually increases the cost a bit because the TPM's primary seed is only 32-bytes.

Ok, but in the zkey/pkea/paes case the secret key is sealed to the hardware crypto cards master key. And  thus has physical security. Upstream s390-tools then went with forcing pbkdf2 key derivation. I reverted that, and we have multiple issues of out-of-memory now if there are too many volumes opened in parallel.

I am now thinking that instead of reverting and using defaults of argon2i, I should too, for the s390x mainframe use the benchmark time of 100ms.

PR #128, opened 2020-12-10
Add an internal luks2 package

chrisccoulson:
This adds an internal luks2 package with the following features:
- Activation with systemd-cryptsetup
- APIs for modifying LUKS2 volumes, which are simple wrappers around
cryptsetup - supported are formatting, adding and deleting keyslots,
importing and removing tokens and adjusting keyslot priority.
- A metadata decoder, providing a read-only go representation of the
LUKS2 header and metadata. This includes the appropriate shared locking.

This is a bit different to the previous PR that contained these changes:
- The locking API and the ability to do exclusive locking is gone (it turns out it's problematic to call cryptsetup with --disable-locks even if the caller has an exclusive lock, as cryptsetup disables its automatic recovery in this mode). There is still internal shared lock support for metadata decoding.
- I've split the code up in to smaller files - the files are quite self-contained and can mostly be reviewed in isolation. They could potentially be submitted as separate PRs too.


internal/luks2/cryptsetup.go
```diff
@@ -0,0 +1,202 @@
...
+	if output, err := cmd.CombinedOutput(); err != nil {
+		return osutil.OutputErr(output, err)
+	}
+	return nil
+}
+
+// KillSlot erases the keyslot with the supplied slot number from the specified LUKS2 container.
+// Note that a valid key for a remaining keyslot must be supplied, in order to prevent the last
+// keyslot from being erased.
+func KillSlot(devicePath string, slot int, key []byte) error {
```
pedronis, line 184: shouldn't this be called KillKeyslot for consistency ?

pedronis (commented):
I started but didn't finish to review the non metadata bits of this. Posting what I have so far before the weekend.

internal/luks2/luks2.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"golang.org/x/sys/unix"
+	"golang.org/x/xerrors"
+)
+
+var (
+	runDir = "/run"
+)
+
+func mkFifo() (string, func(), error) {
```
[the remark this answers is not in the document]
chrisccoulson: Sure, I can do that

internal/luks2/activate.go
```diff
@@ -0,0 +1,97 @@
...
+		// The read end is open and blocked inside read(). Closing our write end will result in the
+		// read end returning 0 bytes (EOF) and exitting cleanly.
+		cmd.Wait()
+		return xerrors.Errorf("cannot pass key to systemd-cryptsetup: %w", err)
+	}
+
+	f.Close()
+	wg.Wait()
+
+	return cmd.Wait()
```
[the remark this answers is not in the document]
chrisccoulson: I can update the documentation to mention this.

internal/luks2/cryptsetup.go
chrisccoulson: I don't mind, although the function name here matches the cryptsetup command name

pedronis: I see, I missed that in my first pass (as I said I didn't quite do a full review yet).  Are there slots which are not keyslots? otherwise for consistency we could instead change SetKeyslotPriority to SetSlotPriority?

I don't have  a strong preference and it's a nitpick but is a bit not ideal to have both Slot and Keyslot interchangebly in the function names.

internal/luks2/metadata.go
```diff
@@ -0,0 +1,874 @@
...
+			// If we are attempting to acquire a lock on a block device, make sure that we
+			// aren't racing with a previous lock holder or a new lock holder.
+			//
+			// Obtain the information about the file currently linked from the lock file path.
+			var st unix.Stat_t
+			if err := unix.Stat(lockPath, &st); err != nil {
+				// The lock file we opened was unlinked by another lock owner between us
+				// opening the file and acquiring the lock. We need to try again.
+				lockFile.Close()
+				continue
```
pedronis, line 161: it's probably slightly hard to setup but this continue and the next are not tested

internal/luks2/metadata.go
```diff
@@ -0,0 +1,874 @@
...
+	if err != nil {
+		return nil, xerrors.Errorf("cannot obtain file info: %w", err)
+	}
+
+	var lockPath string
+	var openFlags int
+
+	switch {
+	case isBlockDevice(fi.Mode()):
+		// Block devices use an advisory lock on a file in /run/cryptsetup. The lock file
```
pedronis, line 101: this is in the main comment but I think it wouldn't hurt to repeat and write this:

// For block devices libcryptsetup uses ...

internal/luks2/metadata.go
```diff
@@ -0,0 +1,874 @@
...
+
+			// We've got an exclusive lock on the lock file we originally opened and locked.
+			// Obtain the information about the file currently at the lock file path.
+			var st unix.Stat_t
+			if err := unix.Stat(lockPath, &st); err != nil {
+				// The lock file we opened has been cleaned up by another process, which acquired
+				// and released the same lock file in between us releasing the lock at the start
+				// of this function, and then acquiring an exclusive lock again. There's nothing
+				// else for us to do.
+				return
```
pedronis, line 218: this return and the next are also not tested

internal/luks2/metadata.go
```diff
@@ -0,0 +1,874 @@
...
+	Salt        [64]byte
+	Uuid        [40]byte
+	Subsystem   [48]byte
+	HdrOffset   uint64
+	Padding     [184]byte
+	Csum        [64]byte
+	Padding4096 [7 * 512]byte
+}
+
+type jsonNumber string
```
pedronis, line 324: is this name used in the luks codebase? it's a bit confusing, if I'm not misreading this,  numbers are serialized as strings?

It's confusing also because there is json.Number that is something else. Unless there some external constraint I would call this luksNumber perhaps or something like that?

internal/luks2/metadata.go
```diff
@@ -0,0 +1,874 @@
...
+
+// Token corresponds to a token object in the JSON metadata of a LUKS2 volume. It
+// describes how to retrieve a passphrase or key for a keyslot.
+type Token struct {
+	Type     string                 // Token type ("luks2-" prefixed types are reserved for cryptsetup)
+	Keyslots []int                  // Keyslots assigned to this token
+	Params   map[string]interface{} // Type-specific parameters for this token
+}
+
+func (t Token) MarshalJSON() ([]byte, error) {
```
pedronis, line 379: this marshal doesn't seem to be tested

internal/luks2/metadata.go
```diff
@@ -0,0 +1,874 @@
...
+	}
+
+	var hdr *binaryHdr
+	var jsonData []byte
+	switch {
+	case primaryHdr != nil && secondaryHdr != nil:
+		// Both headers are valid
+		hdr = primaryHdr
+		jsonData = primaryJSON
+		if secondaryHdr.SeqId > primaryHdr.SeqId {
```
pedronis, line 847: is this prevented by somet invariant?  either way the tests are not hitting this

internal/luks2/metadata.go
```diff
@@ -0,0 +1,874 @@
...
+// Note that this function does not attempt recovery of either header in the event that one of the
+// headers is not valid - we leave this to libcryptsetup, which happens automatically on any
+// cryptsetup or systemd-cryptsetup invocation. This package does not directly perform any
+// modifications to the header.
+//
+// This function requires an advisory shared lock on the LUKS container associated with the
+// specified path. If the mode parameter is LockModeBlocking, this function will block until the
+// lock can be obtained. If the mode parameter is LockModeNonBlocking, a wrapped syscall.Errno
+// error with the value of syscall.EWOULDBLOCK will be returned if the lock can not be obtained.
+func DecodeHdr(path string, lockMode LockMode) (*HdrInfo, error) {
```
pedronis, line 808: is the contraction Hdr a naming preferred by LUKS itself?  because otherwise I would call this DecodeHeader.

Same for other uses of Hdr.

internal/luks2/metadata.go
```diff
@@ -0,0 +1,874 @@
...
+		// well known offsets.
+		for _, off := range []int64{0x4000, 0x8000, 0x10000, 0x20000, 0x40000, 0x80000, 0x100000, 0x200000, 0x400000} {
+			secondaryHdr, secondaryJSON, err = decodeAndCheckHeader(f, off, false)
+			if err == nil {
+				break
+			}
+		}
+	} else {
+		// Try to decode and check the secondary header immediately after the primary header.
+		secondaryHdr, secondaryJSON, _ = decodeAndCheckHeader(f, int64(primaryHdr.HdrSize), false)
```
pedronis, line 837: is  int64(primaryHdr.HdrSize)  and 0x4000 the same?

internal/luks2/metadata.go
```diff
@@ -0,0 +1,874 @@
...
+		// No valid headers :(
+		return nil, xerrors.Errorf("no valid header found, error from decoding primary header: %w", primaryErr)
+	}
+
+	info := &HdrInfo{
+		HdrSize: hdr.HdrSize,
+		Label:   hdr.Label.String()}
+
+	if err := json.NewDecoder(bytes.NewReader(jsonData)).Decode(&info.Metadata); err != nil {
+		return nil, err
```
pedronis, line 870: from the description an error here seems fairly unexpected? it should be prefixed with a bit more contextual info I think than just the bare encoding/json error

internal/luks2/metadata.go
```diff
@@ -0,0 +1,874 @@
...
+	if _, err := io.CopyN(h, tr, int64(hdr.HdrSize)-int64(binary.Size(hdr))); err != nil {
+		return nil, nil, xerrors.Errorf("cannot calculate checksum, error reading JSON metadata: %w", err)
+	}
+
+	if !bytes.Equal(h.Sum(nil), hdr.Csum[0:csumHash.Size()]) {
+		return nil, nil, errors.New("invalid header checksum")
+	}
+
+	// Return the binary header and the memory buffer containing the verified JSON metadata
+	return &hdr, jsonBuffer.Bytes(), nil
```
pedronis, line 779: from the usage it seems this could return the jsonBuffer directly instead of Bytes of it?

pedronis (commented):
I finished a first pass (didn't focus a lot on tests), looks good but some comments/questions

internal/luks2/metadata.go
pedronis: also is the "Decode" part also naming from LUKS?  because "ReadHeader" would be a more natural/clearer name for this

internal/luks2/metadata.go
chrisccoulson: Oh, I thought I'd added a test for this. I can fix that.

internal/luks2/activate.go
```diff
@@ -0,0 +1,99 @@
...
+
+	f, err := os.OpenFile(fifoPath, os.O_WRONLY, 0)
+	if err != nil {
+		// If we fail to open the write end, the read end will be blocked in open()
+		cmd.Process.Kill()
+		return xerrors.Errorf("cannot open FIFO for passing key to systemd-cryptsetup: %w", err)
+	}
+
+	if _, err := f.Write(key); err != nil {
+		f.Close()
```
bboozzoo, line 88: Wondering, shouldn't we rather kill the inferior process here and later call detach in case it did something with the volume?

internal/luks2/activate.go
```diff
@@ -0,0 +1,99 @@
...
+//
+// This will return a *os.ExitError error in the event that systemd-cryptsetup fails.
+func Activate(volumeName, sourceDevicePath string, key []byte) error {
+	fifoPath, cleanupFifo, err := mkFifo()
+	if err != nil {
+		return xerrors.Errorf("cannot create FIFO for passing key to systemd-cryptsetup: %w", err)
+	}
+	defer cleanupFifo()
+
+	cmd := exec.Command(systemdCryptsetupPath, "attach", volumeName, sourceDevicePath, fifoPath, "luks,tries=1")
```
bboozzoo, line 47: Does `-` as input work with systemd-cryptsetup? In which case you could probably avoid using fifo and hook up a bytes.Buffer with the key to cmd.Stdin.

If `-` isn't supported, what may be a bit hackish, I think one could specify /dev/stdin as the input 'file'.

internal/luks2/fifo.go
```diff
@@ -0,0 +1,59 @@
...
+	// /run is not world writable but we create a unique directory here because this
+	// code can be invoked by a public API and we shouldn't fail if more than one
+	// process reaches here at the same time.
+	dir, err := ioutil.TempDir(runDir, filepath.Base(os.Args[0])+".")
+	if err != nil {
+		return "", nil, xerrors.Errorf("cannot create temporary directory: %w", err)
+	}
+
+	cleanup := func() {
+		os.RemoveAll(dir)
```
bboozzoo, line 41: BTW. should we maybe log something to stderr if this fails?

internal/luks2/metadata.go
```diff
@@ -0,0 +1,874 @@
...
+// LockMode defines the locking mode for DecodeHdr.
+type LockMode int
+
+const (
+	LockModeBlocking LockMode = iota
+	LockModeNonBlocking
+)
+
+// acquireSharedLock acquires an advisory shared lock on the LUKS volume associated with the
+// specified path. The ath can either be a block device or file containing a LUKS2 volume with
```
bboozzoo, line 64:
```suggestion
// specified path. The path can either be a block device or file containing a LUKS2 volume with
```

bboozzoo (commented):
Went though half of the PR so far, I'll resume tomorrow morning

internal/luks2/luks2.go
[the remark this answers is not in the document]
chrisccoulson: This is addressed now.

internal/luks2/activate.go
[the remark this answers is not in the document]
chrisccoulson: I've updated the comment to reflect the error type that is returned.

internal/luks2/cryptsetup.go
chrisccoulson: I've renamed SetKeyslotPriority to SetSlotPriority (and the corresponding enum type as well). It's unfortunate that the cryptsetup interface uses both terms interchangeably.

internal/luks2/metadata.go
chrisccoulson: There's no explicit test for it in metadata_test.go, but it should get test coverage as part of the ImportToken tests, although I can add some explicit tests too.

internal/luks2/metadata.go
chrisccoulson: I've added a test for this case now.

internal/luks2/activate.go
chrisccoulson: Ah, passing /dev/stdin as the path seems to work - I verified this manually. I've updated the code to do this instead.

Note that systemd-cryptsetup is mocked in the tests, so I still need to verify that the updated implementation works correctly with the real systemd-cryptsetup.

internal/luks2/metadata.go
chrisccoulson: I've updated the code to do this now.

internal/luks2/metadata.go
chrisccoulson: I've renamed this to luksJsonNumber - ideally the code would just use json.Number but the LUKS2 format uses string-encoded uint64 numbers which aren't supported by json.Number.

internal/luks2/metadata.go
```diff
@@ -0,0 +1,874 @@
...
+// an integral header, or a detached header file associated with a LUKS device.
+//
+// If the mode parameter is LockModeBlocking, this function will block until the lock can be
+// obtained. If the mode parameter is LockModeNonBlocking, a wrapped syscall.Errno error with
+// the value of syscall.EWOULDBLOCK will be returned if the lock can not be obtained.
+//
+// A shared lock is for read-only access. There can be multiple parallel shared lock holders.
+//
+// This function implements the locking logic implemented by libcryptsetup - see
+// lib/utils_device_locking.c from the cryptsetup source code.
```
bboozzoo, line 74: May be useful to mention the revision/version you looked at.

internal/luks2/metadata.go
```diff
@@ -0,0 +1,874 @@
...
+// Note that this function does not attempt recovery of either header in the event that one of the
+// headers is not valid - we leave this to libcryptsetup, which happens automatically on any
+// cryptsetup or systemd-cryptsetup invocation. This package does not directly perform any
+// modifications to the header.
+//
+// This function requires an advisory shared lock on the LUKS container associated with the
+// specified path. If the mode parameter is LockModeBlocking, this function will block until the
+// lock can be obtained. If the mode parameter is LockModeNonBlocking, a wrapped syscall.Errno
+// error with the value of syscall.EWOULDBLOCK will be returned if the lock can not be obtained.
+func DecodeHdr(path string, lockMode LockMode) (*HdrInfo, error) {
```
bboozzoo, line 808: Is it useful for the caller to know whether it's primary or secondary header content?

internal/luks2/metadata.go
```diff
@@ -0,0 +1,874 @@
...
+
+	// Try to decode and check the primary header
+	primaryHdr, primaryJSON, primaryErr := decodeAndCheckHeader(f, 0, true)
+
+	var secondaryHdr *binaryHdr
+	var secondaryJSON *bytes.Buffer
+	if primaryErr != nil {
+		// No valid primary header. Try to decode and check a secondary header from one of the
+		// well known offsets.
+		for _, off := range []int64{0x4000, 0x8000, 0x10000, 0x20000, 0x40000, 0x80000, 0x100000, 0x200000, 0x400000} {
```
bboozzoo, line 829: Reference to an entry in LUKS spec or libcryptsetup source code?

bboozzoo (commented):
Couple more nitpicks.

internal/luks2/metadata.go
chrisccoulson: The LUKS spec does use the abbreviation "hdr" in various places - I've changed DecodeHdr to ReadHeader, HdrInfo to HeaderInfo, but I've kept the "Hdr" abbreviation for the un-exported representation of the binary header.

internal/luks2/metadata.go
chrisccoulson: I've added context to the error now. I've also moved the JSON decoding to before the header selection to make the code a bit more robust (if JSON decoding fails for one header, it provides an opportunity to fallback to the other header if that's ok).

internal/luks2/metadata.go
chrisccoulson: I'm not sure I understand the question here.

internal/luks2/metadata.go
chrisccoulson: Thanks, I've fixed this now.

internal/luks2/activate.go
```diff
@@ -0,0 +1,80 @@
...
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+	go func() {
+		rd := bufio.NewScanner(stdout)
+		for rd.Scan() {
+			fmt.Printf("systemd-cryptsetup: %s\n", rd.Text())
```
pedronis, line 65: as you commented, not completely clear why this should be different and not return this with the error instead?

internal/luks2/metadata.go
```diff
@@ -819,56 +819,73 @@ func ReadHeader(path string, lockMode LockMode) (*HeaderInfo, error) {
...
-			if err == nil {
+			secondaryHdr, secondaryJSONData, secondaryErr = decodeAndCheckHeader(f, off, false)
+			if secondaryErr == nil {
 				break
 			}
 		}
 	} else {
 		// Try to decode and check the secondary header immediately after the primary header.
-		secondaryHdr, secondaryJSON, _ = decodeAndCheckHeader(f, int64(primaryHdr.HdrSize), false)
+		secondaryHdr, secondaryJSONData, secondaryErr = decodeAndCheckHeader(f, int64(primaryHdr.HdrSize), false)
```
pedronis, line 845: my question was because  in if true stanza starts at 0x4000, this one uses int64(primaryHdr.HdrSize) so I'm left wondering if they are the same? or we skip the first possible position for the secondary if the primary can't be read or yet something else?

internal/luks2/metadata.go
pedronis: I reformulated the question again

pedronis (commented):
I did another pass and resolved some points.  I made a couple more comments but also there are some questions from Maciej and me from the first passes still unanswered.

internal/luks2/metadata.go
chrisccoulson: I've updated these comments now.

internal/luks2/activate.go
chrisccoulson: This code has been deleted now.

internal/luks2/fifo.go
chrisccoulson: I've updated this to log a message to stderr if it fails (and in a few other places too).

internal/luks2/metadata.go
chrisccoulson: I've added the git tag to the comment.

internal/luks2/metadata.go
chrisccoulson: I've added a reference here now.

internal/luks2/metadata.go
chrisccoulson: I don't think it would be useful to add this.

internal/luks2/activate.go
chrisccoulson: I've updated this now to not redirect its output to the console.

internal/luks2/cryptsetup.go
```diff
@@ -0,0 +1,202 @@
...
+func ImportToken(devicePath string, token *Token) error {
+	tokenJSON, err := json.Marshal(token)
+	if err != nil {
+		return xerrors.Errorf("cannot serialize token: %w", err)
+	}
+	cmd := exec.Command("cryptsetup", "token", "import", devicePath)
+	cmd.Stdin = bytes.NewReader(tokenJSON)
+	if output, err := cmd.CombinedOutput(); err != nil {
+		return osutil.OutputErr(output, err)
+	}
```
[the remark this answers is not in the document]
chrisccoulson: I've added a helper for this now.

internal/luks2/metadata.go
chrisccoulson: primaryHdr.HdrSize will be set to whatever the size of the header is set to, which can be any one of a finite set of values defined in the LUKS2 spec. The secondary header starts immediately after the end of the primary header. These offsets are the ones tried when the primary header is invalid, starting at 0x4000 (the minimum header size).

internal/luks2/metadata.go
chrisccoulson: There's a test for this now.

internal/luks2/metadata.go
```diff
@@ -0,0 +1,910 @@
...
+	if err != nil {
+		return xerrors.Errorf("invalid offset value: %w", err)
+	}
+	s.Offset = offset
+
+	switch string(d.Size) {
+	case "dynamic":
+		s.DynamicSize = true
+	default:
+		sz, err := d.Size.uint64()
```
pedronis, line 557: this case doesn't seem to be exercised

internal/luks2/metadata.go
```diff
@@ -0,0 +1,910 @@
...
+	var metadata *Metadata
+	switch {
+	case primaryErr == nil && secondaryErr == nil:
+		// Both headers are valid
+		hdr = primaryHdr
+		metadata = &primaryMetadata
+		switch {
+		case secondaryHdr.SeqId < primaryHdr.SeqId:
+			// The secondary header is obsolete. Cryptsetup will recover this automatically.
+			fmt.Fprintf(os.Stderr, "luks2.ReadHeader: secondary header for %s is obsolete\n", path)
```
pedronis, line 882: here and elsewhere you could use something like:

var osStderr = os.Stderr

then in the tests replace it with a buffer,  this avoids spurious output from the tests and also can allow to test what gets printed if desired. At the moment running the tests prints at least this:

```
luks2.ReadHeader: primary header for /tmp/check-5656277456882226488/16/luks2-hdr2-invalid-checksum0.img is invalid: invalid header checksum
luks2.ReadHeader: primary header for /tmp/check-5656277456882226488/20/luks2-hdr-invalid-checksum0.img is invalid: invalid header checksum
luks2.ReadHeader: secondary header for /tmp/check-5656277456882226488/22/luks2-hdr-invalid-checksum1.img is invalid: invalid header checksum
luks2.ReadHeader: primary header for /tmp/check-5656277456882226488/26/luks2-hdr-obsolete0.img is obsolete
```

pedronis (commented):
lgtm but some questions still

internal/luks2/activate_test.go
```diff
@@ -0,0 +1,123 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"math/rand"
+	"path/filepath"
+
+	. "github.com/snapcore/secboot/internal/luks2"
+	snapd_testutil "github.com/snapcore/snapd/testutil"
+
+	. "gopkg.in/check.v1"
```
bboozzoo, line 31:
```suggestion
	. "gopkg.in/check.v1"
	snapd_testutil "github.com/snapcore/snapd/testutil"

	. "github.com/snapcore/secboot/internal/luks2"
```

internal/luks2/cryptsetup.go
```diff
@@ -0,0 +1,216 @@
...
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"strconv"
+	"time"
+
+	"github.com/snapcore/snapd/osutil"
+
+	"golang.org/x/xerrors"
```
bboozzoo, line 34:
```suggestion
	"github.com/snapcore/snapd/osutil"
	"golang.org/x/xerrors"
```

internal/luks2/cryptsetup_test.go
```diff
@@ -0,0 +1,508 @@
...
+	"math/rand"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"time"
+
+	. "github.com/snapcore/secboot/internal/luks2"
+	snapd_testutil "github.com/snapcore/snapd/testutil"
+
+	. "gopkg.in/check.v1"
```
bboozzoo, line 35:
```suggestion
	. "gopkg.in/check.v1"
	snapd_testutil "github.com/snapcore/snapd/testutil"

	. "github.com/snapcore/secboot/internal/luks2"
```

internal/luks2/metadata_test.go
```diff
@@ -0,0 +1,446 @@
...
+	"sync"
+	"time"
+
+	. "github.com/snapcore/secboot/internal/luks2"
+	"github.com/snapcore/secboot/internal/testutil"
+	snapd_testutil "github.com/snapcore/snapd/testutil"
+
+	"golang.org/x/sys/unix"
+
+	. "gopkg.in/check.v1"
```
bboozzoo, line 39:
```suggestion
	snapd_testutil "github.com/snapcore/snapd/testutil"
	"golang.org/x/sys/unix"
	. "gopkg.in/check.v1"

	. "github.com/snapcore/secboot/internal/luks2"	
	"github.com/snapcore/secboot/internal/testutil"
```

bboozzoo (approved):
LGTM. Some tweaks about package import order to make it more like in snapd.

internal/luks2/metadata.go
pedronis: @chrisccoulson this is stil the case it seems? I know you said it was a bit complicated, what's the status on that?

pedronis (approved):
apart the missing coverage as discussed, this looks good

PR #130, opened 2021-03-30
Prepare for platform agnostic APIs

chrisccoulson:
The platform-agnostic APIs will introduce some names that conflict
with existing TPM-specific names. Whilst the TPM code will eventually
move to secboot/tpm, rename some types and move things around a bit
in order to accomodate the new APIs. These changes are a stop-gap
measure until the code moves to its own subpackage.

- Rename keydata.go to keydata_tpm.go
- Rename keyData to tpmKeyData.
- Rename AuthMode to authMode, as the platform-agnostic API
has a different representation for this. Use the existing authMode
type for serialization.
- Introduce a new AuthMode type, which will be a flag type
indicating the enabled auth modes.
- Move the TPM related code in crypt.go to crypt_tpm.go.


mvo5 (approved):
Thank you

PR #133, opened 2021-04-06
Remove GetActivationDataFromKernel and related code

chrisccoulson:
This functionality is unused in snapd at the moment, and the current
implementation works differently to the latest design document - it uses
the key description to pass error status, but error status from the
early boot environment will be preserved differently in a future PR.

As the current functionality is usused, there's no need to support the
current implementation for backwards compatibility reasons, and so it's
going to be easier to just delete this code for now and add bits back
later as required.

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -30,7 +30,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"regexp"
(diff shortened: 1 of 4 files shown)


mvo5 (approved):
Thank you

PR #134, opened 2021-04-06
Add platform agnostic activation

chrisccoulson:


crypt.go
```diff
@@ -237,6 +238,132 @@ func getPassword(sourceDevicePath, description string, reader io.Reader) (string
...
+	return c.volumeName
+}
+
+type activateWithKeyDataError struct {
+	k   *KeyData
+	err error
+}
+
+func (e *activateWithKeyDataError) Error() string {
+	return fmt.Sprintf("%s: %v", e.k.ID(), e.err)
```
pedronis, line 268: as far as I can tell the only way to make a KeyData atm doesn't let specify the id parts, so ID() here could be empty? should some other format than ": error" be used then?

crypt.go
```diff
@@ -237,6 +238,132 @@ func getPassword(sourceDevicePath, description string, reader io.Reader) (string
...
+		return true
+	}
+
+	// TODO: Passphrase support
+
+	// We've failed at this point
+	return false
+}
+
+func makeActivateWithKeyDataState(volumeName, sourceDevicePath string, activateOptions []string, keyringPrefix string, keys []*KeyData) *activateWithKeyDataState {
```
pedronis, line 355: this would be more commonly called new.... as it has also behavior attached

crypt.go
```diff
@@ -271,6 +398,10 @@ func activateWithRecoveryKey(volumeName, sourceDevicePath string, keyReader io.R
 			continue
 		}
 
+		if err := keyring.AddKeyToUserKeyring(key[:], sourceDevicePath, keyringPurposeDiskUnlock, keyringPrefixOrDefault(keyringPrefix)); err != nil {
```
pedronis, line 401: the older code would distinguish at this level whether we had the main key or a recovery key, but now it seems we use "unlock" for either?

crypt.go
```diff
@@ -323,6 +454,100 @@ type ActivateVolumeOptions struct {
 	KeyringPrefix string
 }
 
+type activateVolumeWithKeyDataError struct {
```
pedronis, line 457: should this be exported, similar errors for TPMs are?

crypt.go
```diff
@@ -323,6 +454,100 @@ type ActivateVolumeOptions struct {
...
+		if rErr := activateWithRecoveryKey(volumeName, sourceDevicePath, nil, options.RecoveryKeyTries, activateOptions, options.KeyringPrefix); rErr != nil {
+			// failed with recovery key - return errors
+			var kdErrs []error
+			for _, e := range s.errors() {
+				kdErrs = append(kdErrs, e)
+			}
+			return nil, &activateVolumeWithKeyDataError{kdErrs, rErr}
+		}
+		// succeeded with recovery key
+		return nil, nil
```
pedronis, line 523: the corresponding methods for TPM would list return error info about what failed with the protected keys even in this path, this new code doesn't, it's a bit confusing to have two approaches about this atm

crypt.go
```diff
@@ -237,6 +238,132 @@ func getPassword(sourceDevicePath, description string, reader io.Reader) (string
...
+		if k.err == nil {
+			continue
+		}
+		out = append(out, &activateWithKeyDataError{k: k.KeyData, err: k.err})
+	}
+	return
+}
+
+func (s *activateWithKeyDataState) snapModelChecker() *SnapModelChecker {
+	return &SnapModelChecker{s.volumeName, s.keyData, s.auxKey}
```
pedronis, line 303: what's the plan for TPM going forward about this, to do this and also measure the model, to do only this, that means changing what we measure, also current key locking is based on the PCR used for the model? what will be the relation (if any) of  auxKey vs the reseal auth key?

pedronis (commented):
some comments, some questions about the plans going forward

crypt.go
chrisccoulson: This code executes in early boot where it generally shouldn't be creating new keys, but I've updated NewKeyData to give the key a placeholder name for now.

Originally the ID was going to be based on some storage attributes (in the file case - a device and file path, and in the luks case - a device path + name and revision fields that live inside the JSON metadata but outside of the main key data payload), so the ID is defined once it's persisted to some storage location. But this ended up being too complicated and so I ended up changing it to just a name/revision pair with the name being some arbitrary string defined by the system, and updated the file storage to store a name field.

I wonder whether to just move the name / revision values in to the main KeyData object now instead, so that the name is specified via NewKeyData.

crypt.go
chrisccoulson: I've renamed this.

crypt.go
chrisccoulson: It doesn't matter now. The previous way also tried passing in some error status via the key description, which was a bit of a mess. A future PR will add a mechanism to serialize the unlocking state so that the system can determine if the key used to unlock it was a recovery key, and then use that key to create recreate the primary key(s).

crypt.go
chrisccoulson: I was going to leave this one unexported for now and maybe export it later if we need a way to introspect the errors, but I don't think it will be useful at the moment.

crypt.go
chrisccoulson: See https://github.com/snapcore/secboot/pull/134#discussion_r612570827 - I don't think it's useful to return an error in this case, and a future PR will add an alternative way to access this status.

crypt.go
chrisccoulson: When using the KeyData API, all platforms will use SnapModelChecker, but there's no need to change what we measure to the TPM in the TPM case. Measuring the model there has uses beyond full-disk encryption anyway (eg, providing attestations).

In the TPM case, auxKey will be used to derive the elliptic key used for signing new PCR policies (as opposed to providing the key directly). The current TPMPolicyAuthKey type will go away and be replaced with the new type, but the code can support both old and new keys in order to aid migration - it's easy to determine whether auxKey is actually the elliptic key by deriving the public point from it and comparing it to the stored public key.

crypt.go
pedronis: we have code right now that does that kind of tracking/reporting in snap-bootstrap, it doesn't care about exact errors information but it cares whether the recovery key was used  or not,  the relevant code in snapd is here:

https://github.com/snapcore/snapd/blob/master/secboot/secboot_tpm.go#L594

we need to be able to keep it working even switching to this in some form. Is used only for TPM atm but it would need to work also for this code when we have to wrap the hook code in this.  I don't think we can wait the new code you describe. So this is a regression from our POV.

pedronis (changes requested):
without a sense of when further functionality will come, I think it's a regression that this new code can't communicate with the caller if the recovery key or the main key was used. We don't need error details but we need that bit of information to avoid big changes in the snapd side adopting this.

crypt.go
pedronis: that is reasonable but are there enough bits in aux keys for the double usage to derive both the EC key and the to derive a hmac secret for models?

crypt.go
pedronis: passing the name to NewKeyData seems simpler to me.  But what is the name? it's the name of the full protected key, if I understand correctly?  is not just the name of the underlying key? I mean the same base key protected differently would use different names?

crypt.go
pedronis: sorry, I was just confused, it seems  we can use *SnapModelChecker == nil to detect that the recover key was used, is that correct?

pedronis (commented):
error handling might be fine as is. asked a question

crypt.go
pedronis: I think my "confusion" relates to the fact as well that returning nil, nil like this is kind of atypical and fragile. It's very easy to write code that would panic on the nil pointer by assuming things that aren't true for this function but usually are.

chrisccoulson: Yes, no SnapModelChecker is returned whenever the recovery key is used.

crypt.go
chrisccoulson: The derived keys protect the same resource, so the 32-bits in the "seed" is sufficient to derive both keys.

chrisccoulson: Err, 32 *bytes* :)

crypt.go
pedronis: @chrisccoulson basically I think even just a fixed ErrRecoverKeyUsed or similar would be more in line with common golang behavior patterns.

crypt.go
chrisccoulson: Thinking about this a bit more today, I still think that KeyID is a bit too complicated for its intended purpose (being able to identify the KeyData that unlocked a volume and / or associating KeyData with an error).

I've just created https://github.com/snapcore/secboot/pull/138 which just makes KeyID a genuinely unique value based on the hash of the JSON, with a fresh value automatically created when calling KeyData.ID. It avoids the need for creating arbitrary names that aren't bound in any way to the actual key data, and seems simpler overall.

mvo5: This the changes in #138 this error will become hard to read now, is this true?

crypt.go
```diff
@@ -237,6 +238,132 @@ func getPassword(sourceDevicePath, description string, reader io.Reader) (string
...
+}
+
+func (s *activateWithKeyDataState) errors() (out []*activateWithKeyDataError) {
+	for _, k := range s.keys {
+		if k.err == nil {
+			continue
+		}
+		out = append(out, &activateWithKeyDataError{k: k.KeyData, err: k.err})
+	}
+	return
```
mvo5, line 299: (nitpick) I'm not a fan of "naked" returns

crypt.go
mvo5: I agree with Samuele here, to make our side less confusing I would be in favor of something similar to returning ActivateWithTPMSealedKeyError to keep the hook and TPM  parts somewhat similar (it's already hard to follow). Once we have this future alternative way to access the status we can then switch consistently. WDYT?

mvo5 (commented):
Looks fine, +0.9 but there is this one comment about unlocking with the recovery key that I think woudl be great if addressed.

crypt.go
chrisccoulson: I've added a ErrRecoveryKeyUsed error now.

crypt.go
chrisccoulson: This is fixed now.

crypt.go
chrisccoulson: I'll update this to change the string and not print the full ID, although #138 is currently based on master.

pedronis: This needs to fixed somehow,  I understand now the usage of ID. It doesn't mean we don't still need human-readable information about the provenence/purpose of a key to use in logs/errors?  basically KeyData might still need something like Origin/Source/Provenience that is set by the readers and is not serialized? For the file reader it would/could be the file name?

crypt.go
```diff
@@ -323,6 +454,108 @@ type ActivateVolumeOptions struct {
...
+		if rErr := activateWithRecoveryKey(volumeName, sourceDevicePath, nil, options.RecoveryKeyTries, activateOptions, options.KeyringPrefix); rErr != nil {
+			// failed with recovery key - return errors
+			var kdErrs []error
+			for _, e := range s.errors() {
+				kdErrs = append(kdErrs, e)
+			}
+			return nil, &activateVolumeWithKeyDataError{kdErrs, rErr}
+		}
+		// succeeded with recovery key
+		return nil, ErrRecoveryKeyUsed
```
pedronis, line 530: thanks

pedronis (commented):
thanks, we still need to clarify how to indicate which key an error is about in a way that makes sens for a human afaict

pedronis:
@chrisccoulson @mvo5  grumble, I was trying to dismiss my -1 review but I actually fat-finger merged the PR instead. It seems this repo ins configured to let merge based on tests and ignore if there are negative review votes afaict.

PR #135, opened 2021-04-08
Use the correct recovery key type in crypt_test.go

chrisccoulson:
Also use a 32-bit rather than 64-bit key size during tests, as
this matches the real world.


mvo5 (approved):
Thank you

PR #136, opened 2021-04-09
Remove recoveryKey and primaryKey from cryptSuite

chrisccoulson:
Instead, create them as they are needed by individual tests. In a
future PR, some tests will use different mock key configurations,
and this will just make it a bit simpler.

This is based on https://github.com/snapcore/secboot/pull/135

### DIFF
--- a/crypt_test.go
+++ b/crypt_test.go
@@ -210,8 +210,8 @@ func (ctb *cryptTestBase) addMockKeyslot(c *C, key []byte) {
 	ctb.mockKeyslotsCount++
 }
 
-func (ctb *cryptTestBase) newPrimaryKey(sz int) []byte {
-	key := make([]byte, sz)
+func (ctb *cryptTestBase) newPrimaryKey() []byte {
+	key := make([]byte, 32)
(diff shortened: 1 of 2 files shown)


crypt_test.go
```diff
@@ -220,6 +210,18 @@ func (ctb *cryptTestBase) addMockKeyslot(c *C, key []byte) {
 	ctb.mockKeyslotsCount++
 }
 
+func (ctb *cryptTestBase) newPrimaryKey(sz int) []byte {
```
mvo5, line 213: All primary keys are currently 32 bit so I wonder if maybe this should simply default to this? Or maybe that is the part that will be used in the future? OTOH there were just two PRs that changed the keysizes in the tests to match the real world so I wonder if maybe this should at least be a constant (defaultKeySize) or something?

mvo5 (approved):
Looks fine, just one question

crypt_test.go
chrisccoulson: I ended up just removing the sz argument from this function.

PR #138, opened 2021-04-14
Automatically create a KeyID and make it genuinely unique

chrisccoulson:
Rather than expecting the consumer of the API to supply a key
name which is used to create a KeyID, and rather than storing an
ID inside the KeyData, just make the KeyID automatically generated,
completely unique and make KeyData.ID always return a fresh value.

### DIFF
--- a/keydata.go
+++ b/keydata.go
@@ -153,14 +153,7 @@ type KeyCreationData struct {
 
 // KeyID is the unique ID for a KeyData object. It is used to facilitate the
 // sharing of state between the early boot environment and OS runtime.
-type KeyID struct {
-	Name     string // Unique name for the KeyData
-	Revision uint64 // Revision of the KeyData
-}
-
-func (id KeyID) String() string {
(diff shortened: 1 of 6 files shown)


keydata.go
```diff
@@ -310,8 +299,13 @@ func (d *KeyData) snapModelAuthKey(auxKey AuxiliaryKey) ([]byte, error) {
 }
 
 // ID returns the unique ID for this key data.
-func (d *KeyData) ID() KeyID {
-	return d.id
+func (d *KeyData) ID() (KeyID, error) {
```
pedronis, line 302: This is fine in itself. Fwiw  in snapd we tend to call this kind of methods that return a unique/key-mapping identifier Unique.

But my comment https://github.com/snapcore/secboot/pull/134/files#r613927267 still needs to be addressed somehow.

pedronis (commented):
this is reasonable but after this we are lacking in human/developer-friendly ways to identify keys if given only KeyData.

pedronis (approved):
thank you

PR #141, opened 2021-04-28
Create an efi subpackage

chrisccoulson:
In preparation for moving the TPM code out of the core package into a
subpackage, move all of the EFI-specific code into its own subpackage.
The main advantage of this is that snap-bootstrap will not have to pull
in EFI-related dependencies even on UEFI+TPM platforms.


pedronis (approved):
thank you

(snapd secboot will need updates @mvo)

PR #142, opened 2021-04-28
Remove ActivateOptions field from ActivateVolumeOptions

chrisccoulson:
This is currently unused by snapd.

The field is problematic because it exposes all of systemd-cryptsetup's
options, many of which are incompatible with how we use it in secboot.

If there is a genuine need in the future to expose some extra options,
this can be done with a more explicit API instead.

This also makes it slightly easier to move the TPM code out of the
core secboot package because it's not necessary to duplicate
makeActivateOptions or move it to an internal package.

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -144,14 +144,14 @@ func mkFifo() (string, func(), error) {
 	return fifo, cleanup, nil
 }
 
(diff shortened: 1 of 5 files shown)


pedronis (approved):
thanks

PR #143, opened 2021-04-29
Create tpm subpackage

chrisccoulson:
This moves the TPM code out of the core package and into a subpackage
of its own. This allows non-TPM platforms to avoid pulling in all of the
TPM related code.

Note that some code is duplicated between crypt.go + tpm2/crypt.go and
crypt_test.go + tpm2/crypt_test.go. I haven't made the effort to refactor
this so that the code can be shared because I'm going to try to make
it possible to unlock LUKS devices with the current TPM sealed key files
using the new KeyData based API (crypt.go). This will work by constructing a
dummy KeyData with an empty encrypted payload, the platform data set
to the serialized TPM sealed key data, and a platform name of "tpm2-legacy"
or something like that. The "tpm2-legacy" platform implemented in tpm2/ will
take care of the unsealing.

Future PRs will work towards creating a proper tpm2 KeyData platform
implementation whilst maintaining the ability to use existing TPM key files.
A big pre-requisite for this is to remove the SealedKeyObject hard dependency
on file paths so that it only cares about bytes.

The diff is quite large, but the majority of this is updates to imports, renaming of
things and reorganisation of test code.


tpm2/crypt.go
```diff
@@ -17,17 +17,90 @@
...
+	cmd := exec.Command(
+		"systemd-ask-password",
+		"--icon", "drive-harddisk",
+		"--id", filepath.Base(os.Args[0])+":"+sourceDevicePath,
+		msg)
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	cmd.Stdin = os.Stdin
+	if err := cmd.Run(); err != nil {
+		return "", wrapExecError(cmd, err)
```
pedronis, line 81: just noticed in passing that this code path is not tested, also not for the original copy of the code, but I also wonder if we still need execError as isExecError is not used anymore?

pedronis (commented):
this looks reasonable, the double tpm2 package is slightly annoying though, I wonder if at this level secboot/tpm  would be clear enough?  the old names weren't TPM2-prefixed but just TPM either in most cases.

A passing comment about some error handling that caught my eyes and now seems a bit more complicated than maybe needed.

chrisccoulson:
> this looks reasonable, the double tpm2 package is slightly annoying though, I wonder if at this level secboot/tpm would be clear enough? the old names weren't TPM2-prefixed but just TPM either in most cases.
> 
I was wondering whether to actually name this package secboot_tpm2 (and similarly for ./efi/ to secboot_efi), as they mostly have to be imported as such in order to avoid conflicting with canonical/go-tpm2 and canonical/go-efilib.

tpm2/crypt.go
chrisccoulson: It looks like execError is not needed now.

pedronis:
> I was wondering whether to actually name this package secboot_tpm2 (and similarly for ./efi/ to secboot_efi), as they mostly have to be imported as such in order to avoid conflicting with canonical/go-tpm2 and canonical/go-efilib.

that is not the recommended approach of go package naming conventions. So I think it's either tpm2 or tpm

chrisccoulson:
> > I was wondering whether to actually name this package secboot_tpm2 (and similarly for ./efi/ to secboot_efi), as they mostly have to be imported as such in order to avoid conflicting with canonical/go-tpm2 and canonical/go-efilib.
> 
> that is not the recommended approach of go package naming conventions. So I think it's either tpm2 or tpm

I don't mind too much whether it's tpm2 or tpm - I chose tpm2 because it's explicit about not supporting TPM1.2 devices. One issue with calling it tpm is that this name is used in quite a few places as the tpm2.Connection.

pedronis (approved):
looked a bit at usage numbers of the various identifiers in play. "tpm2" is a bit annoying but better than switching to just "tpm"

PR #145, opened 2021-04-29
Support provisioning the TPM with a custom SRK

chrisccoulson:
TPMConnection.EnsureProvisioned will create and persist a storage
primary key as described in the "TCG TPM v2.0 Provisioning Guidance"
specification (https://trustedcomputinggroup.org/wp-content/uploads/TCG-TPM-v2.0-Provisioning-Guidance-Published-v1r1.pdf).

This specification says that the unique field of the storage primary
key template should be 256 bytes of zeroes, which is the same as the
endorsement key template.

However, many existing tools (including tpm2-tools) define the template
for the storage primary key with an empty unique field, and this creates
a different primary key. This is a deviation from the TCG specifications.

Whilst this isn't a problem for Ubuntu Core, it may be a problem on
systems where the initial encryption and key protection happens before
first boot, and the disk unlock key is protected with a primary key that
is created with third party tools.

In order to support this scenario, introduce a new
TPMConnection.EnsureProvisionedWithCustomSRK API. It is not expected that
this API will be used on Ubuntu Core, but may be used elsewhere.

Note that SealKeyToTPM unconditionally recreates the storage primary
key if it is executed with a TPMConnection that didn't perform the
provisioning step. The reason for this is that creating the primary
key is really the only way to ensure that we are sealing to a key that
we know we can recreate if it is deleted in the future. In order to
avoid extending the SealKeyToTPM API, the new EnsureProvisionedWithCustomSRK
function will store the custom template used to create the primary key
inside the TPM. SealKeyToTPM will use this template stored inside the TPM
(if it exists) to recreate the storage primary key.

### DIFF
--- a/provisioning.go
+++ b/provisioning.go
@@ -174,59 +174,8 @@ func removeStoredSrkTemplate(tpm *tpm2.TPMContext, session tpm2.SessionContext) 
 	return nil
 }
 
-// EnsureProvisionedWithCustomSRK prepares the TPM for full disk encryption. The mode parameter specifies the behaviour of this
-// function.
-//
-// EnsureProvisioned is generally preferred over this function.
-//
-// If mode is ProvisionModeClear, this function will attempt to clear the TPM before provisioning it. If owner clear has been
-// disabled (which will be the case if the TPM has previously been provisioned with this function), then ErrTPMClearRequiresPPI
-// will be returned. In this case, the TPM must be cleared via the physical presence interface by calling RequestTPMClearUsingPPI
-// and performing a system restart. Note that clearing the TPM makes all previously sealed keys permanently unrecoverable. This
-// mode should normally be used when resetting a device to factory settings (ie, performing a new installation).
-//
-// If mode is ProvisionModeClear or ProvisionModeFull, then the authorization value for the lockout hierarchy will be set to
-// newLockoutAuth, owner clear will be disabled, and the parameters of the TPM's dictionary attack logic will be configured to
-// appropriate values.
-//
-// If mode is ProvisionModeClear or ProvisionModeFull, this function performs operations that require the use of the lockout
-// hierarchy (detailed above), and knowledge of the lockout hierarchy's authorization value. This must be provided by calling
-// TPMConnection.LockoutHandleContext().SetAuthValue() prior to this call. If the wrong lockout hierarchy authorization value is
-// provided, then a AuthFailError error will be returned. If this happens, the TPM will have entered dictionary attack lockout mode
-// for the lockout hierarchy. Further calls will result in a ErrTPMLockout error being returned. The only way to recover from this is
-// to either wait for the pre-programmed recovery time to expire, or to clear the TPM via the physical presence interface by calling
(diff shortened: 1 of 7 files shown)


provisioning.go
```diff
@@ -41,6 +42,8 @@ const (
 	maxTries        uint32 = 32
 	recoveryTime    uint32 = 7200
 	lockoutRecovery uint32 = 86400
+
+	srkTemplateHandle tpm2.Handle = 0x01810001
```
pedronis, line 46: does this need a comment about the choice of handle?

provisioning.go
```diff
@@ -90,7 +93,87 @@ func provisionPrimaryKey(tpm *tpm2.TPMContext, hierarchy tpm2.ResourceContext, t
...
+	nv, err := tpm.NVDefineSpace(tpm.OwnerHandleContext(), nil, &nvPub, session)
+	if err != nil {
+		return xerrors.Errorf("cannot define NV index: %w", err)
+	}
+
+	if err := tpm.NVWrite(nv, nv, tmplB, 0, session); err != nil {
+		return xerrors.Errorf("cannot write NV index: %w", err)
+	}
+
+	if err := tpm.NVWriteLock(nv, nv, session); err != nil {
```
pedronis, line 148: is this locking tested?

provisioning_test.go
```diff
@@ -576,3 +510,178 @@ func TestProvisionWithInvalidEkCert(t *testing.T) {
...
+	if err == nil {
+		t.Fatalf("EnsureProvisionedWithCustomSRK should have failed")
+	}
+
+	if err.Error() != "supplied SRK template is not valid for a parent key" {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+}
+
+func TestRemoveStoredCustomSRKTemplate(t *testing.T) {
```
pedronis, line 629: shouldn't this also be called something like TestProvisionDefaultRemoves... ?

pedronis (commented):
this looks alright to me, some comments

provisioning.go
chrisccoulson: I've added a comment to this now.

provisioning.go
chrisccoulson: I've added a test for this now.

provisioning_test.go
chrisccoulson: I've renamed this, but thinking about it a bit more - I'm not sure that the behaviour implemented here is actually correct. I think that EnsureProvisioned should use the custom template if EnsureProvisionedWithCustomSRK was used previously and the TPM isn't being cleared. This would make it possible to inject a custom template into an image, execute EnsureProvisionedWithCustomSRK on first boot and then be able to just use the existing simpler API after that if it's required, without having do do things like "if hasCustomTemplate { EnsureProvisionedWithCustomSRK() } else { EnsureProvisioned() }"

pedronis: yes, your new suggested behavior seems reasonable as well

chrisccoulson: Ok, I've updated this to change the behaviour now

pedronis (approved):
thanks

PR #144, opened 2021-04-29
Update the vendored go-tpm2

chrisccoulson:
I need some new APIs (Public.IsStorage and CreateDuplicationObjectFromSensitive), and I'm also going to start making use of go-tpm2/testutil

PR #146, opened 2021-04-30
Expose the TPM error when the public part of the elliptic key is invalid *debug only PR*

chrisccoulson:
There is a test failure in travis which isn't reproducible locally

### DIFF
--- a/internal/testutil/suites.go
+++ b/internal/testutil/suites.go
@@ -76,7 +76,7 @@ func (b *TPMTestBase) setUpTestBase(c *C, tpm *secboot.TPMConnection) {
 				hc, err = b.TPM.CreateResourceContextFromTPM(h)
 				c.Check(err, IsNil)
 			case tpm2.HandleTypeHMACSession:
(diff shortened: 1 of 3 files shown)


chrisccoulson:
Well, this is fantastic. The tests pass and the only change is between this and https://github.com/snapcore/secboot/pull/144 is https://github.com/snapcore/secboot/pull/146/commits/e3b76f13c3063a31ca744d5002819c75abe6c26c

chrisccoulson:
Huh, this is an obviously intermittent test failure that by some act of magic has never failed before. I think the issue is that go's big.Int truncates numbers to the minimum number of bytes required to hold the number, but the TPM requires key sizes to match the size specified by their configuration - so in the case of an elliptic key using the NIST P-256 curve, the public points must be 32 bytes. This means that keys generated by go's crypto/ecdsa package must be zero-extended to 32 bytes before loading them in to the TPM, something which this test isn't doing and I suspect is the reason for the test failures in https://github.com/snapcore/secboot/pull/144 and https://github.com/snapcore/secboot/pull/145

PR #147, opened 2021-05-04
Import github.com/canonical/go-efilib

chrisccoulson:
This makes use of github.com/canonical/go-efilib, which tcglog-parser also pulls in, and replaces some of the functionality that we had in this repository.


efi/secureboot_policy_test.go
```diff
@@ -340,43 +130,49 @@ func TestComputeDbUpdate(t *testing.T) {
...
+			newSignatures: []int{77},
 		},
 		{
 			desc:          "AppendMsDbxUpdate/2",
 			orig:          "testdata/efivars2/dbx-d719b2cb-3d3a-4596-a3bc-dad00e67656f",
 			update:        "testdata/updates1/dbx/MS-2016-08-08.bin",
 			quirkMode:     SigDbUpdateQuirkModeDedupIgnoresOwner,
 			sha1hash:      testutil.DecodeHexStringT(t, "96f7dc104ee34a0ce8425aac20f29e2b2aba9d7e"),
-			newSignatures: 77,
+			newESLs:       1,
```
pedronis, line 142: does it make sense to have tests where this is > 1 ?

pedronis (approved):
+1,  one question

Skimming: also the for loop using isNewSig could maybe also use a "break Label",  it seems it needs the bool anyway, but it could be slightly more readable

efi/secureboot_policy_test.go
chrisccoulson: I've added the 2020 dbx update now, which has more than 1 ESL.

chrisccoulson:
> +1, one question
> 
> Skimming: also the for loop using isNewSig could maybe also use a "break Label", it seems it needs the bool anyway, but it could be slightly more readable

I've added a label to the outer loop now.

internal/efi/efi.go
```diff
@@ -19,7 +19,13 @@
...
 
+import (
+	"github.com/canonical/go-efilib"
+)
+
 var (
 	EventLogPath = "/sys/kernel/security/tpm0/binary_bios_measurements" // Path of the TCG event log for the default TPM, in binary form
 	EFIVarsPath  = "/sys/firmware/efi/efivars"                          // Default mount point for efivarfs
+
+	ReadVar = efi.ReadVar
```
bboozzoo, line 30: For a minute there I though that we could update snapd to use go-efilib too, but we only read the efi variables so it's probably not really worth it.

bboozzoo (approved):
LGTM

efi/secureboot_policy_test.go
```diff
@@ -160,6 +160,15 @@ func TestComputeDbUpdate(t *testing.T) {
 			newESLs:       1,
 			newSignatures: []int{1},
 		},
+		{
+			desc:          "AppendDbxUpdateWithMultipleESLs",
+			orig:          "testdata/efivars2/dbx-d719b2cb-3d3a-4596-a3bc-dad00e67656f",
+			update:        "testdata/updates5/dbx/dbxupdate_x64_1.bin",
+			quirkMode:     SigDbUpdateQuirkModeNone,
+			sha1hash:      testutil.DecodeHexStringT(t, "306985f67cdc580e25c1572568f9e65e420984d2"),
+			newESLs:       4,
```
pedronis, line 169: thanks

pedronis (approved):
still +1

PR #148, opened 2021-05-05
efi: Allow the EFI environment to be customized by the caller

chrisccoulson:
This makes it possible to precompute PCR profiles outside of the
host for which the profile is computed for.


efi/efi.go
```diff
@@ -32,6 +35,10 @@ const (
 	certTableIndex = 4 // Index of the Certificate Table entry in the Data Directory of a PE image optional header
 )
 
+var (
+	eventLogPath = "/sys/kernel/security/tpm0/binary_bios_measurements" // Path of the TCG event log for the default TPM, in binary form
+)
+
```
pedronis, line 41: nitpick: I would move HostEnviroment definition here

pedronis (approved):
this looks reasonable to me

PR #149, opened 2021-05-06
efi: Compute PK measurement digest based on the current variable value

chrisccoulson:
When precomputing values for PCR7, the digest for the PK mesurement is
replayed directly from the event log even though this is a runtime
updateable variable (as long as you have an update signed with the
current PK!), so this isn't really the correct behaviour. This updates
the code to behave as follows:
- For secure boot configuration events that correspond to non-volatile
runtime variables (PK, KEK, db and dbx, and dbt and dbr if present), the
digest is computed based on the current variable value, as this value is
authoritative.
- For secure boot configuraion events that correspond to volatile
runtime variables (SecureBoot, and DeployedMode if present and disabled),
the digest is replayed directly from the event log. This is because
volatile runtime variables are mirrored by boot services code from
boot service only non-volatile variables, and the backing non-volatile
variables are not available for read or write at runtime. In this case,
the value recorded in the log is authoritative.

To test that the code picks the platform key from the correct source,
I've added a PK variable to each of the efi/testdata/efivars* directories
containing the platform key that the tests originally used from the test
event logs, and I've added an event log from my own Dell machine which
contains a different platform key, and used this log for the tests to
make sure that the computed digests didn't change.


efi/secureboot_policy.go
```diff
@@ -509,21 +501,29 @@ func (b *secureBootPolicyGenBranch) processPreOSEvents(events []*tcglog.Event, s
...
 		case isDbMeasurementEvent(e):
+			// This is the db variable - requires special handling because it updates context
+			// for this branch.
 			if err := b.processDbMeasurementEvent(sigDbUpdates, sigDbUpdateQuirkMode); err != nil {
 				return xerrors.Errorf("cannot process db measurement event: %w", err)
 			}
-		case isDbxMeasurementEvent(e):
-			if err := b.processDbxMeasurementEvent(sigDbUpdates, sigDbUpdateQuirkMode); err != nil {
-				return xerrors.Errorf("cannot process dbx measurement event: %w", err)
+		case isSignatureDatabaseMeasurementEvent(e):
```
pedronis, line 510: if I understand the code change correctly the `data.VariableName == efi.GlobalVariable && data.UnicodeName == pkName` case of this before would have been handled below under `case e.PCRIndex == secureBootPCR` (now ``isSecureBootEvent(e)) ?

efi/secureboot_policy.go
```diff
@@ -257,41 +258,50 @@ func buildSignatureDbUpdateList(keystores []string) ([]*secureBootDbUpdate, erro
...
+		return false
+	}
+
+	data := event.Data.(*tcglog.EFIVariableData)
+	switch {
+	case data.VariableName == efi.GlobalVariable && data.UnicodeName == pkName:
+		return true
+	case data.VariableName == efi.GlobalVariable && data.UnicodeName == kekName:
+		return true
+	case data.VariableName == efi.ImageSecurityDatabaseGuid:
```
pedronis, line 295: this covers db and dbx, but db is treated separately? are there more cases that will hit here now?

pedronis (commented):
I think the code does what the description says, but a couple of questions to make sure I'm reading it right (the change is rather subtle).

efi/secureboot_policy.go
chrisccoulson: That's correct.

efi/secureboot_policy.go
chrisccoulson: Not in this PR, but there will be in a future PR (see https://github.com/chrisccoulson/secboot/commit/df2bc9d598e821f828e9d536bb1f1168c14976ce)

pedronis (approved):
this does what the description explains, and the reasoning in the description looks sound

PR #150, opened 2021-05-07
compattests: Add tests for current v1 TPM metadata format

chrisccoulson:


internal/compattest/v0_test.go
```diff
@@ -122,7 +122,7 @@ func (s *compatTestV0Suite) TestUnsealAfterLock(c *C) {
 	// Test unsealing a v0 file from a newer initramfs using the fence-style locking - this just makes
 	// the PCR values invalid so there's no reason this shouldn't work or require a compatibility test,
 	// but keep this here just to make sure.
-	c.Assert(secboot.BlockPCRProtectionPolicies(s.TPM, []int{12}), IsNil)
 	s.replayPCRSequenceFromFile(c, s.absPath("pcrSequence.1"))
+	c.Assert(secboot.BlockPCRProtectionPolicies(s.TPM, []int{12}), IsNil)
```
pedronis, line 126: why this change?

pedronis (commented):
looks good but a question

internal/compattest/v0_test.go
chrisccoulson: Although the end result is the same (albeit with different PCR values afterwards), the original order isn't what happens on a real system and this just corrects that.

pedronis (approved):
thanks

PR #151, opened 2021-05-10
Add SealKeyToExternalTPMStorageKey API

chrisccoulson:
This allows a key to be protected by an arbitrary TPM key without
access to the actual TPM. The key will be protected by a duplication
object that has to be imported into the target TPM before it is used.

Note that this doesn't yet include an API to persist the imported object,
although unsealing works ok as it imports the object automatically.
The imported object will be persisted automatically on the first modification
to the key data.

### DIFF
--- a/keydata_tpm.go
+++ b/keydata_tpm.go
@@ -252,7 +252,7 @@ type keyDataRaw_v2 struct {
 	KeyPrivate        tpm2.Private
 	KeyPublic         *tpm2.Public
 	AuthModeHint      authMode
-	ImportSymSeed	  tpm2.EncryptedSecret
+	ImportSymSeed     tpm2.EncryptedSecret
 	StaticPolicyData  *staticPolicyDataRaw_v1
 	DynamicPolicyData *dynamicPolicyDataRaw_v0
 }
@@ -265,7 +265,7 @@ type tpmKeyData struct {
 	keyPrivate        tpm2.Private
 	keyPublic         *tpm2.Public
 	authModeHint      authMode
-	importSymSeed	  tpm2.EncryptedSecret
+	importSymSeed     tpm2.EncryptedSecret
 	staticPolicyData  *staticPolicyData
 	dynamicPolicyData *dynamicPolicyData
 }
@@ -297,7 +297,7 @@ func (d tpmKeyData) Marshal(w io.Writer) error {
 			KeyPrivate:        d.keyPrivate,
 			KeyPublic:         d.keyPublic,
 			AuthModeHint:      d.authModeHint,
-			ImportSymSeed:	   d.importSymSeed,
+			ImportSymSeed:     d.importSymSeed,
 			StaticPolicyData:  makeStaticPolicyDataRaw_v1(d.staticPolicyData),
 			DynamicPolicyData: makeDynamicPolicyDataRaw_v0(d.dynamicPolicyData)}
 		if _, err := mu.MarshalToWriter(&tmpW, raw); err != nil {
@@ -377,7 +377,7 @@ func (d *tpmKeyData) Unmarshal(r mu.Reader) error {
 			keyPrivate:        raw.KeyPrivate,
 			keyPublic:         raw.KeyPublic,
 			authModeHint:      raw.AuthModeHint,
-			importSymSeed:	   raw.ImportSymSeed,
+			importSymSeed:     raw.ImportSymSeed,
 			staticPolicyData:  raw.StaticPolicyData.data(),
 			dynamicPolicyData: raw.DynamicPolicyData.data()}
 	default:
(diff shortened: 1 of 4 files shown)


seal.go
```diff
@@ -135,6 +155,148 @@ type KeyCreationParams struct {
...
+// specified by keyPath.
+//
+// The tpmKey argument must correspond to the storage primary key on the target TPM, persisted at the standard handle.
+//
+// This function expects there to be no file at the specified path. If keyPath references a file that already exists, a wrapped
+// *os.PathError error will be returned with an underlying error of syscall.EEXIST. A wrapped *os.PathError error will be returned if
+// the file cannot be created and opened for writing.
+//
+// This function cannot create a sealed key that uses a PCR policy counter. The PCRPolicyCounterHandle field of the params argument
+// must be tpm2.HandleNull.
```
pedronis, line 171: what's the invalidation plan given this?

seal.go
```diff
@@ -44,29 +45,48 @@ func makeSealedKeyTemplate() *tpm2.Public {
 		Params:  &tpm2.PublicParamsU{KeyedHashDetail: &tpm2.KeyedHashParams{Scheme: tpm2.KeyedHashScheme{Scheme: tpm2.KeyedHashSchemeNull}}}}
 }
 
+func makeImportableSealedKeyTemplate() *tpm2.Public {
+	tmpl := makeSealedKeyTemplate()
+	tmpl.Attrs &^= tpm2.AttrFixedTPM | tpm2.AttrFixedParent
+	return tmpl
+}
+
 func computeSealedKeyDynamicAuthPolicy(tpm *tpm2.TPMContext, version uint32, alg, signAlg tpm2.HashAlgorithmId, authKey crypto.PrivateKey,
```
pedronis, line 54: this probably needs a comment about the different behavior depending on tpm != nil or == nil

seal.go
```diff
@@ -135,6 +155,148 @@ type KeyCreationParams struct {
...
+	}
+
+	if params.PCRPolicyCounterHandle != tpm2.HandleNull {
+		return nil, errors.New("PCRPolicyCounter must be tpm2.HandleNull when creating an importable sealed key")
+	}
+
+	succeeded := false
+
+	// Compute metadata.
+
```
pedronis, line 201: there's a bit of code here that is mostly shared up to parameters with the non importable case, maybe a helper could be shared or not

seal.go
```diff
@@ -135,6 +155,148 @@ type KeyCreationParams struct {
...
+	defer f.Close()
+
+	// Create the sensitive data
+	sealedData, err := mu.MarshalToBytes(sealedData{Key: key, AuthPrivateKey: authKey})
+	if err != nil {
+		panic(fmt.Sprintf("cannot marshal sensitive data: %v", err))
+	}
+	sensitive := tpm2.Sensitive{
+		Type:      pub.Type,
+		AuthValue: make(tpm2.Auth, pub.NameAlg.Size()),
```
pedronis, line 263: should this have a comment that no authorization is set?

keydata_tpm.go
```diff
@@ -340,15 +357,66 @@ func (d *tpmKeyData) Unmarshal(r mu.Reader) error {
...
+			importSymSeed:     raw.ImportSymSeed,
+			staticPolicyData:  raw.StaticPolicyData.data(),
+			dynamicPolicyData: raw.DynamicPolicyData.data()}
 	default:
 		return fmt.Errorf("unexpected version number (%d)", version)
 	}
 	return nil
 }
 
+func (d *tpmKeyData) ensureImported(tpm *tpm2.TPMContext, session tpm2.SessionContext) error {
```
pedronis, line 389: this helper merits a comment I think

pedronis (commented):
I did  a first pass, some comments, some questions

seal.go
chrisccoulson: There isn't really going to be a way to support this with these types of objects because we need to be able to create the counter in order to compute the PCR policy.

seal.go
chrisccoulson: I can actually omit setting this field here - I forgot that it gets padded automatically in go-tpm2 instead: https://github.com/canonical/go-tpm2/blob/master/utils.go#L461

keydata_tpm.go
chrisccoulson: I've added a comment to this now.

seal.go
chrisccoulson: I've added a comment to this now.

seal.go
chrisccoulson: I think I could probably create a helper - I didn't do that initially just because it was quicker this way to get something working.

pedronis: my reason to wonder about that is not the code reduction, is mostly to have one place to change if that are needs to be touched (either bug fixes, or refactoring or other changes)

pedronis (approved):
thanks, my comment about a helper is not a blocker but I explained my reasoning for wondering about it

PR #152, opened 2021-05-27
Use internal luks2 package

chrisccoulson:
This ports crypt.go to use internal/luks2, with no other changes yet.

I've kept the original tests that use a mocked cryptsetup, but added
some additional tests that use a real cryptsetup. As these take a while
to run, I've added a --no-expensive-cryptsetup-tests option to run-tests
which skips the new tests (and the cryptsetup tests in internal/luks2).
This just passes an option in via the environment to avoid having to
parse a new argument in every test binary, and to avoid splitting the
single call to go test.

The existing API needs to be able to use the luksAddKey command
with the --key-slot option which internal/luks2 didn't support, so I've
added that too.

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -651,7 +651,10 @@ func addKeyToLUKS2Container(devicePath string, existingKey, key []byte, extraOpt
 //
 // The recovery key is provided via the recoveryKey argument and must be a cryptographically secure 16-byte number.
 func AddRecoveryKeyToLUKS2Container(devicePath string, key []byte, recoveryKey RecoveryKey) error {
-	return luks2.AddKey(devicePath, key, recoveryKey[:], 5*time.Second)
+	options := luks2.AddKeyOptions{
+		KDFTime: 5 * time.Second,
+		Slot:    luks2.AnySlot}
+	return luks2.AddKey(devicePath, key, recoveryKey[:], &options)
 }
 
 // ChangeLUKS2KeyUsingRecoveryKey changes the key normally used for unlocking the LUKS2 container at devicePath. This function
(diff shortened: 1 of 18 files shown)


crypt.go
```diff
@@ -26,23 +26,23 @@ import (
...
-	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/snapcore/secboot/internal/keyring"
+	"github.com/snapcore/secboot/internal/luks2"
```
pedronis, line 37: same nitpick

crypt.go
```diff
@@ -645,95 +541,24 @@ func InitializeLUKS2Container(devicePath, label string, key []byte, options *Ini
...
-		devicePath)
-	cmd := exec.Command("cryptsetup", args...)
-	cmd.Stdin = bytes.NewReader(key)
-	if output, err := cmd.CombinedOutput(); err != nil {
-		return osutil.OutputErr(output, err)
+		opts.MetadataKiBSize = options.MetadataKiBSize
+		opts.KeyslotsAreaKiBSize = options.KeyslotsAreaKiBSize
 	}
 
-	return setLUKS2KeyslotPreferred(devicePath, 0)
```
pedronis, line 686: afaict  setLUKS2KeyslotPreferred is now unused

pedronis (commented):
thanks, this looks good, some small remarks

internal/luks2/luks2test/luks2test.go
```diff
@@ -0,0 +1,63 @@
...
+func CreateEmptyDiskImage(c *C, sz int) string {
+	f, err := os.OpenFile(filepath.Join(c.MkDir(), "disk.img"), os.O_RDWR|os.O_CREATE, 0600)
+	c.Assert(err, IsNil)
+	defer f.Close()
+
+	c.Assert(f.Truncate(int64(sz)*1024*1024), IsNil)
+	return f.Name()
+}
+
+func CheckLUKS2Passphrase(c *C, path string, key []byte) {
```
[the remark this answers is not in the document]
chrisccoulson: I've updated this now.

internal/luks2/cryptsetup_test.go
```diff
@@ -22,14 +22,14 @@ package luks2_test
...
-	"fmt"
 	"math/rand"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"time"
 
 	. "github.com/snapcore/secboot/internal/luks2"
+	"github.com/snapcore/secboot/internal/luks2/luks2test"
+	"github.com/snapcore/secboot/internal/paths/pathstest"
```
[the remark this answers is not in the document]
chrisccoulson: I've changed the order of these imports now, although it's probably worth submitting another PR to correct others too.

crypt.go
chrisccoulson: It is - I've deleted it now.

pedronis (approved):
+1

PR #154, opened 2021-06-29
Add RevokeOldPCRProtectionPolicies API

chrisccoulson:
The UpdatePCRProtectionPolicy API computes a new PCR policy, writes
the updated key data to disk and then revokes old PCR policies.
In order to remove SealedKeyObject's hard dependency on files, the
second step of writing the updated key data to disk has to be dropped,
as this will be handled outside of SealedKeyObject.

This means that the PCR policy revocation must move to its own API,
as revocation has to happen after the updated key data is persisted
to disk.


tpm2/seal.go
```diff
@@ -663,17 +671,49 @@ func (k *SealedKeyObject) UpdatePCRProtectionPolicy(tpm *Connection, authKey Pol
 	return updateKeyPCRProtectionPolicyCommon(tpm.TPMContext, []*SealedKeyObject{k}, ecdsaAuthKey, pcrProfile, tpm.HmacSession())
 }
 
+// RevokeOldPCRProtectionPolicies revokes old PCR protection policies associated with this sealed key. If the key
+// data was not created with a PCR policy counter, then this function does nothing. The caller must also specify
+// the private part of the authorization key that was either returned by SealKeyToTPM or SealedKeyObject.UnsealFromTPM.
+//
+// If validation of the key data fails, a InvalidKeyFileError error will be returned.
+func (k *SealedKeyObject) RevokeOldPCRProtectionPolicies(tpm *Connection, authKey PolicyAuthKey) error {
```
pedronis, line 679: there's no comment nor check that is only applied on the result of UpdatePCRProtectionPolicy, afaict this will not revoke things if that's not the case because the counter will not have been incremented in the policy data?

Maybe it just needs a better doc comment, but my point I suppose is that this does what the name implies only if k.data.dynamicPolicyData.policyCount had changed

tpm2/policy_test.go
```diff
@@ -62,25 +58,25 @@ func TestIncrementPcrPolicyCounter(t *testing.T) {
...
-	initialCount, err := ReadPcrPolicyCounter(tpm.TPMContext, CurrentMetadataVersion, policyCounterPub, nil, tpm.HmacSession())
+	h, err := NewPcrPolicyCounterHandleV1(policyCounterPub, keyPublic)
 	if err != nil {
-		t.Errorf("ReadPcrPolicyCounter failed: %v", err)
+		t.Fatalf("NewPcrPolicyCounterHandleV1 failed: %v", err)
 	}
 
-	if err := IncrementPcrPolicyCounter(tpm.TPMContext, CurrentMetadataVersion, policyCounterPub, nil, key, keyPublic, tpm.HmacSession()); err != nil {
-		t.Fatalf("IncrementPcrPolicyCounter failed: %v", err)
+	if err := h.Set(tpm.TPMContext, initialCount+1, key, tpm.HmacSession()); err != nil {
```
pedronis, line 66: is there a check that Set supports a delta larger than 1, we don't use that, but the code is supposed to support and with some complexity so it would be good to test for that?

tpm2/policy.go
```diff
@@ -151,189 +152,263 @@ func makeStaticPolicyDataRaw_v1(data *staticPolicyData) *staticPolicyDataRaw_v1
...
 	// index.
-	keyLoaded, err := tpm.LoadExternal(nil, keyPublic, tpm2.HandleEndorsement)
+	keyLoaded, err := tpm.LoadExternal(nil, c.updateKey, tpm2.HandleEndorsement)
 	if err != nil {
 		return xerrors.Errorf("cannot load public part of key used to verify authorization signature: %w", err)
 	}
 	defer tpm.FlushContext(keyLoaded)
 
-	// Execute the policy assertions
-	if version == 0 {
```
pedronis, line 255: this if had very asymmetric sides to start with but the new code is even harder to follow because the logic is spread between a callback that is used only in one case and fixed field of the common struct etc.

tpm2/policy.go
```diff
@@ -151,189 +152,263 @@ func makeStaticPolicyDataRaw_v1(data *staticPolicyData) *staticPolicyDataRaw_v1
...
+			if err := cb(policySession); err != nil {
+				return err
+			}
+		}
+
+		if _, _, err := tpm.PolicySigned(keyLoaded, policySession, true, nil, nil, 0, &signature); err != nil {
 			return xerrors.Errorf("cannot execute assertion to increment counter: %w", err)
 		}
-		if err := tpm.PolicyNvWritten(policySession, true); err != nil {
+		if err := tpm.PolicyOR(policySession, c.authPolicies); err != nil {
```
pedronis, line 275: this PolicyOR is done once per for loop but always with the same value, I suppose it's needed this way but maybe it merits a comment

pedronis (commented):
the new code is a bit hard to follow tbh, the old was already a bit so because of v0 vs v1 differences. Anyway a first pass with some comments/questions

tpm2/seal.go
```diff
@@ -62,37 +62,26 @@ func makeImportableSealedKeyTemplate() *tpm2.Public {
 // the TPM and the new PCR policy will have a count of this value + 1. If tpm is nil then
```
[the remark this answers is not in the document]
chrisccoulson: I've updated this comment now.

tpm2/seal.go
```diff
@@ -585,15 +563,13 @@ func updateKeyPCRProtectionPolicyCommon(tpm *tpm2.TPMContext, keys []*SealedKeyO
...
-
 	// Compute a new dynamic authorization policy
 	if pcrProfile == nil {
 		pcrProfile = &PCRProtectionProfile{}
 	}
-	policyData, err := computeSealedKeyDynamicAuthPolicy(tpm, primaryData.version, primaryData.keyPublic.NameAlg, authPublicKey.NameAlg, authKey,
-		pcrPolicyCounterPub, v0PinIndexAuthPolicies, pcrProfile, session)
+	policyData, err := computeSealedKeyDynamicAuthPolicy(tpm, primaryData.version, primaryData.keyPublic.NameAlg,
+		primaryData.staticPolicyData.authPublicKey.NameAlg, authKey, pcrPolicyCounterPub, pcrProfile,
+		primaryData.dynamicPolicyData.policyCount+1, session)
```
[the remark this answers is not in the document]
chrisccoulson: I've added a comment here

tpm2/policy_test.go
chrisccoulson: I've added a test for that now - the delta will be larger than one if you call UpdatePCRProtectionPolicy more than once before calling RevokeOldPCRProtectionPolicies.

tpm2/policy.go
chrisccoulson: I've refactored this now. Whilst doing this, I also added a helper API to go-tpm2 for creating the signed assertions used by this code (https://pkg.go.dev/github.com/canonical/go-tpm2/util@master#SignPolicyAuthorization)

tpm2/policy.go
```diff
@@ -151,189 +152,263 @@ func makeStaticPolicyDataRaw_v1(data *staticPolicyData) *staticPolicyDataRaw_v1
...
-// NV index created by createPcrPolicyCounter (for current key files) or on a NV index created by (the now deleted)
-// createPinNVINdex for version 0 key files.
-//
-// This requires a signed authorization. For current key files, the keyPublic argument must correspond to the updateKeyName argument
-// originally passed to createPcrPolicyCounter. For version 0 key files, this must correspond to the key originally passed to
-// createPinNVIndex. The private part of that key must be supplied via the key argument. For version 0 key files, the authorization
-// policy digests returned from createPinNVIndex must be supplied via the nvAuthPolicies argument.
-func incrementPcrPolicyCounter(tpm *tpm2.TPMContext, version uint32, nvPublic *tpm2.NVPublic, nvAuthPolicies tpm2.DigestList, key crypto.PrivateKey, keyPublic *tpm2.Public, hmacSession tpm2.SessionContext) error {
-	index, err := tpm2.CreateNVIndexResourceContextFromPublic(nvPublic)
+func (c *pcrPolicyCounterCommon) increment(tpm *tpm2.TPMContext, from, to uint64, key crypto.PrivateKey, cb func(tpm2.SessionContext) error, session tpm2.SessionContext) error {
```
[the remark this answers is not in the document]
chrisccoulson: I've refactored this to avoid using the callback.

pedronis (approved):
thanks

PR #155, opened 2021-06-30
Switch to github actions

chrisccoulson:


.github/workflows/test.yaml
```diff
@@ -0,0 +1,42 @@
...
+        path: ./src/github.com/snapcore/secboot
+    - name: Install dependencies
+      run: |
+          sudo apt-get update
+          sudo apt-get install -y sbsigntool
+          sudo snap install core core18
+          sudo snap install --edge tpm2-simulator-chrisccoulson
+    - name: Install govendor
+      run: go get -u github.com/kardianos/govendor
+    - name: Cache Go dependencies
```
sergiocazzolato, line 29: Why do you need to cache go vendor file?

.github/workflows/test.yaml
```diff
@@ -0,0 +1,42 @@
...
+jobs:
+  tests:
+    runs-on: ubuntu-20.04
+    env:
+      GOPATH: ${{ github.workspace }}
+    steps:
+    - name: Set up Go 1.9
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.9
```
bboozzoo, line 16: perhaps it'd be useful to set up a strategy with go 1.9 and maybe the latest branch. Hopefully we'll move snapd to 1.13 soon, so it can replace 1.9 in this workflow too.

bboozzoo (approved):
LGTM

.github/workflows/test.yaml
chrisccoulson: This is just creating a cache of the dependencies, keyed by the hash of vendor/vendor.json

pedronis (approved):
thanks

PR #156, opened 2021-07-14
Big update to EFI tests

chrisccoulson:
The tests in ./efi/ depend on a lot of checked-in binary data,
all of which is either not really reproducible, or only manually
reproducible following steps that are mostly stored in my head.
This makes it difficult to update or add to the testdata when
required - a task that currently can only really be done by me.

This makes several changes:
- Adds a efi/testdata/src directory containing a fairly minimal set
of sources for testdata. This is a mix of public certs, source code
for mock EFI binaries, public uefi.org dbx updates, checked in test
RSA private keys and an eventlog from a Dell box to verify a specific
bug fix.
- Adds tools/make-efi-testdata to (mostly reproducibly) compile
the actual test data in efi/testdata. The compiled testdata is still
checked in because reproducibility of the mock EFI binaries requires
the use of the same toolchain, but it can be updated or re-created
by anyone. This tool creates mock event logs, mock EFI variable stores,
mock EFI binaries, mock signature database updates and creates mock
certificates from the checked-in RSA private keys.
- Updates the tests to use gopkg.in/check.v1, with some updates to
the actual tests performed.

These updates did find one bug - we would compute the wrong digest
where grub and the kernel are both authenticated using a CA stored
in the UEFI authorized signature database with the latest shim,
because the status of a flag was not correctly propagated between
two steps.

This requires updating both go-efilib and tcglog-parser. This required
quite a bit of work in both projects to support creating the test
data.

One limitation of this is that signing EFI binaries and signature
database updates is not quite reproducible because the signatures
contain timestamps. A refresh of the testdata will always create files
with a different hash, but the different file hashes don't affect the
actual tests.


pedronis:
@chrisccoulson 

I get errors like these when trying to run tests here:

```
 ./run-tests --with-mssim
# github.com/snapcore/secboot/tools/gen-compattest-data [github.com/snapcore/secboot.test]
tools/gen-compattest-data/main.go:64: undefined: tcglog.ParseLog
FAIL	github.com/snapcore/secboot [build failed]
# github.com/snapcore/secboot/tools/gen-compattest-data [github.com/snapcore/secboot/efi.test]
tools/gen-compattest-data/main.go:64: undefined: tcglog.ParseLog
FAIL	github.com/snapcore/secboot/efi [build failed]
# github.com/snapcore/secboot/tools/gen-compattest-data [github.com/snapcore/secboot/internal/compattest.test]
tools/gen-compattest-data/main.go:64: undefined: tcglog.ParseLog
FAIL	github.com/snapcore/secboot/internal/compattest [build failed]
# github.com/snapcore/secboot/tools/gen-compattest-data [github.com/snapcore/secboot/internal/keyring.test]
tools/gen-compattest-data/main.go:64: undefined: tcglog.ParseLog
FAIL	github.com/snapcore/secboot/internal/keyring [build failed]
# github.com/snapcore/secboot/tools/gen-compattest-data [github.com/snapcore/secboot/internal/luks2.test]
tools/gen-compattest-data/main.go:64: undefined: tcglog.ParseLog
FAIL	github.com/snapcore/secboot/internal/luks2 [build failed]
...
```

tools/make-efi-testdata/main.go
```diff
@@ -0,0 +1,65 @@
+package main
```
pedronis, line 1: this directory probably merits a README file

efi/testdata/src/shim/Makefile
```diff
@@ -0,0 +1,42 @@
+TOPDIR	?= $(shell pwd)
```
pedronis, line 1: this directory probably merits also a README

efi/testdata/src/app/Makefile
```diff
@@ -0,0 +1,25 @@
+TOPDIR	?= $(shell pwd)
```
pedronis, line 1: same

pedronis (commented):
got a bit blocked on tests not running. left some initial comments about having some documentation README for the new testdata/src directories with code

tools/make-efi-testdata/README
```diff
@@ -0,0 +1,22 @@
...
+
+Note that recreating the EFI binaries and signature databases
+is not entirely reproducible because their signatures include
+timestamps. This means they are recreated with different file
+hashes, but this doesn't affect the actual tests.
+
+Reproducibility of test data is dependent on the environment
+in which the data is built. This is particularly true for the
+EFI binaries, which are only reproducible when built with the
+same toolchain.
```
pedronis, line 20: I tried to rebuild the testdata and of course the tests then failed.

tools/make-efi-testdata/README
```diff
@@ -0,0 +1,22 @@
...
+
+This tool depends on make, sbsigntools, gcc and binutils.
```
pedronis, line 22: maybe here or somewhere you should list the actual package versions you used to build the checked-in testdata. I suppose gcc and binutils make a more material difference than make and sbsigntools

pedronis (commented):
one more request

tools/make-efi-testdata/README
chrisccoulson: Was it the tests in efi/bootmanager_policy_test.go that failed?

pedronis: yes, but secureboot_policy_test.go as well

tools/make-efi-testdata/README
chrisccoulson: I've updated the tool to record details about the build environment in yaml format, and checked that in now.

pedronis (approved):
thank you

PR #157, opened 2021-07-14
Add internal/argon2 package

chrisccoulson:
This adds an internal package for deriving keys from passphrases
using the Argon2i algorithm, and will be used for passphrase support.

There are 2 core APIs:
- Key is just a wrapper around the argon2.Key API provided by the
golang.org/x/crypto/argon2 package, but using the CostParams type
returned by Benchmark.
- Benchmark determines the cost parameters for the algorithm based
on the supplied target duration, maximum memory cost and key length.
It is based on the algorithm used by cryptsetup, which prioritizes
memory cost over time cost.

The algorithm consumes a lot of memory and isn't suitable for long
lived system processes in a garbage collected language. For the
secboot public APIs that execute argon2, I'll add an interface that
allows the caller to either run the algorithm in-process if
appropriate, or delegate it to a short-lived utility process that can
be defined outside of secboot.

Benchmark runs the algorithm multiple times and can cause the current
process to rapidly run out of memory. To support delegating this to
a short-lived utility process, it provides a callback argument for
running each actual measurement and returning a duration. The
KeyDuration function provides the in-process implementation of this,
and it is expected that this will be executed in a short-lived utility
process using a mechanism defined outside of this package. It is
possible to run the entire benchmark in-process by running the
garbage collector after each measurement, and this is what the unit
tests are doing, but this isn't really appropriate for production
code.


internal/argon2/argon2.go
```diff
@@ -0,0 +1,293 @@
...
+// It may be desirable to execute it in a short-lived utility process.
+func KeyDuration(params *CostParams, keyLen uint32) time.Duration {
+	start := time.Now()
+	Key(benchmarkPassword, benchmarkSalt, params, keyLen)
+	return time.Now().Sub(start)
+}
+
+// KeyDurationFn provides a mechanism to delegate key derivation measurements
+// to a short-lived utility process during benchmarking.
+type KeyDurationFn func(params *CostParams, keyLen uint32) (time.Duration, error)
```
pedronis, line 253: go naming more typically would have s/Fn/Func/ here

internal/argon2/argon2.go
```diff
@@ -0,0 +1,293 @@
...
+// The package hard codes a minimum time cost of 4 iterations, and a minimum
+// memory cost of 32MiB. A maximum memory cost of half of the total RAM or 4GB
+// provides a ceiling to the supplied maximum memory cost. The algorithm will
+// set 1 thread per CPU, up to a limit of 4 threads.
+//
+// The supplied callback is used to actually run the key derivation measurement,
+// which will consume a lot of memory depending on the supplied parameters. Each
+// measurement should generally be delegated to a short-lived utility process. If
+// the measurement is performed in the current process, the garbage collector must
+// be executed at the end of each measurement.
```
pedronis, line 275: this should probably mention KeyDuration

chrisccoulson:
> the benchmark tests can fail when running with coverage it seems:
> 
> ```
> $go test -cover
> 
> ----------------------------------------------------------------------
> FAIL: argon2_test.go:88: argon2Suite.TestBenchmark2s
> 
> argon2_test.go:89:
>     s.testBenchmark(c, &testBenchmarkData{
>         params: &BenchmarkParams{
>             MaxMemoryCostKiB: 1 * 1024 * 1024,
>             TargetDuration:   2 * time.Second},
>         keyLen: 32})
> argon2_test.go:85:
>     c.Check(int(elapsed/time.Millisecond), testutil.IntLessThan, int(float64(data.params.TargetDuration/time.Millisecond)*1.2))
> ... a int = 2720
> ... b int = 2400
> ... relation 2720 < 2400 is not true
> 
> 
> ----------------------------------------------------------------------
> FAIL: argon2_test.go:104: argon2Suite.TestBenchmarkReducedMemory
> 
> argon2_test.go:105:
>     s.testBenchmark(c, &testBenchmarkData{
>         params: &BenchmarkParams{TargetDuration: 2 * time.Second},
>         keyLen: 32})
> argon2_test.go:85:
>     c.Check(int(elapsed/time.Millisecond), testutil.IntLessThan, int(float64(data.params.TargetDuration/time.Millisecond)*1.2))
> ... a int = 2618
> ... b int = 2400
> ... relation 2618 < 2400 is not true
> 
> OOPS: 8 passed, 2 FAILED
> --- FAIL: Test (27.12s)
> FAIL
> coverage: 84.4% of statements
> exit status 1
> FAIL	github.com/snapcore/secboot/internal/argon2	27.211s
> ```

I'm not sure how best to fix this - it can be a bit intermittent and some of the cryptsetup tests in internal/luks2 have the same problem as this as well. In the internal/luks2 case, we want to test that we're passing the right target duration to cryptsetup using the correct flag, and the only way to do that really is to time the performance of the KDF afterwards.

In this case, we actually want to test that the benchmark works, and I can't think of a way to do that properly without timing it in the same way which adds an element of flakiness - looking at the computed cost parameters isn't sufficient because they are context dependent (they'll be different depending on what hardware you run the tests on).

pedronis (approved):
thank you

PR #158, opened 2021-07-15
Remove tpm2 PIN support

chrisccoulson:
This has never been used in UC20, and removing it will help
facilitate the existing (soon-to-be-legacy) TPM sealed objects
to use the secboot.ActivateVolumeWithKeyData API, which will
allot tpm2/crypt.go to be deleted.

Passphrase support is going to be added to secboot.KeyData, and
it will primarily encrypt the platform encrypted key data with a
passphrase derived key. This will integrate with the platform
implementation, which in the TPM case will be by setting the sealed
key object's authorization value to some key derived from the
passphrase. As the authorization policy for this will look the same
as today, I've left the PIN-specific parts of the authorization
policy in place, which will prevent additional complexity later on.


pedronis (approved):
+1

PR #159, opened 2021-07-19
tpm2: Make ReadSealedKeyObject take a io.Reader

chrisccoulson:
In order to create a tpm2 platform implementation for KeyData,
SealedKeyObject needs to be storage agnostic instead of having
a hard dependency on file paths.

This makes ReadSealedKeyObject take an io.Reader.

SealedKeyObject.UpdatePCRProtectionPolicy now only modifies the
in-memory SealedKeyObject. SealedKeyObject.WriteAtomic has been
introduced for persisting changes to some storage mechanism.

Two additional APIs have been introduced to work with existing
sealed key files - ReadSealedKeyObjectFromFile and
NewFileSealedKeyObjectWriter.

Existing sealed key files use a AF-splitter. This has limited use
on solid state drives and was introduced to make PIN changes more
secure with key data that is stored inside a filesystem. For that
reason, the generic ReadSealedKeyObject and
SealedKeyObject.WriteAtomic do not use the AF-splitter - this is
implemented only for the file reader / writer. Note that PIN support
is removed in a previous PR anyway.

The key data metadata needs to be versioned regardless of how it
is stored. One challenge here is that this metadata version is
stored before the metadata for the AF-slitter, which is specific
to key data files. To work around this, the file reader does some
gymnastics to ensure that the stream read by ReadSealedKeyObject
contains this version. Similarly, the file writer performs the
reverse gymnastics to ensure that the version gets written to the
correct place.

This doesn't modify the existing SealKeyToTPM API yet - this API
will continue to only create files for now. I did consider making
SealKeyToTPM return a SealedKeyObject that then has to be written
somewhere persistent, but this API is going to look different when
creating new style sealed keys compatible with the secboot.KeyData
API - it will return a secboot.KeyCreationData that can be passed
to secboot.NewKeyData. I didn't see the point in changing this API
twice.

### DIFF
--- a/internal/compattest/compattest_test.go
+++ b/internal/compattest/compattest_test.go
@@ -151,7 +151,7 @@ func (s *compatTestSuiteBase) copyFile(c *C, path string) string {
 }
 
 func (s *compatTestSuiteBase) testUnsealCommon(c *C) {
-	k, err := secboot_tpm2.ReadSealedKeyObject(s.absPath("key"))
+	k, err := secboot_tpm2.ReadSealedKeyObjectFromFile(s.absPath("key"))
 	c.Assert(err, IsNil)
 
 	key, authPrivateKey, err := k.UnsealFromTPM(s.TPM)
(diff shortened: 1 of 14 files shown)


pedronis (commented):
lgtm

pedronis (approved):

PR #160, opened 2021-07-28
Ensure keyring tests are skipped in unsuitable environments

chrisccoulson:
Some tests that touch the keyring are failing since the switch
to github actions. There is a check in other suites to skip
keyring tests in environments that are unsuitable, but not in
the tests that are failing. This adds that check there too.

(We need to figure out how to make these tests work in
github actions, so this is a temporary fix for now just to
get the tests to pass).

Note that I also changed the go version to 1.9 (from 1.10)
when switching to github actions because that was the
minimum required by snapd, but that makes go vet fail.
I've changed the go version to 1.13 now that is the
minimum required by snapd.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -10,10 +10,10 @@ jobs:
     env:
       GOPATH: ${{ github.workspace }}
     steps:
-    - name: Set up Go 1.9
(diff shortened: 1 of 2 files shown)


pedronis (approved):
thanks

PR #161, opened 2021-07-28
tpm2: Add "legacy" KeyData platform implementation

chrisccoulson:
This adds a KeyData platform implementation to allow
keys to be recovered from the current (soon to be legacy)
TPM sealed key files using the new secboot.KeyData API. This
permits the use of the secboot.ActivateVolumeWithKeyData API
for TPM sealed key files, allowing the code in tpm2/crypt.go
to be deleted.

In order to do this, a new function for reading TPM sealed key
files is introduced - NewKeyDataFromSealedKeyObjectFile, which
constructs a secboot.KeyData from a sealed key file. Note that
the snap model authorization API can't be used for these KeyData
instances, and snap-bootstrap shouldn't use this API for files
opened via it. That's ok though, because the PCR profile contains
an assertion of the snap model.

When the tpm2 package becomes a proper KeyData platform, it's
expected that these newer KeyData's will use the snap model
authorization API like any other platform. We need to be careful
to ensure that the new KeyData can't be manipulated in such a way
to trick snap-bootstrap into reading it via the (what will be
legacy) NewKeyDataFromSealedKeyObjectFile API, recovering the key
successfully via the legacy platform and skipping the snap model check.

The deleted code in tpm2/crypt.go attempted to create a primary key
if the initial unseal failed, before attempting it one more time.
To preserve this behaviour, the code is now part of UnsealFromTPM.


tpm2/platform_legacy.go
```diff
@@ -0,0 +1,132 @@
...
+
+	return secboot.MarshalKeys(key, secboot.AuxiliaryKey(authKey)), nil
+}
+
+// NewKeyDataFromSealedKeyObjectFile creates a secboot.KeyData for the TPM
+// sealed key object at the supplied path, in order to enable keys to be
+// recovered from the TPM sealed key object using the secboot.KeyData API.
+//
+// Note that the returned KeyData does not support the snap model authorization
+// API, and consumers of this function should not attempt to use this API.
```
pedronis, line 96: this needs to be known implicitly ATM?  a modchecker will still be returned?

tpm2/unseal.go
```diff
@@ -28,6 +30,64 @@ import (
...
+			if tpm2.IsResourceUnavailableError(err, tcg.SRKHandle) {
+				// No SRK - save the error and try creating a transient
+				lastError = ErrTPMProvisioning
+				continue
+			} else if err != nil {
+				// This is an unexpected error
+				return nil, xerrors.Errorf("cannot create context for SRK: %w", err)
+			}
+		} else {
+			var err error
```
pedronis, line 55: is this the code that correspond to the code you mention in description that was in crypt.go?  could you add a comment here a bot the purpose of this?

pedronis (commented):
looks good but some questions

tpm2/unseal.go
chrisccoulson: It is - I've added some documentation to this function now.

tpm2/platform_legacy.go
chrisccoulson: That's right - if this API is used to load the key data, then IsModelAuthorized shouldn't be used because it will return false.

When we have new-style TPM objects, I can add a function to determine which API to use to load a file.

PR #163, opened 2021-08-02
crypt: add pbkdf options for LUKS container manipulation

kubiko:
add support for additional pbkd options to be passed for operations on LUKS container.

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -643,6 +643,21 @@ type InitializeLUKS2ContainerOptions struct {
 	KeyslotsAreaKiBSize int
 }
 
+// AddRecoveryKeyToLUKS2ContainerOptions carries options for recovery key to LUKS2
+// containers.
+type PbkdfLUKS2ContainerOptions struct {
+	// PbkdfMmemory set the memory cost for PBKDF (for Argon2i/id the number
+	// represents kilobytes).  Note that it is maximal value,
+	// PBKDF benchmark or available physical memory can decrease it.
+	PbkdfMmemory int
+	// Avoid PBKDF benchmark and set time cost (iterations) directly.
+	PkdfForceIterations int
+	// Set the parallel cost for PBKDF (number of threads, up to 4).
+	// Note that it is maximal value, it is decreased
+	// automatically if CPU online count is lower.
+	PbkdfParallel int
+}
+
 func validateInitializeLUKS2Options(options *InitializeLUKS2ContainerOptions) error {
 	if options == nil {
 		return nil
@@ -688,7 +703,7 @@ func validateInitializeLUKS2Options(options *InitializeLUKS2ContainerOptions) er
 //
 // WARNING: This function is destructive. Calling this on an existing LUKS container will make the data contained inside of it
 // irretrievable.
-func InitializeLUKS2Container(devicePath, label string, key []byte, options *InitializeLUKS2ContainerOptions) error {
+func InitializeLUKS2Container(devicePath, label string, key []byte, options *InitializeLUKS2ContainerOptions, pbkdfOptions *PbkdfLUKS2ContainerOptions) error {
 	if len(key) < 32 {
 		return fmt.Errorf("expected a key length of at least 256-bits (got %d)", len(key)*8)
 	}
@@ -711,7 +726,7 @@ func InitializeLUKS2Container(devicePath, label string, key []byte, options *Ini
 		// entropy of at least 32 bytes, and increased cost doesn't provide a security benefit because
 		// this key and these settings are already more secure than the recovery key. Increased cost
 		// here only slows down unlocking.
-		"--pbkdf", "argon2i", "--iter-time", "100",
+		"--pbkdf", "argon2i",
 		// set LUKS2 label
 		"--label", label,
 	}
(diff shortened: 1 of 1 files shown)


crypt.go
```diff
@@ -725,6 +740,24 @@ func InitializeLUKS2Container(devicePath, label string, key []byte, options *Ini
 				"--luks2-keyslots-size", fmt.Sprintf("%dk", options.KeyslotsAreaKiBSize))
 		}
 	}
+	if pbkdfOptions != nil {
+		if pbkdfOptions.PbkdfMmemory != 0 {
+			args = append(args,
+				"--pbkdf-memory", fmt.Sprintf("%d", pbkdfOptions.PbkdfMmemory))
```
pedronis, line 746: instead of Sprintf this can use strconv.Itoa

crypt.go
```diff
@@ -643,6 +643,21 @@ type InitializeLUKS2ContainerOptions struct {
 	KeyslotsAreaKiBSize int
 }
 
+// AddRecoveryKeyToLUKS2ContainerOptions carries options for recovery key to LUKS2
+// containers.
+type PbkdfLUKS2ContainerOptions struct {
```
pedronis, line 648: by naming conventions this should be PBKDFLUKS2...

crypt.go
```diff
@@ -725,6 +740,24 @@ func InitializeLUKS2Container(devicePath, label string, key []byte, options *Ini
...
+			args = append(args,
+				"--pbkdf-memory", fmt.Sprintf("%d", pbkdfOptions.PbkdfMmemory))
+		}
+		if pbkdfOptions.PkdfForceIterations != 0 {
+			args = append(args,
+				"--pbkdf-force-iterations", fmt.Sprintf("%d", pbkdfOptions.PkdfForceIterations))
+		} else {
+			args = append(args, "--iter-time", "100")
+		}
+		if pbkdfOptions.PbkdfParallel != 0 {
```
pedronis, line 754: I'm not entirely sure why this is not under the if/else ?  I let @chrisccoulson comment/double check on this

crypt.go
```diff
@@ -725,6 +740,24 @@ func InitializeLUKS2Container(devicePath, label string, key []byte, options *Ini
 				"--luks2-keyslots-size", fmt.Sprintf("%dk", options.KeyslotsAreaKiBSize))
 		}
 	}
+	if pbkdfOptions != nil {
```
pedronis, line 743: this logic here and repeated in the other functions could be  extracted as a method on *PbkdfLUKS2ContainerOptions

```
func (o *PbkdfLUKS2ContainerOptions) appendToArgs(args []string, fallbackIterTimeMs int) []string {
   ...
}
```

crypt.go
```diff
@@ -643,6 +643,21 @@ type InitializeLUKS2ContainerOptions struct {
...
+	// PbkdfMmemory set the memory cost for PBKDF (for Argon2i/id the number
+	// represents kilobytes).  Note that it is maximal value,
+	// PBKDF benchmark or available physical memory can decrease it.
+	PbkdfMmemory int
+	// Avoid PBKDF benchmark and set time cost (iterations) directly.
+	PkdfForceIterations int
+	// Set the parallel cost for PBKDF (number of threads, up to 4).
+	// Note that it is maximal value, it is decreased
+	// automatically if CPU online count is lower.
+	PbkdfParallel int
```
pedronis, line 658: as it's already in the struct name we can leave out the Pkdf prefixes here

pedronis (commented):
did a pass

crypt.go
```diff
@@ -795,10 +828,33 @@ func addKeyToLUKS2Container(devicePath string, existingKey, key []byte, extraOpt
...
-func AddRecoveryKeyToLUKS2Container(devicePath string, key []byte, recoveryKey RecoveryKey) error {
-	return addKeyToLUKS2Container(devicePath, key, recoveryKey[:], []string{
-		// use argon2i as the KDF with an increased cost
-		"--pbkdf", "argon2i", "--iter-time", "5000"})
+func AddRecoveryKeyToLUKS2Container(devicePath string, key []byte, recoveryKey RecoveryKey, pbkdfOptions *PbkdfLUKS2ContainerOptions) error {
+	args := []string{
+		// use argon2i as the KDF with reduced cost. This is done because the supplied input key has an
+		// entropy of at least 32 bytes, and increased cost doesn't provide a security benefit because
+		// this key and these settings are already more secure than the recovery key. Increased cost
+		// here only slows down unlocking.
```
chrisccoulson, line 836: I don't think replacing the original comment was intentional.

crypt.go
chrisccoulson, line 658: Is an option to configure the parallelism necessary?

chrisccoulson (commented):
Thanks for this - I've left a few comments inline. Were you planning to add some tests for this?

crypt.go
```diff
@@ -725,6 +740,24 @@ func InitializeLUKS2Container(devicePath, label string, key []byte, options *Ini
...
+	if pbkdfOptions != nil {
+		if pbkdfOptions.PbkdfMmemory != 0 {
+			args = append(args,
+				"--pbkdf-memory", fmt.Sprintf("%d", pbkdfOptions.PbkdfMmemory))
+		}
+		if pbkdfOptions.PkdfForceIterations != 0 {
+			args = append(args,
+				"--pbkdf-force-iterations", fmt.Sprintf("%d", pbkdfOptions.PkdfForceIterations))
+		} else {
+			args = append(args, "--iter-time", "100")
```
pedronis, line 752: we should also make iter-itme itself an option (and use the defaults if it's unset as well)

crypt.go
```diff
@@ -795,10 +828,33 @@ func addKeyToLUKS2Container(devicePath string, existingKey, key []byte, extraOpt
...
+		}
+		if pbkdfOptions.PbkdfParallel != 0 {
+			args = append(args,
+				"--pbkdf-parallel", fmt.Sprintf("%d", pbkdfOptions.PbkdfParallel))
+		}
+	} else {
+		args = append(args, "--iter-time", "5000")
```
pedronis, line 855: same

pedronis (commented):
we should also let specify the actual iter-time possibly from outside

crypt.go
pedronis: for this one maybe is not so important

crypt.go
pedronis: here is useful though

PR #164, opened 2021-08-05
crypt: add KDF options to API

chrisccoulson:
This is based on https://github.com/snapcore/secboot/pull/163 with some of the comments addressed.

PR #165, opened 2021-08-13
internal/luks2: Add KDFOptions for Format and AddKey

chrisccoulson:
This ports part of https://github.com/snapcore/secboot/pull/164 to
the internal/luks2 package. There will be a follow up to expose
this via the public APIs, but I've got some other changes planned
for those as well.


pedronis (approved):
thanks

PR #167, opened 2021-08-19
internal/luks2: Add support for external token implementations

chrisccoulson:
Only 2 token fields are required for all tokens (type + keyslots),
with remaining fields being type specific. To handle this, tokens
are represented by a struct containing the 2 required fields and
a map[string]interface{} to handle the remaining fields.

Working with a map[string]interface{} type is a pain, so introduce
a way to decode to type-specific token representations instead.
Handlers for these are registered by calling RegisterTokenHandler
from an external package.

### DIFF
--- a/internal/luks2/metadata.go
+++ b/internal/luks2/metadata.go
@@ -298,6 +298,12 @@ const (
 	KeyslotTypeLUKS2 KeyslotType = "luks2"
 )
 
+type TokenType string
+
+const (
+	TokenTypeKeyring TokenType = "luks2-keyring"
+)
+
 // AFType corresponds to an anti-forensic splitter algorithm.
 type AFType string
 
@@ -341,13 +347,16 @@ type binaryHdr struct {
 	Padding4096 [7 * 512]byte
 }
 
-type luksJsonNumber string
+// JsonNumber represents a JSON number literal. It is similar to
+// json.Number but supports uint64 and int literals as required by
+// the LUKS2 specification.
+type JsonNumber string
 
-func (n luksJsonNumber) int() (int, error) {
+func (n JsonNumber) Int() (int, error) {
 	return strconv.Atoi(string(n))
 }
 
-func (n luksJsonNumber) uint64() (uint64, error) {
+func (n JsonNumber) Uint64() (uint64, error) {
 	return strconv.ParseUint(string(n), 10, 64)
 }
 
@@ -361,8 +370,8 @@ type Config struct {
 
 func (c *Config) UnmarshalJSON(data []byte) error {
 	var d struct {
-		JSONSize     luksJsonNumber `json:"json_size"`
-		KeyslotsSize luksJsonNumber `json:"keyslots_size"`
+		JSONSize     JsonNumber `json:"json_size"`
+		KeyslotsSize JsonNumber `json:"keyslots_size"`
 		Flags        []string
 		Requirements []string
 	}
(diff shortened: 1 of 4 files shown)


internal/luks2/metadata.go
```diff
@@ -388,64 +397,96 @@ func (c *Config) UnmarshalJSON(data []byte) error {
...
+// ReadHeader will be represented by a type-specific implementation if a
+// TokenHandler is registered for it, or GenericToken.
+type Token interface {
+	GetType() TokenType // Token type ("luks2-" prefixed types are reserved for cryptsetup)
+	GetKeyslots() []int // Keyslots assigned to this token
+}
+
+// TokenHandler provides a mechanism for an external package to decode
+// custom token types.
+type TokenHandler func([]byte) (Token, error)
```
pedronis, line 429: should this be called TokenDecoder or TokenInterpreter for a bit more clarity?

internal/luks2/metadata.go
```diff
@@ -388,64 +397,96 @@ func (c *Config) UnmarshalJSON(data []byte) error {
...
+
 // Token corresponds to a token object in the JSON metadata of a LUKS2 volume. It
-// describes how to retrieve a passphrase or key for a keyslot.
-type Token struct {
-	Type     string                 // Token type ("luks2-" prefixed types are reserved for cryptsetup)
+// describes how to retrieve a passphrase or key for a keyslot. Tokens decoded by
+// ReadHeader will be represented by a type-specific implementation if a
+// TokenHandler is registered for it, or GenericToken.
+type Token interface {
+	GetType() TokenType // Token type ("luks2-" prefixed types are reserved for cryptsetup)
```
pedronis, line 423: golang naming convention would have these called Type() and Keyslots() and then use uglier/longer names for the concrete struct field names

internal/luks2/metadata.go
```diff
@@ -910,3 +964,7 @@ func ReadHeader(path string, lockMode LockMode) (*HeaderInfo, error) {
 		Label:      hdr.Label.String(),
 		Metadata:   *metadata}, nil
 }
+
+func RegisterTokenHandler(typ TokenType, handler TokenHandler) {
```
pedronis, line 968: this needs a doc comment

internal/luks2/metadata.go
```diff
@@ -388,64 +397,96 @@ func (c *Config) UnmarshalJSON(data []byte) error {
...
+// TokenHandler provides a mechanism for an external package to decode
+// custom token types.
+type TokenHandler func([]byte) (Token, error)
+
+var tokenHandlers = make(map[TokenType]TokenHandler)
+
+// GenericToken corresponds to a token that doesn't have a more type-specific
+// representation.
+type GenericToken struct {
+	Type     TokenType              // Token type ("luks2-" prefixed types are reserved for cryptsetup)
```
pedronis, line 436: this could be TokenType and KeyslotIndexes for example or something like that

pedronis (commented):
lgtm but some naming considerations

internal/luks2/metadata.go
chrisccoulson: Yes, I think TokenDecoder is a bit better than TokenHandler.

internal/luks2/metadata.go
chrisccoulson: I've renamed these now to TokenType and TokenKeyslots.

internal/luks2/metadata.go
chrisccoulson: I've added documentation for this now.

internal/luks2/metadata.go
```diff
@@ -965,6 +965,9 @@ func ReadHeader(path string, lockMode LockMode) (*HeaderInfo, error) {
 		Metadata:   *metadata}, nil
 }
 
-func RegisterTokenHandler(typ TokenType, handler TokenHandler) {
-	tokenHandlers[typ] = handler
+// RegisterTokenDecoder registers a custom decoder for the specified token type,
+// in order for external packages to be able to create type-spcific token structures
```
pedronis, line 969: typo: specific

pedronis (approved):
lgtm but there's a typo in the new doc comment

PR #171, opened 2021-08-25
efi: correctly handle optional dbt and dbr signature databases

chrisccoulson:
The TCG PC Client Platform Firmware Profile Specification says
that the firmware must always measure the contents of the
SecureBoot, PK, KEK, db (authorized signature database) and dbx
(forbidden signature database) variables to PCR7 even if they
are empty.

The UEFI specification defines 2 additional signature databases:
- dbt (timestamp signature database, containing entities authorized
to cross-sign binaries for timestamp based revocation on firmware
that supports this).
- dbr (recovery signature database, containing entities authorized
to create OsRecovery variables for booting recovery software on
firmware that supports this).

The TCG PC Client PFP spec says that these optional databases must
be measured to PCR7 if they exist and are not empty.

The current code will just replay values from the event log if
they are present, which works but isn't quite correct. As with other
signature databases, they are updateable from the OS with a properly
signed payload and so the current variable value is authoritative
rather than the one that appears in the event log.

Handle this by adding dbt and dbr events to the precomputed PCR
profile using the current variable contents if the variables exist
and aren't empty.


chrisccoulson:
Closing this as the code has long since been deleted and it's not the right way to support it. https://github.com/snapcore/secboot/issues/306 is tracking this.

PR #173, opened 2021-09-01
tpm2: TPMConnection got renamed to just Connection.

xnox:
tpm2: TPMConnection got renamed to just Connection.


chrisccoulson (approved):
Thanks, this looks ok to me

PR #174, opened 2021-09-02
crypt: add KDF options to API

chrisccoulson:
This ports https://github.com/snapcore/secboot/pull/164 to master.

As with the code that landed previously, there are no new tests for
the additional functionality yet. I've got some other work in progress
that touches these APIs and that work will incorporate the appropriate
tests.

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -458,6 +458,38 @@ func DeactivateVolume(volumeName string) error {
 	return luks2Deactivate(volumeName)
 }
 
+// KDFOptions specifies parameters for the Argon2 KDF used by cryptsetup.
+type KDFOptions struct {
+	// MemoryKiB specifies the maximum memory cost in KiB when ForceIterations
+	// is zero. If ForceIterations is not zero, then this is used as the
+	// memory cost.
+	MemoryKiB int
+
+	// TargetDuration specifies the target duration for the KDF which
+	// is used to benchmark the time and memory cost parameters. If it
+	// is zero then the cryptsetup default is used. If ForceIterations
+	// is not zero then this field is ignored.
+	TargetDuration time.Duration
+
+	// ForceIterations can be used to turn off KDF benchmarking by
+	// setting the time cost directly. If this is zero then the cost
+	// parameters are benchmarked based on the value of TargetDuration.
+	ForceIterations int
+
+	// Parallel sets the maximum number of parallel threads for the
+	// KDF (up to 4). Cryptsetup will adjust this downwards based on
+	// the actual number of CPUs.
+	Parallel int
+}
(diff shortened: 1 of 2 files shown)


crypt.go
```diff
@@ -471,6 +503,17 @@ type InitializeLUKS2ContainerOptions struct {
 	// expressed in multiples of 1024 bytes. The value must be aligned to
 	// 4096 bytes, with the maximum size of 128MB.
 	KeyslotsAreaKiBSize int
+
+	// KDFOptions sets the KDF options for the initial keyslot. If this
+	// is nil then the defaults are used.
+	KDFOptions *KDFOptions
+}
+
+func (o *InitializeLUKS2ContainerOptions) formatOpts() *luks2.FormatOptions {
```
pedronis, line 512: given usage, this probably could just return the struct instead of a pointer

crypt.go
```diff
@@ -522,22 +565,28 @@ func InitializeLUKS2Container(devicePath, label string, key []byte, options *Ini
...
+	// does slow down unlocking. If an adversary is going to attempt to brute force this key,
+	// then they could instead turn their attention to one of the other keys involved in the
+	// protection of this key, some of which can be verified without running a KDF. For
+	// example, with a TPM sealed object, you can verify the parent storage key's seed by
+	// computing the key object's HMAC key and verifying the integrity value on the outer wrapper.
+	defaultKdfOptions := &KDFOptions{TargetDuration: 100 * time.Millisecond}
+	if options == nil {
+		options = &InitializeLUKS2ContainerOptions{KDFOptions: defaultKdfOptions}
+	}
+	if options.KDFOptions == nil {
```
pedronis, line 581: this could be an `else if`

pedronis (approved):
lgtm, couple small comments

crypt.go
```diff
@@ -458,6 +458,38 @@ func DeactivateVolume(volumeName string) error {
...
+	// parameters are benchmarked based on the value of TargetDuration.
+	ForceIterations int
+
+	// Parallel sets the maximum number of parallel threads for the
+	// KDF (up to 4). Cryptsetup will adjust this downwards based on
+	// the actual number of CPUs.
+	Parallel int
+}
+
+func (o *KDFOptions) internalOpts() *luks2.KDFOptions {
```
pedronis, line 485: given the usage this could return the struct instead of a pointer

crypt.go
chrisccoulson: Thanks, I've updated this now.

crypt.go
chrisccoulson: I've changed this to return a value

PR #177, opened 2021-09-29
tpm2: remove methods from keyData

chrisccoulson:
The internal keyData struct has a bunch of methods because
once-upon-a-time, SealedKeyObject was only created for unsealing
and we only created keyData internall for other operations.
Now that's SealedKeyObject is used for everything, just make the
methods of keyData unexported methods of SealedKeyObject, as the
current split appears to be completely arbitrary now.


pedronis (approved):
lgtm

PR #178, opened 2021-10-01
tpm2: refactor key data validation

chrisccoulson:
SealedKeyObject has a mechanism to perform some consistency
checks of the key data, but this function is huge and has lots
of branches in it to handle the different metadata versions.
Abstract the validation behind an interface that has version
specific implementations to make it a bit more readable.

This also splits validation of the key data and validation of
the auth key into 2 distinct steps, as the latter is only
required before signing PCR policy updates.


chrisccoulson:
Actually, let's leave this one for now - I started also refactoring the complicated code in tpm2/policy.go to make it a bit more readable, and I might need to think about them both together.

pedronis:
@chrisccoulson I removed it from my queue for now then

chrisccoulson:
Closing as this was superceded by https://github.com/snapcore/secboot/pull/179

PR #179, opened 2021-10-06
tpm2: refactor keyData

chrisccoulson:
SealedKeyObject is just a container around keyData, which
then has several on-disk representations (keyDataRaw_v*). It
also has a mechanism to perform some consistency checks of the
key data (SealedKeyData.validate), but this function is huge
and has branches in it to handle the different metadata versions.

This turns keyData into an interface. Each metadata version
is represented by an implementation of it, with each
implementation containing the version-specific functionality.
Each keyData implementation has its own version-specific
validate function, hopefully making the code a bit more
readable.

This makes the conversion from the on-disk representation
(keyDataRaw_v*) to keyData unnecessary.

This also splits validation of the key data and validation of
the auth key into 2 distinct steps, as the latter is only
required before signing PCR policy updates.

A subsequent PR is going to create an abstraction for some
of the code in tpm2/policy.go, which has similar issues.

### DIFF
--- a/internal/tpm2test/tpm.go
+++ b/internal/tpm2test/tpm.go
@@ -235,8 +235,9 @@ func newTPMConnectionFromExisting(tpm *secboot_tpm2.Connection, tcti *TCTI) (*se
 // NewTPMConnectionFromExistingT creates a new connection and TCTI from the
 // supplied ones. This is useful in scenarios where test fixture setup and
 // test execution require a different connection. The returned connection
-// uses the same underlying connection as the one supplied.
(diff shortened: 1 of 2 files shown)


chrisccoulson:
This is a replacement for https://github.com/snapcore/secboot/pull/178.

I have a WIP branch based on this that does something similar for policy.go, although I haven't updated the tests yet (I've only got it working with the compatibility tests so far) - https://github.com/chrisccoulson/secboot/tree/refactor-tpm2-policy

PR #180, opened 2021-10-06
tpm2: refactor policy.go

chrisccoulson:
The functions in policy.go are used across all metadata versions, and
some are quite complicated with branches to handle the different
versions and some of them also take a lot of arguments because they
have global scope - this is particularly true of
computeSealedKeyDynamicAuthPolicy (this is in seal.go, but handles
policy stuff), and is a hangover from when we didn't use SealedKeyObject
for updating the PCR policy.

Also in policy.go are some representations of policy data that are
shared between metadata versions, and some version-specific on-disk
representations (similar to the keyData / keyDataRaw_v* split).

This refactors the code in a similar way to the keyData refactor,
such that:
- There is a new interface representing a policy (keyDataPolicy), which
is obtained via the keyData interface.
- The new interface mostly provides a mechanism for updating the PCR
policy and executing the policy.
- The pcrPolicyCounterHandle interface is replaced by
pcrPolicyCounterContext, and can be obtained via the keyDataPolicy
interface.
- Each metadata version has an implementation of keyDataPolicy,
representing the static and PCR policy data.
- The tree structure used for executing PolicyOR assertions has
been split into 2 - there is the existing "flattened" tree
on-disk representation, but this is now converted to an actual
tree.

This does cause a bit of extra code duplication - I've tried to
keep this to a minimum, but I think it makes things more readable
overall.


tpm2/policy.go
```diff
@@ -20,420 +20,255 @@
 package tpm2
```
pedronis, line 20: the diff of this file is too messy to follow, there's too much renaming/removing and moving going on at the same time, maybe it's hard to avoid but  my main suggestion would be to try to avoid some of  the moving and do a separate PR that fixes the order of things, examples, we probably want the new interfaces at the top and then

*  pcrPolicyParams close to where dynamicPolicyComputeParams was
* policyOrNode close to where policyOrDataNode was
* policyOrTree close to where policyOrDataTree was
* the errors closer to where the previous error def was
* ensureSufficientORDigests where it was before
* pcrPolicyCounterContext  maybe where pcrPolicyCounterHandle was
* (t *policyOrTree) executeAssertions close to where executePolicyORAssertions was ?
* newPolicyOrTree close to where computeDynamicPolicy was ?

etc, maybe it's impossible let me know?

pedronis (commented):
it's good that the tests outside of policy*_test.go are mostly unchanged, this is still doing too much, though I understand that the art of splitting this kind of large refactoring is not simple but concretely this quite is hard to review especially the policy.go changes, I made some suggestions about reordering things in that file, maybe it can help, maybe it doesn't

tpm2/policy.go
chrisccoulson: Thanks, I've moved things around a bit. It has reduced the diff slightly (+5,031 −3,291 -> +5,009 −3,269), although I'm not sure if it's made it much more readable.

pedronis (approved):
lgtm, this is probably known but v0 code is less tested than v1/2.  There are some paths in keydata.go about it that aren't touched by the tests

PR #181, opened 2021-10-08
Add AuthRequestor interface

chrisccoulson:
This adds an AuthRequestor interface that is used by the Activate*
APIs to request a credential (passphrase or recovery key) via some
agent.

It also adds an implementation of AuthRequestor that uses
systemd-ask-password. The new API provides a simple template based
way to define the message that is displayed.

### DIFF
--- a/auth_requestor_systemd.go
+++ b/auth_requestor_systemd.go
@@ -1,0 +1,122 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2021 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


auth_requestor_systemd.go
```diff
@@ -0,0 +1,122 @@
...
+	cmd := exec.Command(
+		"systemd-ask-password",
+		"--icon", "drive-harddisk",
+		"--id", filepath.Base(os.Args[0])+":"+sourceDevicePath,
+		msg)
+	out := new(bytes.Buffer)
+	cmd.Stdout = out
+	cmd.Stdin = os.Stdin
+	if err := cmd.Run(); err != nil {
+		return "", err
```
pedronis, line 56: this error path is not tested

pedronis (commented):
looks good overall, small question about testing one main error path

auth_requestor_systemd.go
chrisccoulson: Thanks, I've added a test for this now.

pedronis (approved):
+1

PR #184, opened 2021-11-10
luks2: add DetectCryptsetupFeatures

chrisccoulson:
We depend on some cryptsetup features that aren't in all of the
cryptsetup versions we have in Ubuntu, so add a simple internal
feature detection API for those. This is motivated by the fact
that we're going to depend on --token-replace, which has just been
approved to land upstream and which we'll need to backport.

### DIFF
--- a/internal/luks2/cryptsetup.go
+++ b/internal/luks2/cryptsetup.go
@@ -70,7 +70,8 @@ const (
 	FeatureTokenImport
 
 	// FeatureTokenReplace indicates that tokens can be atomically replaced with
-	// ImportToken (yet to be implemented).
+	// ImportToken (yet to be implemented). This was introduced to cryptsetup by:
+	// https://gitlab.com/cryptsetup/cryptsetup/-/commit/98cd52c8d7bddf5b4c1ff775158a48bbb522acb2
 	FeatureTokenReplace
 )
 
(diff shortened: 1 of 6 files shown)


internal/luks2/cryptsetup.go
```diff
@@ -76,6 +107,34 @@ func cryptsetupCmd(stdin io.Reader, callback func(cmd *exec.Cmd) error, args ...
...
+			if n == 3 {
+				if major >= 3 || (major == 2 && minor >= 1) {
+					features |= FeatureHeaderSizeSetting
+				}
+				if major >= 3 || (major == 2 && minor >= 1) || (major == 2 && minor == 0 && patch >= 3) {
+					features |= FeatureTokenImport
+				}
+			}
+		}
+		if err := cryptsetupCmd(nil, nil, "--test-args", "token", "import", "--token-id", "0",
```
pedronis, line 130: is --test-args an undocumented (I don't see it in the man page) but supported option?

pedronis (commented):
a question

internal/luks2/cryptsetup.go
chrisccoulson: It is undocumented. It's currently used by cryptsetup's unit tests for validating combinations of commandline arguments.

chrisccoulson: It is documented in the `--help` output though:
```
      --test-args                       Do not run action, just validate all
                                        command line parameters
```

pedronis (approved):
+1

PR #187, opened 2021-12-09
Salt the digest used for verifying the Snap model HMAC key

chrisccoulson:
KeyData contains a list of HMACs of Snap models that are authorized to
access the data on the device unlocked by it, with the HMAC key being
derived from a hardware protected key that is recovered during early
boot.

In order to verify that the correct HMAC key is supplied when updating
the list of authorized Snap models, we store a digest of the HMAC key.
Update this to store a salted digest of the HMAC key instead, preserving
compatibility with existing KeyData objects.


chrisccoulson:
@setharnold Fancy taking a look at this too?

setharnold:
It looks fine to me, but I think it'd be nice if there were a few more test cases, with eg `crc32` and `adler32` or `sha1` and `md5` inputs, salts that are just `""`, or a hundred bytes of `A`, or a bunch of `====` etc.

Is an early failure due to a corrupted or incorrect key the best answer? I've often seen cryptographic tools instead perform computations all the way through with whatever key they were supplied, only to discard garbage at the very end of the process (before returning it to the 'user', whatever the user may look like), so success/failure of oracles is harder to determine.

Keying the hash will prevent one kind of brute force operation but may not help with other kinds of brute force operations. Is the "better" approach removing the hash entirely?

Thanks

pedronis (approved):
code itself lgtm

PR #188, opened 2022-01-17
Add passphrase support to KeyData

chrisccoulson:
This adds passphrase support to KeyData, so that it can be configured to
require a passphrase in order to recover the key. This works by further
encrypting the payload that is protected by the platform's secure device
(such as a TPM) using a passhrase derived key which is derived using the
Argon2i KDF. The encryption is performed with AES-256-CFB.

Platform integration is supported so that passphrase support can take
advantage of features provided by the platform's secure device (such as
dictionary attack protection on TPM devices). The platform implementation
is expected to require the secure device in order to verify whether a
passphrase is correct, both when recovering keys (in order to return an
appropriate error) and when changing the passphrase. This is to prevent
offline attacks against the passphrase. For the TPM platform, this will
be achieved by setting the authorization value of the sealed key object
to a key derived from the passphrase derived key, ensuring that a
compromise of the TPM doesn't reveal the passphrase or passphrase derived
key. The TPM will require knowledge of the authorization value for
unsealing and changing the passphrase.

KeyData makes a distinction between having passphrase support enabled
and having an empty passphrase. There are 3 new APIs for configuring a
passphrase:
- KeyData.SetPassphrase is used to enable passphrase support with the
supplied passphrase if it isn't enabled.
- KeyData.ChangePassphrase is used to update the passphrase.
- KeyData.ClearPassphraseWithPassphrase is used to disable passphrase
support. It is named "ClearPassphraseWithPassphrase" because support for
additional authentication mechanisms in the future could facilitate
clearing the passphrase via another mechanism or just by deleting the
"passphrase_protected_payload" field of KeyData if another authentication
mechanism is enabled (KeyData.ClearPassphrase).

KeyData also provides a new API for recovering keys using a passphrase -
KeyData.RecoverKeysWithPassphrase.


argon2.go
```diff
@@ -0,0 +1,161 @@
...
+		return params, nil
+	}
+}
+
+// CostParams defines the cost parameters for key derivation using Argon2.
+type CostParams = argon2.CostParams
+
+// KDF is an interface to abstract use of the Argon2 KDF to make it possible
+// to delegate execution to a short-lived utility process where required.
+type KDF interface {
```
pedronis, line 133: shouldn't this be called Argon2KDF from the description? otherwise the fact that CostParams is actually argon2.CostParams is odd anyway

argon2.go
```diff
@@ -0,0 +1,161 @@
...
+
+var argon2iKDF = argon2iKDFImpl{}
+
+// Argon2iKDF returns the in-process Argon2i implementation of KDF. This
+// shouldn't be used in long-lived system processes - these processes should
+// instead provide their own KDF implementation which delegates to a short-lived
+// utility process which will use the in-process implementation.
+func Argon2iKDF() KDF {
+	return argon2iKDF
+}
```
pedronis, line 161: this are trivial but nevertheless not tested

crypt.go
```diff
@@ -193,20 +207,64 @@ func (s *activateWithKeyDataState) run() (success bool) {
...
 		}
 
-		return true
+		return true, nil
 	}
 
-	// TODO: Passphrase support
+	// Try keys that require a passphrase
+	tries := s.passphraseTries
+	var lastErr error
```
pedronis, line 215: wouldn't passphraseError be a  clearer name for this?

crypt.go
```diff
@@ -193,20 +207,64 @@ func (s *activateWithKeyDataState) run() (success bool) {
...
+		// ubuntu-data).
+		passphrase, err := s.authRequestor.RequestPassphrase(s.volumeName, s.sourceDevicePath)
+		if err != nil {
+			lastErr = xerrors.Errorf("cannot obtain passphrase: %w", err)
+			continue
+		}
+
+		for _, k := range s.keys {
+			if k.AuthMode()&AuthModePassphrase == 0 {
+				continue
```
pedronis, line 236: tests don't seem to reach here, so there's not test for mixing different modes?

crypt.go
```diff
@@ -193,20 +207,64 @@ func (s *activateWithKeyDataState) run() (success bool) {
...
-		return true
+		return true, nil
 	}
 
-	// TODO: Passphrase support
+	// Try keys that require a passphrase
+	tries := s.passphraseTries
+	var lastErr error
+
+	for tries > 0 && numPassphraseKeys > 0 {
```
pedronis, line 217: should the ActivateVolumeOptions.PassphraseTries doc be clear that tries are across keys?

pedronis (commented):
I did a first pass on argon2.go and crypt.go changes. Will look at the rest later.

keydata.go
```diff
@@ -302,6 +313,99 @@ func (d *KeyData) snapModelAuthKey(auxKey AuxiliaryKey) ([]byte, error) {
...
+	if handler == nil {
+		return ErrNoPlatformHandlerRegistered
+	}
+
+	if kdfOptions == nil {
+		var defaultOptions KDFOptions
+		kdfOptions = &defaultOptions
+	}
+
+	keyLen := passphraseDerivedKeyLen + aes.BlockSize
```
pedronis, line 327: this probably merits a comment about why this is chosen this way

platform.go
```diff
@@ -73,22 +81,21 @@ type PlatformKeyDataHandler interface {
...
+	RecoverKeysWithAuthKey(data *PlatformKeyData, key []byte) (KeyPayload, error)
 
-	// ChangeAuthValue is called to notify the platform implementation that the
-	// passphrase is being changed. The oldAuthValue and newAuthValue parameters
-	// are passphrase derived keys. On success, it should return an updated
-	// PlatformKeyData.
-	// ChangeAuthValue(data *PlatformKeyData, oldAuthValue, newAuthValue []byte) (*PlatformKeyData, error)
+	// ChangeAuthKey is called to notify the platform implementation that the
+	// passphrase is being changed. The old and new parameters are passphrase derived
+	// keys. On success, it should return an updated handle.
```
pedronis, line 97: this should mention that old can be nil if this is the first time the auth key is set?

pedronis (commented):
finished a first pass, haven't looked much at tests

platform.go
chrisccoulson: I've updated this now

keydata.go
chrisccoulson: I've added a comment to this now.

mockable.go
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
```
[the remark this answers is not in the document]
chrisccoulson: I've reverted this

crypt.go
chrisccoulson: I've updated the documentation now

crypt.go
chrisccoulson: Yeah, that sounds ok

argon2.go
chrisccoulson: We already have KDFOptions, so I'm not sure that renaming this to Argon2KDF without renaming KDFOptions makes sense.

What I've done instead is rename CostParams to KDFCostParams (and it's no longer type-aliased either), so we have KDF, KDFOptions and KDFCostParams.

argon2.go
chrisccoulson: I've added test cases for this now

crypt.go
chrisccoulson: I've added a test for this.

pedronis (approved):
thanks

PR #189, opened 2022-02-25
move Snap model authorization to ActivateVolumeWith*KeyData

chrisccoulson:
The ActivateVolumeWith*KeyData functions currently return a
SnapModelChecker to delegate authorization of the Snap model to
snap-bootstrap. This means that authorization of the model is
performed after activation of the container. There are a couple
of issues with this though:
- If multiple KeyData are supplied and the container is activated
with one of these but the model isn't authorized, it requires another
call to ActivateVolumeWithMultipleKeyData in order to try others.
- If the model isn't auhtorized with any KeyData, the container has
to be deactivated again.

The first issue is difficult to solve with the current architecture
when KeyData are loaded from the LUKS metadata rather than being
supplied externally from a file.

This implements a different approach where SnapModelChecker is removed,
and the model authorization is done in ActivateVolumeWith*KeyData before
the container is activated, with a KeyData being rejected before
activation if the model authorization check fails.

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -342,7 +342,7 @@ type ActivateVolumeOptions struct {
 	// kernel keys created during activation.
 	KeyringPrefix string
 
-	// SnapModel is the snap device model that will access the data
+	// Model is the snap device model that will access the data
 	// on the encrypted container. The ActivateVolumeWith* functions
 	// will check that this model is authorized via the KeyData
 	// binding before unlocking the encrypted container.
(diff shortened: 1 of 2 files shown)


crypt.go
```diff
@@ -352,6 +341,19 @@ type ActivateVolumeOptions struct {
...
+	// on the encrypted container. The ActivateVolumeWith* functions
+	// will check that this model is authorized via the KeyData
+	// binding before unlocking the encrypted container.
+	//
+	// The caller of the ActivateVolumeWith* API is responsible for
+	// validating the associated model assertion and snaps.
+	//
+	// Set this to SkipSnapModelCheck to skip the check.
+	//
+	// It is ignored by ActivateVolumeWithRecoveryKey.
```
pedronis, line 355: ... but otherwise it cannot be left set to nil

crypt.go
```diff
@@ -352,6 +341,19 @@ type ActivateVolumeOptions struct {
...
+	SnapModel SnapModel
```
pedronis, line 356: nitpick: I think the field itself could just be called Model

pedronis (approved):
lgtm, couple small comments

PR #191, opened 2022-02-25
Add keydata platform handle accessors

chrisccoulson:
This is a prerequisite for creating a tpm2 platform, as we need a way
to create and update a SealedKeyObject from a KeyData object.

There are 2 new functions:
- KeyData.UnmarshalPlatformHandle() - unmarshals the JSON platform
handle payload to the supplied handle.
- KeyData.MarshalAndUpdatePlatformHandle() - marshals the handle to
JSON and updates the in-memory KeyData object.

The intention is for the code in tpm2/ to provide helpers that use
these, rather than these functions being used outside of secboot.

For creating a SealedKeyObject from a KeyData object, there will be
a new API that looks like this:

 func NewSealedKeyObjectFromKeyData(data *secboot.KeyData) (*SealedKeyObject, error)

This will pass a SealedKeyObject (which will implement
json.Unmarshaler) to KeyData.UnmarshalPlatformHandle.

SealedKeyObject will implement json.Marshaler, so you can update the
PCR policy like this:

 sko, _ := NewSealedKeyObjectFromKeyData(keyData)
 sko.UpdatePCRProtectionPolicy(tpm, authKey, profile)
 keyData.WriteAtomic(w)

In this case, functions that mutate the SealedKeyObject will call
KeyData.MarshalAndUpdatePlatformHandle.

The representation of the platform handle in this API isn't currently
consistent with the one in KeyCreationData. I've got a follow-up
for this.


keydata.go
```diff
@@ -446,6 +446,28 @@ func (d *KeyData) AuthMode() (out AuthMode) {
 	return out
 }
 
+// UnmarshalPlatformHandle unmarshals the JSON platform handle payload to the
```
pedronis, line 449: s/to/into/

this should also clarify that handle needs to be a non-nil pointer (that's a requirement from Unmarshal itself but it's better to be explicit).

pedronis (approved):
this looks reasonable though I might have clearer opinions once seen in use concretely

PR #193, opened 2022-03-22
Always run cheap unmocked cryptsetup tests

chrisccoulson:
This splits cryptsetupSuite in internal/luks2 into cryptsetupSuite
and cryptsetupSuiteExpensive, with the latter being for tests that
run cryptsetup's KDF benchmarking. This allows most of the tests to
run whilst still being able to skip the subset of tests that take a
long time when running them locally.

This also splits cryptSuiteExpensive in crypt_test.go into
cryptSuiteUnmocked and cryptSuiteUnmockedExpensive, with the latter
being for tests that run cryptsetup's KDF benchmarking. All other
unmocked tests will run as part of the cryptSuiteUnmocked suite. Right
now, all of the tests run as part of cryptSuiteUnmockedExpensive,
but this is just preparation for follow up PRs that are going to add
more tests here.

### DIFF
--- a/crypt_test.go
+++ b/crypt_test.go
@@ -2126,29 +2126,35 @@ func (s *cryptSuite) TestChangeLUKS2KeyUsingRecoveryKey4(c *C) {
 	})
 }
 
-type cryptSuiteExpensive struct {
+type cryptSuiteUnmockedBase struct {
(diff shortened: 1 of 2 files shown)


pedronis (approved):
lgtm

PR #194, opened 2022-03-22
luks2: Move format argument validation to luks2 package

chrisccoulson:
luks2.Format already returns an error if a custom metadata area
or keyslot area size is specified but cryptsetup doesn't support
these options, but secboot.InitializeLUKS2Container performs its
own checking of these parameters. This just moves them into the
same place.


pedronis (approved):
looks reasonable

chrisccoulson:
luks2.Format already returns an error if a custom metadata area
or keyslot area size is specified but cryptsetup doesn't support
these options, but secboot.InitializeLUKS2Container performs its
own checking of these parameters. This just moves them into the
same place.

PR #195, opened 2022-03-22
update LUKS2 KDF defaults to match snapd

chrisccoulson:
Snapd is already tuning the KDF cost parameters to their minimum,
so just use these defaults in secboot.

### DIFF
--- a/export_test.go
+++ b/export_test.go
@@ -19,6 +19,48 @@
 
 package secboot
 
+import (
+	"fmt"
+	"strconv"
+	"time"
+)
+
+func (o *InitializeLUKS2ContainerOptions) CryptsetupArguments() []string {
+	if o == nil {
+		o = &InitializeLUKS2ContainerOptions{}
+	}
+	kdfOptions := o.KDFOptions
+	if kdfOptions == nil {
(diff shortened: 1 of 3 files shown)


export_test.go
```diff
@@ -19,6 +19,48 @@
 
 package secboot
 
+import (
+	"fmt"
+	"strconv"
+	"time"
+)
+
+func (o *InitializeLUKS2ContainerOptions) CryptsetupArguments() []string {
```
pedronis, line 28: this matches corresponding code in luks2/internal, its complexity its a bit odd given is used in tests, the previous approach of spelling out things by hand seemed ok to me, so I'm probably missing something. Anyway if kept ExpectedCryptsetupArguments seems a bit of a clearer name for it

pedronis (commented):
lgtm, but not sure I fully grasp the test changes

export_test.go
chrisccoulson: Ah, I can't remember why I moved this, particularly because the next PR actually deletes this bit of code

pedronis (approved):
if the new helper goes away soon this is fine, otherwise the name with Expected in it would be clearer

PR #196, opened 2022-03-22
crypt: refactor mocked tests

chrisccoulson:
Mocked tests in crypt_test.go use the internal/luks2 package but mock
the cryptsetup binary. This is a bit strange because the use of cryptsetup
is an implementation detail that's internal to internal/luks2, and is left
over from before that package was created.

Refactor the mocked tests to mock the use of internal/luks2 instead, and
move cryptsetup commandline checks to the internal/luks2 package. This
is going to make it easier to implement tests in future PRs which introduce
APIs that work with tokens.


crypt_test.go
```diff
@@ -120,162 +118,208 @@ func (r *mockAuthRequestor) RequestRecoveryKey(volumeName, sourceDevicePath stri
...
+	return slot
+}
 
-	mockLUKS2ActivateCalls []struct {
-		volumeName       string
-		sourceDevicePath string
+// mockLUKS2 mocks a device's global LUKS2 state. It provides mock
+// implementations of the various LUKS2 operations.
+type mockLUKS2 struct {
+	operations []string                       // A log of LUKS2 operations recorded during a test
```
pedronis, line 144: what we do sometimes is having a generic op structure to cover possible arg combinations and then have a slice of those:

```
ops []op

type op struct {
  op string
  paths []string
  slot int
  addKeyOptions *luks2.AddKeyOptions
 // etc ...
}
```

it doesn't matter too much, the main issue is that different structs might have the same Sprint output

chrisccoulson: Thanks - I'll take a look at this at the end of this series of PRs

PR #197, opened 2022-03-22
crypt: allow keyslots to be named

chrisccoulson:
This is draft for now because it depends on https://github.com/snapcore/secboot/pull/196

This updates the APIs in crypt.go that add / modify keys to create
named keyslots - a keyslot with an associated token with the type
"ubuntu-fde" or "ubuntu-fde-recovery" - each of these tokens has
a unique name.

- IntializeLUKS2Container now creates a named keyslot with the
type "ubuntu-fde" and the name "default" unless another name is
supplied.
- AddRecoveryKeyToLUKS2Container has been renamed to
SetLUKS2ContainerRecoveryKey and creates a named recovery keyslot
with the type "ubuntu-fde-recovery" and the name "default-recovery"
unless another name is supplied.
- ChangeLUKS2ContainerUsingRecoveryKey has been renamed to
SetLUKS2ContainerUnlockKey and creates a named keyslot with the type
"ubuntu-fde" and the name "default" unless another name is supplied.
- DeleteLUKS2ContainerKey deletes a named keyslot.
- ListLUKS2ContainerUnlockKeyNames and ListLUKS2ContainerRecoveryKeyNames
can be used to list the names of existing named keyslots.

A subsequent PR will provide an API for updating the "ubuntu-fde"
tokens in order to attach and update the associated KeyData.

The SetLUKS2Container*Key functions are not atomic and they cannot
be used to change the key associated with an existing keyslot. If
the name of an existing keyslot is supplied then it will be deleted
first (as long as it is not the last keyslot) and then recreated.
The ability to change keys isn't particularly useful anyway, because
although cryptsetup can change a key atomically, changing a
"ubuntu-fde" keyslot always requires 2 operations (change the keyslot
and then import the updated KeyData).


crypt.go
```diff
@@ -540,83 +560,189 @@ func (o *InitializeLUKS2ContainerOptions) formatOpts() *luks2.FormatOptions {
...
+
+	if err := luks2AddKey(devicePath, existingKey, newKey, &luks2.AddKeyOptions{KDFOptions: options.luksOpts(), Slot: freeSlot}); err != nil {
+		return xerrors.Errorf("cannot add key: %w", err)
+	}
+
+	// XXX: If we fail between AddKey and ImportToken, then we end up with a
+	//  used keyslot that cannot be identified and no way to roll back the
+	//  interrupted transaction safely. Ideally we'd be able to add a key and
+	//  token in a single atomic operation, but this isn't even something that
+	//  is possible with the libcryptsetup API.
```
pedronis, line 661: yes, this is a legitimate concern

pedronis (commented):
did a first pass. What's the backward compatibility/migration strategy? the old API is not available anymore after this change. snapd has switched to use directly the luks2 package so is not affected as such I think. Do we need/will there be an API to add a name to a preexisiting keyslot? or should it be done using luks2 and luksview directly? that would violate encapsulation of details here a bit

chrisccoulson:
> did a first pass. What's the backward compatibility/migration strategy? the old API is not available anymore after this change. snapd has switched to use directly the luks2 package so is not affected as such I think. Do we need/will there be an API to add a name to a preexisiting keyslot? or should it be done using luks2 and luksview directly? that would violate encapsulation of details here a bit

I hadn't really thought about this too much - I guess I could add an additional API to add a name to an existing keyslot, although it might be better in another PR.

crypt.go
```diff
@@ -540,83 +560,189 @@ func (o *InitializeLUKS2ContainerOptions) formatOpts() *luks2.FormatOptions {
...
-	defaultKdfOptions := &KDFOptions{MemoryKiB: 32, ForceIterations: 4}
 	if options == nil {
-		options = &InitializeLUKS2ContainerOptions{KDFOptions: defaultKdfOptions}
-	} else if options.KDFOptions == nil {
-		options.KDFOptions = defaultKdfOptions
+		var defaultOptions InitializeLUKS2ContainerOptions
+		options = &defaultOptions
+	}
+	if options.KDFOptions == nil {
+		options.KDFOptions = &KDFOptions{MemoryKiB: 32, ForceIterations: 4}
```
pedronis, line 601: it might be unexpected to modify in place the passed in options

pedronis (approved):
thanks, we do need to solve the concern in future PRs and also the issue of whether we need an API to deal with preexisting slots, also a comment about in-place modification of arguments

PR #198, opened 2022-03-22
Add KeyDataReader and KeyDataWriter implementations for LUKS

chrisccoulson:
This adds support for reading and writing KeyData objects to the LUKS
header. A follow up PR will add support for making use of these during
activation.

### DIFF
--- a/keydata_luks.go
+++ b/keydata_luks.go
@@ -1,0 +1,136 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2021 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package secboot
+
+import (
+	"bytes"
+	"errors"
+
+	"golang.org/x/xerrors"
+
+	"github.com/snapcore/secboot/internal/luks2"
+	"github.com/snapcore/secboot/internal/luksview"
+)
+
+// LUKS2KeyDataReader provides a mechanism to read a KeyData from a LUKS2 token.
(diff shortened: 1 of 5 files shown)


keydata_luks.go
```diff
@@ -0,0 +1,136 @@
...
+// Priority indicates the priority of the keyslot associated with the token from
+// which this KeyData is read. The default priority is 0 with higher numbers
+// indicating a higher priority.
+func (r *LUKS2KeyDataReader) Priority() int {
+	return r.priority
+}
+
+// NewLUKS2KeyDataReader is used to read a LUKS2 token containing key data with
+// the specified name on the specified LUKS2 container.
+func NewLUKS2KeyDataReader(devicePath, name string) (*LUKS2KeyDataReader, error) {
```
pedronis, line 59: I never remarked this but golang style usually put constructs/New functions first after the struct definition before the methods. secboot doesn't seem to follow this convention. Might be easier to do a separate PR to address this at some point

keydata_luks.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// NewLUKS2KeyDataWriter creates a new LUKS2KeyDataWriter for atomically writing a
+// KeyData to a LUKS2 token with the specicied name and priority on the specified
+// LUKS2 container.
+//
+// The container must already contain a token of the correct type with the supplied
+// name. The initial token is bootstrapped by InitializeLUKS2Container or
+// SetLUKS2ContainerUnlockKey.
+func NewLUKS2KeyDataWriter(devicePath, name string, priority int) (*LUKS2KeyDataWriter, error) {
```
pedronis, line 114: shouldn't there be a way to reuse the priority of the preexisting token/slot ? without further context is a bit unclear why we would re-specify this all the times

keydata_luks.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	token, _, exists := view.TokenByName(name)
+	if !exists {
+		return nil, errors.New("a keyslot with the specified name does not exist")
+	}
+
+	kdToken, ok := token.(*luksview.KeyDataToken)
+	if !ok {
+		return nil, errors.New("named keyslot is the wrong type")
```
pedronis, line 72: s/is/has/ ? as below

pedronis (commented):
did a pass, some comments/questions

keydata_luks.go
chrisccoulson: Thanks, I've reordered this one.

keydata_luks.go
chrisccoulson: I've removed the priority argument from this function and added a SetPriority function instead.

pedronis (approved):
thank you

PR #201, opened 2022-03-22
Allow the platform handle to be any arbitrary JSON value in KeyCreationData

chrisccoulson:
This aligns the representation of the platform handle in KeyCreationData
with the representation introduced to the KeyData API in
https://github.com/snapcore/secboot/pull/191.

Note that the representation of the platform handle in PlatformKeyDataHandler
remains unchanged (it is the encoded JSON). This is becuase
PlatformKeyDataHandler is used to call from KeyData to an implementation,
and it is not possible to decode the JSON in this scenario because the
data is opaque.

### DIFF
--- a/keydata.go
+++ b/keydata.go
@@ -148,8 +148,16 @@ const (
 // KeyCreationData is the data required to create a new KeyData object.
 // It should be produced by a platform implementation.
 type KeyCreationData struct {
-	PlatformKeyData
-	PlatformName string // Name of the platform that produced this data
+	// Handle contains metadata required by the platform in order to recover
+	// this key. It is opaque to this go package. It should be a value that can
+	// be encoded to JSON using go's encoding/json package, which could be
+	// something as simple as binary data stored in a byte slice or a more complex
+	// JSON object, depending on the requirements of the implementation. A handle
+	// already encoded to JSON can be supplied using the json.RawMessage type.
+	Handle interface{}
+
+	EncryptedPayload []byte // The encrypted payload
+	PlatformName     string // Name of the platform that produced this data
 
 	// AuxiliaryKey is a key used to authorize changes to the key data.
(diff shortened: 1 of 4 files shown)


platform.go
```diff
@@ -62,15 +62,9 @@ func (e *PlatformHandlerError) Unwrap() error {
...
+// platform implementations via the PlatformKeyDataHandler.
 type PlatformKeyData struct {
-	// Handle contains metadata required by the platform in order to recover
-	// this key. It is opaque to this go package. It should be an encoded JSON
-	// value, which could be something as simple as a single string containing
-	// a base64 encoded binary payload or a more complex JSON object, depending
-	// on the requirements of the implementation.
-	Handle []byte
-
+	Handle           []byte // The JSON encoded platform handle
```
pedronis, line 67: it is a bit confusing that there are two important fields with the same name but different types. In theory this one should be called MarshaledHandle

keydata.go
```diff
@@ -720,8 +728,9 @@ func ReadKeyData(r KeyDataReader) (*KeyData, error) {
 // the platform's secure device and the associated handle required for subsequent
 // recovery of the keys.
 func NewKeyData(creationData *KeyCreationData) (*KeyData, error) {
-	if !json.Valid(creationData.Handle) {
-		return nil, errors.New("handle is not valid JSON")
+	handle, err := json.Marshal(creationData.Handle)
```
pedronis, line 731: same for this local var

pedronis (commented):
looks alright, naming wondering

platform.go
chrisccoulson: How about `EncodedHandle` instead?

pedronis: that's fine too

chrisccoulson: Thanks, I've updated this now.

pedronis (approved):
thank you

PR #202, opened 2022-03-29
tpm2: Add initial PlatformKeyDataHandler implementation

chrisccoulson:
This adds an initial TPM2 implementation of PlatformKeyDataHandler, plus
new APIs for sealing keys to make use of secboot.KeyData.

The new sealing APIs are ProtectKeyWithTPM, ProtectKeysWithTPM and
ProtectKeyWithExternalStorageKey, to replace the SealKeyToTPM,
SealKeyToTPMMultiple and SealKeyToExternalTPMStorageKey. They are
implemented in a way to facilitate adding another API in the future to
protect additional keys using the same policy as an existing key.

Also added is a new API to update the PCR policy for multiple keys in
one go - UpdateKeyDataPCRProtectionPolicy, which replaces
UpdateKeyPCRProtectionPolicyMultiple.

Obtaining a SealedKeyObject data structure associated with a key is done
differently for keys created with the new APIs. There is a new API to
decode a SealedKeyObject from a secboot.KeyData object -
NewSealedKeyObjectFromKeyData.

### DIFF
--- a/tpm2/keydata.go
+++ b/tpm2/keydata.go
@@ -21,14 +21,17 @@ package tpm2
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 
 	"github.com/canonical/go-tpm2"
+	"github.com/canonical/go-tpm2/mu"
 
 	"golang.org/x/xerrors"
 
+	"github.com/snapcore/secboot"
 	"github.com/snapcore/secboot/internal/tcg"
 )
 
@@ -49,7 +52,7 @@ func isKeyDataError(err error) bool {
 	return xerrors.As(err, &e)
 }
 
-// keyData represents the actual data for a SealedKeyObject.
+// keyData represents the actual data for a SealedKeyData or legacy SealedKeyObject.
 type keyData interface {
 	// Version is the metadata version. Note that the keyData
 	// implementation is not responsible for serializing this.
@@ -200,4 +203,67 @@ func (k *sealedKeyDataBase) validateData(tpm *tpm2.TPMContext, session tpm2.Sess
 	}
 
 	return pcrPolicyCounterPub, nil
+}
+
+// SealedKeyData corresponds to a sealed key data object created by
+// [ProtectKeyWithTPM], [ProtectKeysWithTPM], or [ProtectKeyWithExternalStorageKey].
+type SealedKeyData struct {
+	sealedKeyDataBase
+	k *secboot.KeyData
+}
+
+// NewSealedKeyData returns a SealedKeyData from the supplied secboot.KeyData
+// object.
+func NewSealedKeyData(k *secboot.KeyData) (*SealedKeyData, error) {
+	var skd *SealedKeyData
+	if err := k.UnmarshalPlatformHandle(&skd); err != nil {
+		return nil, InvalidKeyDataError{err.Error()}
+	}
+	skd.k = k
+
+	return skd, nil
+}
+
+// Version returns the version number that this sealed key object was created with.
+func (k *SealedKeyData) Version() uint32 {
+	return k.data.Version()
+}
+
+// PCRPolicyCounterHandle indicates the handle of the NV counter used for PCR policy revocation for this sealed key object (and for
+// PIN integration for version 0 key files).
+func (k *SealedKeyData) PCRPolicyCounterHandle() tpm2.Handle {
+	return k.data.Policy().PCRPolicyCounterHandle()
+}
+
+func (k *SealedKeyData) MarshalJSON() ([]byte, error) {
+	w := new(bytes.Buffer)
+	if _, err := mu.MarshalToWriter(w, k.data.Version()); err != nil {
+		return nil, err
+	}
+	if err := k.data.Write(w); err != nil {
+		return nil, err
+	}
+	return json.Marshal(w.Bytes())
+}
+
+func (k *SealedKeyData) UnmarshalJSON(data []byte) error {
+	var b []byte
+	if err := json.Unmarshal(data, &b); err != nil {
+		return err
+	}
+
+	r := bytes.NewReader(b)
+
+	var version uint32
+	if _, err := mu.UnmarshalFromReader(r, &version); err != nil {
+		return err
+	}
+
+	kd, err := readKeyData(r, version)
+	if err != nil {
+		return err
+	}
+
+	k.data = kd
+	return nil
 }
--- a/tpm2/seal.go
+++ b/tpm2/seal.go
@@ -1,0 +1,445 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2022 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package tpm2
+
+import (
+	"crypto"
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
+	"errors"
+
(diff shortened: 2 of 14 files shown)


tpm2/keydata.go
```diff
@@ -238,31 +248,74 @@ func (k *SealedKeyObject) PCRPolicyCounterHandle() tpm2.Handle {
 
 // WriteAtomic will serialize this SealedKeyObject to the supplied writer.
```
pedronis, line 249: should the doc comment mention something about this not being usable with new objects? is this going away at some point?

tpm2/update.go
```diff
@@ -201,12 +203,21 @@ func updateKeyPCRProtectionPolicyCommon(tpm *tpm2.TPMContext, keys []*SealedKeyO
...
 		k.data.Policy().SetPCRPolicyFrom(primaryKey.data.Policy())
 	}
 
+	for _, k := range keys {
+		if k.k == nil {
+			continue
+		}
+		if err := k.k.MarshalAndUpdatePlatformHandle(k); err != nil {
+			return xerrors.Errorf("cannot update TPM platform handle on KeyData: %w", err)
+		}
```
pedronis, line 218: because of the k.k accesses it feels like this bit of code should be some sort of helper method on SealedKeyObject

tpm2/seal.go
```diff
@@ -0,0 +1,445 @@
...
+
+var (
+	secbootNewKeyData = secboot.NewKeyData
+)
+
+// ProtectKeyParams provides arguments for the ProtectKey* APIs.
+type ProtectKeyParams struct {
+	// PCRProfile defines the profile used to generate the initial PCR protection
+	// policy for the newly created sealed key data. This can be updated later on
+	// by calling SealedKeyData.UpdatePCRProtectionPolicy.
```
pedronis, line 44: is that SealedKeyObject ? I don't remember something called SealedKeyData

tpm2/seal.go
```diff
@@ -0,0 +1,445 @@
...
+
+type keyDataParams struct {
+	PCRPolicyCounterHandle tpm2.Handle
+	PCRProfile             *PCRProtectionProfile
+}
+
+// makeNewKeyData protects the supplied keys using the supplied keySealer and
+// parameters. If required, a PCR policy counter is created. The returned key
+// will have an initial PCR policy as specified via the supplied parameters.
+func makeNewKeyData(tpm *tpm2.TPMContext, key secboot.DiskUnlockKey, authKey secboot.AuxiliaryKey, params *keyDataParams,
```
pedronis, line 216: why there's a New in this method name? it doesn't carry a lot of clear information

tpm2/keydata.go
```diff
@@ -140,6 +141,8 @@ func newKeyData(keyPrivate tpm2.Private, keyPublic *tpm2.Public, importSymSeed t
 // SealedKeyObject corresponds to a sealed key data file.
 type SealedKeyObject struct {
 	data keyData
+
+	k *secboot.KeyData
```
pedronis, line 145: is there some way to avoid this? have two types? when reading all the code is quite confusing to have to keep track of when a SeledKeyDataObject has this set or not

tpm2/keydata.go
chrisccoulson: This is fixed by PR #250 now.

tpm2/update.go
chrisccoulson: This is fixed by PR #250 now.

tpm2/keydata.go
chrisccoulson: This is fixed by PR #250 now.

tpm2/keydata.go
```diff
@@ -201,3 +204,66 @@ func (k *sealedKeyDataBase) validateData(tpm *tpm2.TPMContext, session tpm2.Sess
 
 	return pcrPolicyCounterPub, nil
 }
+
+// SealedKeyData corresponds to a sealed key data object created by
+// [ProtectKeyWithTPM], [ProtectKeysWithTPM], or [ProtectKeyWithExternalStorageKey].
```
pedronis, line 209: nitpick: we weren't using \[\] (consistently) before?

tpm2/seal.go
```diff
@@ -0,0 +1,445 @@
...
+
+	if err == nil {
+		return
+	}
+
+	index, err := tpm2.CreateNVIndexResourceContextFromPublic(c.pub)
+	if err != nil {
+		return
+	}
+	c.tpm.NVUndefineSpace(c.tpm.OwnerHandleContext(), index, c.session)
```
pedronis, line 148: this path doesn't seem to be used/reached by tests

tpm2/seal.go
```diff
@@ -0,0 +1,445 @@
...
+}
+
+// makeKeyDataPolicy creates the policy data required to seal a key with makeKeyDataWithPolicy
+// and creates a PCR policy counter if required.
+func makeKeyDataPolicy(tpm *tpm2.TPMContext, pcrPolicyCounterHandle tpm2.Handle, authKey secboot.AuxiliaryKey,
```
pedronis, line 153: this should probably be closer to keyDataPolicyParams definition, the latter name maybe needs tweaks

tpm2/seal.go
```diff
@@ -0,0 +1,445 @@
...
+		return nil, nil, nil, xerrors.Errorf("cannot create initial policy data: %w", err)
+	}
+
+	return &keyDataPolicyParams{
+		Alg:    nameAlg,
+		Data:   policyData,
+		Digest: authPolicy}, pcrPolicyCounter, authKey, nil
+}
+
+type keyDataParams struct {
```
pedronis, line 208: it's not super clear from the name alone what distinguishes keyDataParams vs keyDataPolicyParams

tpm2/seal.go
```diff
@@ -0,0 +1,445 @@
...
+	// Create the initial policy data
+	policyData, authPolicy, err := newKeyDataPolicy(nameAlg, authPublicKey, pcrPolicyCounter.Pub(), pcrPolicyCount)
+	if err != nil {
+		return nil, nil, nil, xerrors.Errorf("cannot create initial policy data: %w", err)
+	}
+
+	return &keyDataPolicyParams{
+		Alg:    nameAlg,
+		Data:   policyData,
+		Digest: authPolicy}, pcrPolicyCounter, authKey, nil
```
pedronis, line 205: data and digest are not the most clear names here, especially because the digest is also a "policy"

pedronis (commented):
looking good, some comments, most of them are naming vs clarity wondering

tpm2/keydata.go
chrisccoulson: I've not used that consistently before because I wasn't aware of it - I've started using it in recent commits because it's nice to have the local doc links.

tpm2/seal.go
chrisccoulson: I'm not sure how to fix this other than adding a comment to both of them - I've tried thinking of other names for them, but I'm not sure if I can come up with anything better than this.

tpm2/seal.go
chrisccoulson: I can probably come up with better names for these fields.

tpm2/seal.go
```diff
@@ -0,0 +1,450 @@
...
+			session: session,
+			pub:     pub}
+
+		defer func() { pcrPolicyCounter.undefineOnError(err) }()
+	}
+
+	alg := tpm2.HashAlgorithmSHA256
+
+	// Create the initial policy data
+	policyData, authPolicy, err := newKeyDataPolicy(alg, authPublicKey, pcrPolicyCounter.Pub(), pcrPolicyCount)
```
pedronis, line 201: I notice that the  pre-existing doc comment for newKeyDataPolicy doesn't say much about authPolicy ?

tpm2/seal.go
```diff
@@ -0,0 +1,450 @@
...
+	if err != nil {
+		return
+	}
+	c.tpm.NVUndefineSpace(c.tpm.OwnerHandleContext(), index, c.session)
+}
+
+// keyDataPolicyParams corresponds to the parameters of a key's computed authorization
+// policy, consisting of the digest algorithm, the policy digest and the associated policy
+// data.
+type keyDataPolicyParams struct {
```
pedronis, line 148: maybe this should be called keyDataPolicyAuthParams? or keyDataAuthPolicy? or keyDataBaseAuthPolicy?

tpm2/seal.go
pedronis: this maybe should be called keyDataPCRParams? or keyDataPolicyPCRParams?

pedronis (approved):
thanks, some considerations still about names

PR #206, opened 2022-05-30
tpm2: require a simulator for createPcrPolicyCounter test

chrisccoulson:
policySuite.TestCreatePcrPolicyCounter uses
TPMTest.NextAvailableHandle() from go-tpm2's testutil package to
pick a free handle for a NV index, but this is inappropriate because
the test requires that the index is created at a specific handle,
and will fail if that handle is already in use and the test picks
a different one. Stop using it, and ensure the test only runs on a
simulator, where we can guarantee that the selected handle is free.


pedronis (approved):
looks alright to me

pedronis:
@chrisccoulson the change here looks fine to me but the tests are actually failing

chrisccoulson:
> @chrisccoulson the change here looks fine to me but the tests are actually failing

Ah, I'd fixed this but forgot to do a git add before I committed and pushed it

PR #208, opened 2022-08-16
many: update to latest go-efilib

alfonsosanchezbeato:


alfonsosanchezbeato:
@chrisccoulson could you please take a look at this one when possible? We would need latest go-efilib here to support arm64 (it would be the last bit).

chrisccoulson (approved):

PR #209, opened 2022-08-26
Support measuring extended classic models with a distinct hash

pedronis:
The simplest approach seemed to be to integrate that info as a high
bit in the grade code, for clarity snapd can grow an internal constant
as well to reserve this.

I moved snapd only partially forward because go-tpm2/testutil needs
fixing first, the signature of UserCommonDataDir has changed,
eventually that code will need to consider two possible directories.

### DIFF
--- a/snap.go
+++ b/snap.go
@@ -35,7 +35,9 @@ import (
 
 var sha3_384oid = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 9}
 
-const ClassicCode uint32 = 0x80000000
+// ClassicModelGradeMask is ORed with the model grade code when
+// measuring a classic snap model.
+const ClassicModelGradeMask uint32 = 0x80000000
 
 // SnapModel exposes the details of a snap device model that are bound
 // to an encrypted container.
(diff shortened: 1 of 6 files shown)


snap.go
```diff
@@ -35,12 +35,15 @@ import (
 
 var sha3_384oid = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 9}
 
+const ClassicCode uint32 = 0x80000000
```
chrisccoulson, line 38: As this is exported, it could probably do with a comment.

I'm trying to think of a better name for this, as "ClassicCode" is a bit vague in the context of the wider API, but I can't think of a better way of describing it without making it much longer (eg, ClassicModelGradeMask or something)

chrisccoulson (approved):
I think this looks ok - there's a comment about the naming though

snap.go
pedronis: thanks, I'm fine with the longer name. I'm pushing that together with a doc comment

chrisccoulson:
The update LGTM - thanks!

PR #211, opened 2022-09-05
Support measuring extended classic models with a distinct hash

pedronis:
The simplest approach seemed to be to integrate that info as a high
bit in the grade code, for clarity snapd can grow an internal constant
as well to reserve this.

I moved snapd only partially forward because go-tpm2/testutil needs
fixing first, the signature of UserCommonDataDir has changed,
eventually that code will need to consider two possible directories.

### DIFF
--- a/snap.go
+++ b/snap.go
@@ -1,7 +1,7 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
 
 /*
- * Copyright (C) 2021 Canonical Ltd
+ * Copyright (C) 2021-2022 Canonical Ltd
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License version 3 as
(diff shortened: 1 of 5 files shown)


snap.go
```diff
@@ -35,12 +35,17 @@ import (
 
 var sha3_384oid = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 9}
 
+// ClassicModelGradeMask is ORed with the model grade code when
+// measuring a classic snap model.
+const ClassicModelGradeMask uint32 = 0x80000000
```
alfonsosanchezbeato, line 40: As a follow-up, maybe it could be good to leave a comment in https://github.com/snapcore/snapd/blob/2.57.2/asserts/model.go#L367 mentioning that now the highest bit is reserved as it is used in the model HMAC for differentiating classic/core?

alfonsosanchezbeato (approved):
LGTM

snap.go
pedronis: yes, I plan to do something for that

mvo5 (approved):
Thank you!

PR #215, opened 2022-10-14
tpm2: refactor the execution of PCRProtectionProfile

chrisccoulson:
PCRProtectionProfile contains 2 execution loop implementations - one for String() and one for ComputePCRValues(), each containing fairly messy state management and making use of type assertions for instruction execution.

Clean this up by implementing a single execution loop that makes use of an interface for instruction execution and a handler interface supplied by both of the functions that use it instead.

This enables the opportunity to easily add other methods that might need to "execute" or loop over a profile, eg, to add a method to serialize a profile to a binary format in the future and retaining it in the public part of the key data, which could be useful for debugging purposes.

### DIFF
--- a/tpm2/pcr_profile.go
+++ b/tpm2/pcr_profile.go
@@ -23,6 +23,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
 	"math"
 	"runtime"
 	"strings"
@@ -38,6 +39,7 @@ import (
 const maxPCR = ((math.MaxUint8 + 1) * 8) - 1
 
 // pcrValuesList is a list of PCR value combinations computed from PCRProtectionProfile.
+// It has one entry for each branch created by a profile.
 type pcrValuesList []tpm2.PCRValues
 
 // setValue sets the specified PCR to the supplied value for all branches.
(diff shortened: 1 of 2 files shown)


tpm2/pcr_profile.go
```diff
@@ -77,34 +79,100 @@ func (l pcrValuesList) copy() (out pcrValuesList) {
 	return
 }
 
+// pcrProtectionProfileInstr is a building block of PCRProtectionProfile.
+type pcrProtectionProfileInstr interface {
+	run(context *pcrProtectionProfileExecContext) error
+}
+
+type pcrProtectionProfileInstrList []pcrProtectionProfileInstr
+
```
pedronis, line 88: I think the definitions of type pcrProtectionProfileInstrHandler and pcrProtectionProfileExecContext should come here for clarity

tpm2/pcr_profile.go
```diff
@@ -467,180 +545,248 @@ func (p *PCRProtectionProfile) AddProfileOR(profiles ...*PCRProtectionProfile) *
...
-	return b.String()
+	return nil
 }
 
-// pcrProtectionProfileComputeContext records state used when computing PCR
-// values for a PCRProtectionProfile
-type pcrProtectionProfileComputeContext struct {
-	parent *pcrProtectionProfileComputeContext
-	values pcrValuesList
+type pcrProtectionProfileStringContext struct {
```
pedronis, line 682: it is strange to call this a contex, as now it's really a handler, maybe just pcrProtectionProfileStringifier ?

tpm2/pcr_profile.go
```diff
@@ -467,180 +545,248 @@ func (p *PCRProtectionProfile) AddProfileOR(profiles ...*PCRProtectionProfile) *
...
-	return c.parent == nil
+type pcrProtectionProfileBranchComputeContext struct {
+	values          pcrValuesList
+	subBranchValues pcrValuesList
 }
 
-// pcrProtectionProfileComputeContextStack is a stack of *pcrProtectionProfileComputeContext, with the top of the stack associated
-// with the profile branch from which instructions are currently being processed.
-type pcrProtectionProfileComputeContextStack []*pcrProtectionProfileComputeContext
+type pcrProtectionProfileComputeContext struct {
```
pedronis, line 736: as about these, maybe just s/ComputeContext/Computer/ ?

pedronis (commented):
did a first pass, some high level code org remarks

pedronis (approved):
thank you

PR #217, opened 2022-10-26
keydata: use HKDF to derive the snap model HMAC key

chrisccoulson:
The current implementation uses a DRBG where a simpler KDF would be more appropriate. DRBGs provide some additional features that are just not required here, which are intended to protect prior and future random numbers generated from the same source. All of the extra keys derived from the recovered key payload are all used for protection of the same resource.

HKDF is a core part of the TLS1.3 key schedule. Using this rather than a module we've developed is probably going to make future FIPS 140-3 efforts quite a bit easier given that the go standard library already makes use of it internally (the version it vendors will most likely require openssl integration which we could just copy for external use)

This maintains backwards compatibility with existing keys, with a test to ensure that continues to work.

### DIFF
--- a/keydata.go
+++ b/keydata.go
@@ -29,10 +29,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"hash"
 	"io"
 
 	"github.com/canonical/go-sp800.90a-drbg"
 
+	"golang.org/x/crypto/hkdf"
 	"golang.org/x/xerrors"
 )
 
@@ -40,6 +42,10 @@ const (
 	kdfType                 = "argon2i"
 	passphraseDerivedKeyLen = 32
 	passphraseEncryption    = "aes-cfb"
+)
+
+var (
+	snapModelHMACKDFLabel = []byte("SNAP-MODEL-HMAC")
 )
 
 // ErrNoPlatformHandlerRegistered is returned from KeyData methods if no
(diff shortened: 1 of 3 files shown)


keydata.go
```diff
@@ -255,14 +261,21 @@ type keyDigest struct {
...
 
+type hkdfData struct {
+	Alg  hashAlg `json:"alg"`
+	Salt []byte  `json:"salt"`
+}
+
 type authorizedSnapModelsRaw struct {
 	Alg       hashAlg           `json:"alg"`
+	KDF       *hkdfData         `json:"kdf,omitempty"`
 	KeyDigest json.RawMessage   `json:"key_digest"`
```
pedronis, line 272: it might be a good time to add some comments about the fields' roles

keydata.go
```diff
@@ -404,6 +419,30 @@ func (d *KeyData) snapModelAuthKey(auxKey AuxiliaryKey) ([]byte, error) {
...
+		return d.snapModelAuthKeyLegacy(auxKey)
+	}
+	if !kdf.Alg.Available() {
+		return nil, errors.New("invalid KDF digest algorithm")
+	}
+
+	alg := d.data.AuthorizedSnapModels.alg
+	if alg.Hash == crypto.Hash(0) {
+		return nil, errors.New("invalid digest algorithm")
+	}
```
pedronis, line 434: is there some constraints between alg.Size() and kdf.AlgSize() ?  the code below anyway seems to use the same algo for both

pedronis (commented):
looks good, couple small questions/comments

keydata.go
chrisccoulson: I've added some comments now.

keydata.go
pedronis: > The size of the key we derive should be at least the size of the output of the digest used for the HMAC

shouldn't this bit be checked then?

pedronis (approved):
+1

PR #218, opened 2022-11-11
Print part of TCG log when we could not unseal

alfonsosanchezbeato:
Print part of TCG log when we could not unseal the disk key, before asking for the recovery key. This can give us a hint on what failed.

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -24,13 +24,17 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


alfonsosanchezbeato:
Depends on https://github.com/canonical/tcglog-parser/pull/4

alfonsosanchezbeato:
I will propose to the u22 branch when this one gets merged.

PR #219, opened 2022-11-24
tpm2: update go-tpm2 dependency

chrisccoulson:
This includes a hardening fix for data that is deserialized from disk (see https://github.com/canonical/go-tpm2/commit/1a80c4b47c2b8f0e7a013b4ff5df27d331c48548).

There are some other changes, mostly removal of some checkers from the testutil package because they aren't TPM-specific and don't really belong there.

This adds a few things I want to use in secboot.

### DIFF
--- a/internal/compattest/compattest_test.go
+++ b/internal/compattest/compattest_test.go
@@ -89,7 +89,7 @@ func (s *compatTestSuiteBase) SetUpTest(c *C) {
 		}()
 	}
 
-	simulatorShutdown, err := tpm2_testutil.LaunchTPMSimulator(&tpm2_testutil.TPMSimulatorOptions{SourceDir: s.dataPath})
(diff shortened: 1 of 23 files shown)


pedronis (approved):
lgtm.

The description should also mention that there are some comment reformatting/cleanups going on.

> This adds a few things I want to use in secboot.

does this refer to the new checkers?

chrisccoulson:
> lgtm.
> 
> The description should also mention that there are some comment reformatting/cleanups going on.
> 
> > This adds a few things I want to use in secboot.
> 
> does this refer to the new checkers?

Sorry, yes - I've switched to go 1.19 recently and so gofmt is applying some documentation fixes which is a bit of a pain.

There is a new checker in go-tpm2 that will be useful (TPMValueDeepEquals), but I also want to make use of [tpm2.PCRValues](https://pkg.go.dev/github.com/canonical/go-tpm2#PCRValues) now being a serializable type in a future PR, and an improvement to [mu.CopyValue](https://pkg.go.dev/github.com/canonical/go-tpm2/mu#CopyValue) which makes it possible to copy to an interface type which will simplify a test I'm currently writing for another PR.

PR #220, opened 2022-12-07
tpm2: add Connection.InDALockoutMode() helper

mvo5:
This helper will be used in the snapd code to detect early if the system is already in DA lockout mode so that the install will not even offer to try to install in full-disk-encrypted mode (because this will fail later when the TPM is tried to be provisioned).


chrisccoulson:
I think the test failure here can be fixed by adding `tpm2test.TPMFeatureLockoutHierarchy` in `tpmSuitePlatform.SetUpSuite`

PR #221, opened 2023-01-04
cryptsetup: add support for ICE flag

kubiko:
Adding support for InlineCryptoEngine flag.
This is passed to `cryptsetup` as an additional `--inline-crypto-engine` flag.

This requires a modified `cryptsetup` supporting `inline-crypto-engine` feature.

### DIFF
--- a/internal/luks2/cryptsetup_test.go
+++ b/internal/luks2/cryptsetup_test.go
@@ -211,15 +211,20 @@ func (s *cryptsetupSuite) TestFormatWithCustomKeyslotsAreaSize(c *C) {
 }
 
 func (s *cryptsetupSuite) TestFormatWithInlineCryptoEngine(c *C) {
+	mockCryptsetup := snapd_testutil.MockCommand(c, "cryptsetup", "")
+	defer mockCryptsetup.Restore()
+
 	key := make([]byte, 32)
 	rand.Read(key)
-	s.testFormat(c, &testFormatData{
-		label: "test",
-		key:   key,
-		options: &FormatOptions{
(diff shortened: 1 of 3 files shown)


internal/luks2/cryptsetup_test.go
```diff
@@ -210,6 +210,18 @@ func (s *cryptsetupSuite) TestFormatWithCustomKeyslotsAreaSize(c *C) {
...
+	key := make([]byte, 32)
+	rand.Read(key)
+	s.testFormat(c, &testFormatData{
+		label: "test",
+		key:   key,
+		options: &FormatOptions{
+			KDFOptions:          KDFOptions{MemoryKiB: 32 * 1024, ForceIterations: 4},
+			KeyslotsAreaKiBSize: 2 * 1024,
+			InlineCryptoEngine:  true}})
+}
```
chrisccoulson, line 223: As this test fails in the github workflow and this branch doesn't contain the feature detection parts that would allow it to be skipped, it might be better to just mock this specific test just to verify that the expected commandline arguments are supplied.

chrisccoulson (changes requested):
Thanks - I've added a suggestion for the test.

internal/luks2/cryptsetup_test.go
mvo5: I pushed a fix for this now

chrisccoulson (approved):
Thanks - this looks ok to me

mvo5:
This also needs a version for `master`.

kubiko:
@mvo5 thank you for the test fix.
I have also opened PR for `master` here [#236](https://github.com/snapcore/secboot/pull/236)

PR #223, opened 2023-01-13
tpm2: Squash TPM2_PCR_Read commands

chrisccoulson:
A profile that has AddPCRValueFromTPM called multiple times executes
a TPM2_PCR_Read command for each addition. This is unnecessary as
multiple values can be fetched in a single command.

Reduce this to a minimal sequence of commands by calling
TPMContext.PCRRead once and having go-tpm2 fetch all of the values
in the smallest number of commands possible.


pedronis (approved):
doesn't look unreasonable, I notice though that current policies don't seem to use this. I see only calls to AddPCRValue() and ExtendPCR()

PR #224, opened 2023-01-13
tpm2: elide empty branch points or those with only a single sub-branch

chrisccoulson:
This doesn't affect generated policy, but does improve the readability
of profiles when printed as a string for debugging purposes - especially
given our fairly liberal use of branches in order to simplify profile
construction.


pedronis (approved):
lgtm

PR #226, opened 2023-01-16
tpm2: add internal key sealer interface

chrisccoulson:
This is a new internal interface that is going to be used by the new key
sealing APIs in order to increase code use between them.

### DIFF
--- a/tpm2/key_sealer.go
+++ b/tpm2/key_sealer.go
@@ -1,0 +1,97 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2022 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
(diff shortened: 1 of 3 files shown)


tpm2/key_sealer.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/canonical/go-tpm2"
+	"github.com/canonical/go-tpm2/templates"
+	"github.com/canonical/go-tpm2/util"
+
+	"golang.org/x/xerrors"
+)
+
+// keySealer is an abstraction for creating a sealed key object
+type keySealer interface {
+	CreateSealedObject(data []byte, nameAlg tpm2.HashAlgorithmId, policy tpm2.Digest) (tpm2.Private, *tpm2.Public, tpm2.EncryptedSecret, error)
```
pedronis, line 32: this probably merits a comment so that the argument can be described a bit

tpm2/key_sealer.go
```diff
@@ -0,0 +1,97 @@
...
+	if err != nil {
+		return nil, nil, nil, xerrors.Errorf("cannot create sealed object: %w", err)
+	}
+	return priv, pub, nil, err
+}
+
+// importableObjectKeySealer is an implementation of keySealer that seals keys to
+// a storage key in the form of an importable object.
+type importableObjectKeySealer struct {
+	tpmKey *tpm2.Public
```
pedronis, line 83: this merits a comment, I notice that SealKeyToExternalTPMStorageKey had quite a long comment

pedronis (commented):
thanks, some remarks about having a few more comments

chrisccoulson:
I've added some extra comments to this now.

PR #227, opened 2023-01-16
tpm2: Add V3 metadata support

chrisccoulson:
This adds version 3 metadata support which will be the baseline for the
new style tpm2 keydata objects, although it's currently unused in this
PR. The main difference between v1/v2 and v3 right now is how the auth
key is created, but some follow up PRs are going to make additional
changes.

New keys using the existing APIs will still use the current metadata
version.

### DIFF
--- a/tpm2/keydata_v3.go
+++ b/tpm2/keydata_v3.go
@@ -1,0 +1,136 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019-2023 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package tpm2
+
+import (
+	"bytes"
+	"errors"
+	"io"
+
+	"github.com/canonical/go-tpm2"
+	"github.com/canonical/go-tpm2/mu"
+	"github.com/canonical/go-tpm2/util"
+
+	"golang.org/x/xerrors"
+)
+
+// keyData_v3 represents version 3 of keyData.
+type keyData_v3 struct {
+	KeyPrivate       tpm2.Private
+	KeyPublic        *tpm2.Public
+	KeyImportSymSeed tpm2.EncryptedSecret
+	PolicyData       *keyDataPolicy_v3
+}
+
+func readKeyDataV3(r io.Reader) (keyData, error) {
+	var d *keyData_v3
+	if _, err := mu.UnmarshalFromReader(r, &d); err != nil {
+		return nil, err
+	}
+	return d, nil
+}
+
+func (d *keyData_v3) Version() uint32 {
+	return 3
+}
+
+func (d *keyData_v3) Private() tpm2.Private {
+	return d.KeyPrivate
+}
+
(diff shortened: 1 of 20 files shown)


tpm2/keydata_v3.go
```diff
@@ -0,0 +1,136 @@
...
+	return pcrPolicyCounter, nil
+}
+
+func (d *keyData_v3) Write(w io.Writer) error {
+	_, err := mu.MarshalToWriter(w, d)
+	return err
+}
+
+func (d *keyData_v3) Policy() keyDataPolicy {
+	return d.PolicyData
```
pedronis, line 135: this is not used/covered by tests

tpm2/keydata_v3.go
```diff
@@ -0,0 +1,136 @@
...
+
+func (d *keyData_v3) Imported(priv tpm2.Private) {
+	if d.KeyImportSymSeed == nil {
+		panic("does not need to be imported")
+	}
+	d.KeyPrivate = priv
+	d.KeyImportSymSeed = nil
+}
+
+func (d *keyData_v3) ValidateData(tpm *tpm2.TPMContext, session tpm2.SessionContext) (tpm2.ResourceContext, error) {
```
pedronis, line 74: ValidateData for v1 (which is also used by v2) seems to have much better coverage of the error handling than the similar code here

tpm2/policy.go
```diff
@@ -235,6 +249,46 @@ func newKeyDataPolicy(alg tpm2.HashAlgorithmId, key *tpm2.Public, pcrPolicyCount
...
+	trial.PolicyAuthorize(computeV3PcrPolicyRefFromCounterName(pcrPolicyCounterName), keyName)
+	trial.PolicyAuthValue()
+
+	return &keyDataPolicy_v3{
+		StaticData: &staticPolicyData_v3{
+			AuthPublicKey:          key,
+			PCRPolicyCounterHandle: pcrPolicyCounterHandle},
+		PCRData: &pcrPolicyData_v3{
+			PolicySequence:            pcrPolicySequence,
+			AuthorizedPolicySignature: &tpm2.Signature{SigAlg: tpm2.SigSchemeAlgNull}}}, trial.GetDigest(), nil
```
pedronis, line 262: all previous code wasn't setting AuthorizedPolicySignature at this stage, any particular reason to do so now?

pedronis (commented):
Did a pass, as the code is indeed not yet using the new newKeyDataPolicy it's a bit hard to judge what is going on overall. The description of this should probably be a bit more explicit about the different approach, also I notice that no comment in the code seems to mention the new derivation approach from the aux key, it is just implemented. Some remarks about code coverage and a question.

tpm2/policy.go
chrisccoulson: Setting AuthorizedPolicySignature makes it a valid object so that it can be serialized and unserialized, as the zero value for SigAlg is invalid. This is needed here due to the way sealed object creation works with the new sealing API that's built on this.

tpm2/keydata_v3.go
chrisccoulson: I think this was just an oversight - I pulled this code out of the branch I had for the new sealing API, and I hadn't copied all of the tests across yet. I've added some extra tests now.

chrisccoulson:
I've updated this but an unrelated test has started to time out because the TPM simulator doesn't respond to a command. I've retried it a couple of times to make sure it's not a transient issue, but the same test keeps failing, although it's failing on a different TPM command each time :/

chrisccoulson:
Ah, this isn't an individual test timing out - the issue is that go's testing package has a timeout of 10 minutes for the entire suite, and the new tests in this PR push it over the limit when running in github.

tpm2/keydata_v3.go
chrisccoulson: It looks like there's no explicit coverage for this in previous versions either - it gets used by other code. That said, it looks like the extra tests I've added for ValidateData use this now.

pedronis (approved):
lgtm

https://github.com/snapcore/secboot/pull/227/commits/020ad591f31de365d54f84627d83b483305e3c07 is a bit strange because it seems to include the extra tests but the next commit only mentions them which is a small commit that just tweaks one test

PR #228, opened 2023-01-16
tpm2: Make newKeyData support multiple metadata versions

chrisccoulson:
In order to support the ability to add additional keys to an existing
system in the future, newKeyData must be able to return a keyData
implementation that has the same metadata version as the supplied
keyDataPolicy implementation - this may not always be the latest
version.

### DIFF
--- a/tpm2/keydata.go
+++ b/tpm2/keydata.go
@@ -113,12 +113,34 @@ func readKeyData(r io.Reader, version uint32) (keyData, error) {
 	}
 }
 
-func newKeyData(keyPrivate tpm2.Private, keyPublic *tpm2.Public, importSymSeed tpm2.EncryptedSecret, policy keyDataPolicy) keyData {
-	return &keyData_v2{
-		KeyPrivate:       keyPrivate,
-		KeyPublic:        keyPublic,
-		KeyImportSymSeed: importSymSeed,
-		PolicyData:       policy.(*keyDataPolicy_v2)}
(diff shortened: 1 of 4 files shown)


tpm2/keydata.go
```diff
@@ -113,12 +113,34 @@ func readKeyData(r io.Reader, version uint32) (keyData, error) {
...
+	case *keyDataPolicy_v2:
+		return &keyData_v2{
+			KeyPrivate:       keyPrivate,
+			KeyPublic:        keyPublic,
+			KeyImportSymSeed: importSymSeed,
+			PolicyData:       p}, nil
+	// v1 and v2 are the same. Always return v2 from here, as it's automatically
+	// downgraded to v1 when serialized if it is not importable
+	// case *keyDataPolicy_v1:
+	case *keyDataPolicy_v0:
```
pedronis, line 133: do we need to support v0, the preeexisting code supported only v1/v2?

pedronis (commented):
this looks alright but the summary is a bit confusing as this is making newKeyData that supported v2/v1 also support v3, so I would expect it to say s/older/newer/ or "v3 and older..." ? am I missing something

tpm2/keydata.go
chrisccoulson: I've deleted this now.

chrisccoulson:
> this looks alright but the summary is a bit confusing as this is making newKeyData that supported v2/v1 also support v3, so I would expect it to say s/older/newer/ or "v3 and older..." ? am I missing something

I've updated the summary now.

pedronis (approved):
thank you

PR #230, opened 2023-01-26
Only use the expand part of HKDF

chrisccoulson:
https://github.com/snapcore/secboot/pull/217 introduced HKDF to derive a
HMAC key. The TPM code is going to use HKDF to derive an elliptic
curve key from the same input key. When working on the latter one and
wondering whether to generate another salt for this, I realized that we
only need to perform the expand step of HKDF. The code I introduced is
currently performing both the extract and expand steps.

From the RFC (https://www.rfc-editor.org/rfc/rfc5869):

> In many applications, the input keying material is not necessarily
> distributed uniformly, and the attacker may have some partial
> knowledge about it (for example, a Diffie-Hellman value computed by a
> key exchange protocol) or even partial control of it (as in some
> entropy-gathering applications).  Thus, the goal of the "extract"
> stage is to "concentrate" the possibly dispersed entropy of the input
> keying material into a short, but cryptographically strong,
> pseudorandom key.  In some applications, the input may already be a
> good pseudorandom key; in these cases, the "extract" stage is not
> necessary, and the "expand" part can be used alone.
>
> The second stage "expands" the pseudorandom key to the desired
> length; the number and lengths of the output keys depend on the
> specific cryptographic algorithms for which the keys are needed.

In our case, the input key is secret and cryptopgraphically strong, which
means that the extract step is unnecessary and we can skip it. This also
means that we don't need an additional salt for the key derivation, as
that's only used in this step.

### DIFF
--- a/keydata.go
+++ b/keydata.go
@@ -39,9 +39,10 @@ import (
 )
 
 const (
-	kdfType                 = "argon2i"
-	passphraseDerivedKeyLen = 32
-	passphraseEncryption    = "aes-cfb"
+	kdfType                         = "argon2i"
+	nilHash                 hashAlg = 0
+	passphraseDerivedKeyLen         = 32
+	passphraseEncryption            = "aes-cfb"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pedronis (commented):
what's the backward compatibility story for this? I suppose snapd is not using this yet, so from that POV it doesn't matter

chrisccoulson:
> what's the backward compatibility story for this? I suppose snapd is not using this yet, so from that POV it doesn't matter

There's no backward compatibility with this change, but that should be ok as nothing is using it yet.

pedronis (approved):
+1

PR #232, opened 2023-02-10
keydata: Use the saved key size consistently when using a passphrase

chrisccoulson:
The passphrase derived encryption key size is stored in the key data when a passphrase is set but is then not used consistently after that. It's used to calculate the length of key material to derive from the passphrase, but then the compiled in constant key length is used to index into the returned bytes in order to separate the encryption key from the IV. This will cause a panic if the saved key data is modified to reduce the value of the key size.

This also renames the passphraseDerivdKeyLen constant - it's not the correct description as the actual derived key length is one block size larger than this.

### DIFF
--- a/keydata.go
+++ b/keydata.go
@@ -39,9 +39,9 @@ import (
 )
 
 const (
-	kdfType                 = "argon2i"
-	passphraseDerivedKeyLen = 32
-	passphraseEncryption    = "aes-cfb"
+	kdfType                    = "argon2i"
+	passphraseEncryptionKeyLen = 32
+	passphraseEncryption       = "aes-cfb"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


keydata.go
```diff
@@ -543,10 +543,15 @@ func (d *KeyData) openWithPassphrase(passphrase string, kdf KDF) (payload []byte
 		return nil, nil, fmt.Errorf("unexpected KDF type \"%s\"", data.KDF.Type)
 	}
 	if data.Encryption != passphraseEncryption {
-		// Only AES-256-CFB is supported
+		// Only AES-CFB is supported
 		return nil, nil, fmt.Errorf("unexpected encryption algorithm \"%s\"", data.Encryption)
 	}
+	if data.KeySize > 32 {
```
pedronis, line 549: is there/should there be a minimal KeySize as well?

pedronis (commented):
question

keydata.go
chrisccoulson: This is enforced implicitly when calling aes.NewCipher, which returns an error if it's not 16, 24 or 32 bytes. The maximum is bounds checked here to avoid a potential integer overflow that can result in a panic before calling aes.NewCipher.

pedronis (approved):
lgtm

PR #233, opened 2023-02-10
tpm2: Add passphrase support

chrisccoulson:
(Draft for now because it depends on https://github.com/snapcore/secboot/pull/202)


tpm2/platform.go
```diff
@@ -63,8 +83,21 @@ func (h *platformKeyDataHandler) recoverKeysCommon(data *secboot.PlatformKeyData
...
 	}
+	if !k.data.Public().NameAlg.IsValid() {
+		return nil, &secboot.PlatformHandlerError{
+			Type: secboot.PlatformHandlerErrorInvalidData,
+			Err:  fmt.Errorf("invalid name algorithm for sealed object: %d", k.data.Version())}
+	}
 
-	symKey, err := k.unsealDataFromTPM(tpm.TPMContext, tpm.HmacSession())
+	var authValue []byte
+	if authKey != nil {
```
sespiros, line 93: why not `len(authKey) > 0` similar to calling spots below for consistency?

tpm2/keydata_v0.go
```diff
@@ -188,6 +192,10 @@ func (d *keyData_v0) ValidateData(tpm *tpm2.TPMContext, session tpm2.SessionCont
 	return pcrPolicyCounter, nil
 }
 
+func (d *keyData_v0) DeriveAuthValue(_ []byte) ([]byte, error) {
+	return nil, errors.New("not supported")
+}
```
pedronis, line 197: what is this about? this method is not part of keyData, nor I see calls to methods of this name?

tpm2/platform.go
pedronis: I think this is considering nil as not providing an authKey vs possibly an empty key, though I don't know if we have empty key tests?

tpm2/platform.go
```diff
@@ -111,11 +144,115 @@ func (h *platformKeyDataHandler) RecoverKeys(data *secboot.PlatformKeyData) (sec
...
+			Type: secboot.PlatformHandlerErrorUninitialized,
+			Err:  ErrTPMProvisioning}
+	case err != nil:
+		// This is an unexpected error
+		return nil, xerrors.Errorf("cannot load sealed key object into TPM: %w", err)
+	}
+	defer tpm.FlushContext(keyObject)
+
+	if len(old) > 0 {
+		v, err := deriveAuthValue(old, keyObject.Name().Algorithm().Size())
```
pedronis, line 217: but indeed here the code does something else, does the derivation turns an empty passphrase into an empty key? otherwise this doesn't seem quite right

tpm2/platform.go
```diff
@@ -111,11 +144,115 @@ func (h *platformKeyDataHandler) RecoverKeys(data *secboot.PlatformKeyData) (sec
...
+		}
+	}
+
+	priv, err := tpm.ObjectChangeAuth(keyObject, srk, newAuthValue, tpm.HmacSession().IncludeAttrs(tpm2.AttrCommandEncrypt))
+	if err != nil {
+		if tpm2.IsTPMSessionError(err, tpm2.ErrorAuthFail, tpm2.CommandObjectChangeAuth, 1) {
+			return nil, &secboot.PlatformHandlerError{
+				Type: secboot.PlatformHandlerErrorInvalidAuthKey,
+				Err:  err}
+		}
```
pedronis, line 238: shouldn't the case where the wrong old key is provided be tested?, where will that fail? here?

pedronis (commented):
did a pass, some comments/questions

tpm2/platform.go
chrisccoulson: I think checking `len(authKey) > 0` is fine - it is passed nil when no passphrase is set, but it's always passed a key with a length greater than zero when a passphrase is set.

tpm2/keydata_v0.go
chrisccoulson: That's just some code I forgot to delete.

tpm2/platform.go
chrisccoulson: `old` in this context is not a passphrase - it's a passphrase derived key. It always has a non-zero size when a passphrase is set, regardless of the length of the passphrase (which could be zero).

tpm2/platform.go
chrisccoulson: I'll add some extra tests for this.

pedronis (approved):
+1

PR #234, opened 2023-02-10
Rename KeyCreationData to KeyCreationParams

chrisccoulson:
"Creation data" is a bit of an overloaded term when using a TPM, because it specifically refers to data that can be used to prove the creation environment of a TPM object. I'm going to add support to the tpm2 package for saving the creation data to facilitate subsequent attestation of the state of the initial install environment, so rename this struct to avoid confusion.

The PR I've proposed to add a new ProtectKeyWithTPM API (https://github.com/snapcore/secboot/pull/202) also uses "params" in this context, so it's consistent with that.

### DIFF
--- a/keydata.go
+++ b/keydata.go
@@ -151,9 +151,9 @@ const (
 	AuthModePassphrase AuthMode = 1 << iota
 )
 
-// KeyCreationData is the data required to create a new KeyData object.
+// KeyCreationParams provides parameters required to create a new KeyData object.
 // It should be produced by a platform implementation.
-type KeyCreationData struct {
+type KeyCreationParams struct {
 	// Handle contains metadata required by the platform in order to recover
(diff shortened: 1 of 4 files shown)


keydata.go
```diff
@@ -151,9 +151,9 @@ const (
 	AuthModePassphrase AuthMode = 1 << iota
 )
 
-// KeyCreationData is the data required to create a new KeyData object.
+// KeyCreationParams provides parameters required to create a new KeyData object.
 // It should be produced by a platform implementation.
-type KeyCreationData struct {
+type KeyCreationParams struct {
```
pedronis, line 156: in principle this could also be just called KeyParams?

pedronis (approved):
+1, optional suggestion

keydata.go
chrisccoulson: Yeah, that's ok with me

PR #235, opened 2023-02-28
cryptsetup: add support for ICE flag (master)

mvo5:
Cherry pick https://github.com/snapcore/secboot/pull/221 back into master

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -554,13 +554,17 @@ type InitializeLUKS2ContainerOptions struct {
 	// the initial keyslot. If this is empty, then the name will be
 	// set to "default".
 	InitialKeyslotName string
+
+	// InlineCryptoEngine set flag if to use Inline Crypto Engine
+	InlineCryptoEngine bool
 }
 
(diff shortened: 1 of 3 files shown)


mvo5:
Hey @chrisccoulson - I think we need this in order to move to secboot@latest in snapd - it fails right now with https://github.com/snapcore/snapd/actions/runs/6261001841/job/17000166011?pr=13210#step:3:6849 :
```
../secboot/encrypt_sb.go:73:3: unknown field 'InlineCryptoEngine' in struct literal of type "github.com/snapcore/secboot".InitializeLUKS2ContainerOptions
```
(and a bunch more but that is one low hanging fruit).

chrisccoulson (approved):
I think this looks ok to unblock snapd, but it would be worth following up with adding another flag to `DetectCryptsetupFeatures` (which didn't exist on the previous branch) and adding some unit tests conditional on this that make use of the real cryptsetup binary.

xnox:
How is this supported on classic 22.04 and 23.10 hybrid systems?

PR #238, opened 2023-03-13
efi: updates to ImageLoadEvent

chrisccoulson:
This is the start of a series of changes that significantly refactor the code in the efi package. These changes will:
- Make the code more modular and maintainable.
- Add support for applying secure boot revocations, including updating shim.
- Replace the existing AddSecureBootPolicyProfile and AddBootManagerCodeProfile APIs with a single API that will add an entire profile. This will also replace the need to use the existing AddSystemdStubProfile and tpm2.AddSnapModelProfile APIs.

This last one requires the ability to specify external parameters that are not part of an executable binary, which is partly what this PR adds. Note that this does add some APIs that aren't used yet. It also implements ImageLoadEventSource as one of these new parameters, which is a bit strange - this is temporary because it's going to be removed entirely in a future PR, but it needs to stay for now to keep the existing code working.

### DIFF
--- a/efi/image.go
+++ b/efi/image.go
@@ -1,0 +1,240 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2019 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package efi
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/snapcore/secboot"
+	"github.com/snapcore/snapd/snap"
+)
+
+// ImageReader corresponds to an open handle from which to read a binary image from.
+type ImageReader interface {
+	io.ReaderAt
+	io.Closer
+	Size() int64
+}
+
+// Image provides an image that is loaded during the boot process before ExitBootServices.
+type Image interface {
+	fmt.Stringer
+	Open() (ImageReader, error) // Open a handle to the image for reading
+}
+
+// SnapFileImage provides an image contained within a snap package that is loaded
+// during the boot process before ExitBootServices.
+type SnapFileImage struct {
+	Container snap.Container
+	FileName  string // The filename within the snap squashfs
+}
+
+// NewSnapFileImage creates a new SnapFileImage for the file at the
+// specified relative path within the supplied snap.
+func NewSnapFileImage(container snap.Container, filename string) *SnapFileImage {
+	return &SnapFileImage{
+		Container: container,
+		FileName:  filename}
+}
+
+// String implements [fmt.Stringer].
+func (f SnapFileImage) String() string {
+	return fmt.Sprintf("%#v:%s", f.Container, f.FileName)
+}
+
+// Open implements [ImageSource.Open].
+func (f SnapFileImage) Open() (ImageReader, error) {
+	return f.Container.RandomAccessFile(f.FileName)
+}
+
+type fileImageReader struct {
+	*os.File
+	size int64
+}
+
+func (h *fileImageReader) Size() int64 {
+	return h.size
+}
+
+// FileImage provides an image from a file that is loaded during the boot process
+// before ExitBootServices.
+type FileImage string
+
+// NewFileImage creates a new FileImage for the file at the specified path.
+func NewFileImage(path string) FileImage {
+	return FileImage(path)
+}
+
+// String implements [fmt.Stringer].
+func (p FileImage) String() string {
+	return string(p)
+}
+
+// Open implements [ImageSource.Open].
+func (p FileImage) Open() (ImageReader, error) {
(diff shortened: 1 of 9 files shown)


efi/image.go
```diff
@@ -0,0 +1,240 @@
...
+		f.Close()
+		return nil, err
+	}
+	return &fileImageReader{File: f, size: fi.Size()}, nil
+}
+
+// ImageLoadParams provides one or more values for an external parameter that
+// is supplied to an image which is loaded during the boot process.
+type ImageLoadParams interface {
+	apply(params ...loadParams) []loadParams
```
pedronis, line 109: given how it seems to work, applyTo might be a better name for this method?

efi/image.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+type baseImageLoadEvent struct {
+	sourceImage Image
+	nextImages  []ImageLoadEvent
+	loadParams  imageLoadParamsSet
+}
+
+func (e *baseImageLoadEvent) Next(images ...ImageLoadEvent) ImageLoadEvent {
+	e.nextImages = images
```
pedronis, line 210: shouldn't this be an append? or fail if nextImages is already set? or should Next be called SetNext to fit more golang naming conventions?

efi/secureboot_policy.go
```diff
@@ -879,7 +879,14 @@ func (g *secureBootPolicyGen) processOSLoadEvent(branches []*secureBootPolicyGen
 		return xerrors.Errorf("cannot determine image type: %w", err)
 	}
 
-	if err := g.computeAndExtendVerificationMeasurement(branches, r, event.Source); err != nil {
+	var source ImageLoadEventSource
+	params := event.params().Resolve(new(loadParams))
```
pedronis, line 883: are there tests where NewImageLoadEvent is passed loadParams in and this matters? I don't see new tests in the test file for this file but maybe I miss the new tests/relevant changes

efi/image.go
```diff
@@ -0,0 +1,240 @@
...
+
+func (s imageLoadParamsSet) Resolve(inherited *loadParams) []loadParams {
+	params := []loadParams{*inherited}
+	for _, p := range s {
+		params = p.apply(params...)
+	}
+	return params
+}
+
+type loadParams struct {
```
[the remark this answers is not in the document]
chrisccoulson: I've moved this now.

efi/image.go
chrisccoulson: I've renamed this.

efi/image.go
chrisccoulson: SetNext looked a bit strange when I was writing the tests for this, which is why it ended up as Next. Given the information that the API is intended to capture, I wonder whether this should be a verb of some kind instead, so it looks something like this:
```
    NewImageLoadEvent("/shimx64.efi").Loads(
        NewImageLoadEvent("/grubx64.efi").Loads(
            NewImageLoadEvent("/kernel1.efi"),
            NewImageLoadEvent("/kernel2.efi"),
        )
    )
```

Although in this case, `ImageLoadEvent` sounds a bit weird as well - maybe it should be `ImageLoadData` instead.

efi/image.go
```diff
@@ -0,0 +1,240 @@
...
+	return e.nextImages
+}
+
+func (e *baseImageLoadEvent) params() imageLoadParamsSet {
+	return e.loadParams
+}
+
+type imageLoadParamsSet []ImageLoadParams
+
+func (s imageLoadParamsSet) Resolve(inherited *loadParams) []loadParams {
```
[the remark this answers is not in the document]
chrisccoulson: initial is probably ok here - the intention here is that the initial properties are inherited, although that can be expressed in the code that uses this API.

efi/secureboot_policy.go
chrisccoulson: I'm not sure I understand the question - there are quite a few tests that pass parameters to NewImageLoadEvent, eg, all of the ones in efi/secureboot_policy_test.go, such as https://github.com/snapcore/secboot/pull/238/files#diff-b20e11ef97bb6e34236a8af13ffa1af982feea79b1317d8c300169c4c123682aR305

I probably should make clear in the code but this specific use of loadParams here is temporary - the new code determines the appropriate behaviour here automatically.

efi/image.go
pedronis: yes, a verb would be clearer, not a fan of struct called *Data though, one option s/Event/Activity/

efi/secureboot_policy.go
pedronis: sorry it seems I didn't look close enough but they are all using ImageLoadEvenSource?  are you saying that this call to Resolve will go away, it's the only call to Resolve so far? or the usage of ImageLoadEvenSource (that one is commented as going away)

efi/image.go
chrisccoulson: I've called this `ImageLoadActivity` and changed the method to `Loads` now.

efi/image.go
```diff
@@ -0,0 +1,240 @@
...
+		out = append(out, param)
+	}
+	return out
+}
+
+// ImageLoadEvent corresponds to the execution of an image during the boot
+// process, before ExitBootServices.
+type ImageLoadEvent interface {
+	// Next specifies a set of images that are permitted to be executed by the
+	// image associated with this event.
```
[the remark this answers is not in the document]
chrisccoulson: This is renamed now (and the document note has changed as well)

efi/secureboot_policy.go
chrisccoulson: Yes, the usage of `ImageLoadEventSource` will be going away in a subsequent PR, but also this call to `Resolve` as this code is going to be replaced as well - there will be another call to `Resolve` that will use a `loadParams` as input which provides inherited parameters for the current node in the profile generation. It's probably easier for me to explain it if I push the current WIP code somewhere.

pedronis (approved):
+1

PR #245, opened 2023-04-12
Update github.com/canonical/go-efilib dependency

chrisccoulson:
This addresses an issue where some valid signed variable update
payloads could fail to parse correctly because although the UEFI
spec permits signatures with the PKCS#7 SignedData wrapped in an
optional outer ContentInfo structure, go-efilib didn't recognize
signatures in the wrapped case. I discovered this when creating
mock signatures for unit tests in secboot.

This also bumps secboot's go requirement to 1.18 to match snapd,
as go-efilib is pulling in the latest version of x/cryptobyte which
has dependencies that require this as well now.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -9,11 +9,12 @@ jobs:
     runs-on: ubuntu-20.04
     env:
       GOPATH: ${{ github.workspace }}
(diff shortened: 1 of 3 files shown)


pedronis (approved):
thx

PR #247, opened 2023-04-17
efi: Add workaround to make it possible to use shim 15.7

chrisccoulson:
This is a temporary solution which should make it possible to work with shim 15.7. It corrects a typo in one of the host SbatLevel payloads, and also adds a workaround for a bug introduced into shim 15.7. It borrows a little bit of code from a PR for master (shimVersion).

### DIFF
--- a/efi/secureboot_policy.go
+++ b/efi/secureboot_policy.go
@@ -32,6 +32,8 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
+	"strconv"
 	"strings"
 
 	efi "github.com/canonical/go-efilib"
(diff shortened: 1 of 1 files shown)


chrisccoulson:
Note that this makes all of the tests fail because the test binaries don't have the right section in them.

mvo5 (approved):
Thanks, this looks reasonable to me

efi/secureboot_policy.go
```diff
@@ -892,6 +969,14 @@ func (g *secureBootPolicyGen) processShimExecutableLaunch(branches []*secureBoot
 		flags |= shimVariableAuthorityEventsMatchSpec
 	}
 
+	version, err := shim.version()
+	if err != nil {
+		return xerrors.Errorf("cannot determine shim version: %w", err)
+	}
+	if version.Compare(shimVersion{15, 7}) >= 0 {
```
pedronis, line 976: as discussed, a gadget with a fixed shim will need to assume a snapd that stops doing this or produces branches both for the fixed shim and the buggy shim

pedronis (approved):
lgtm

mvo5:
This was tested now on real devices and seems to work fine so I will merge it.

PR #248, opened 2023-04-28
efi: allow the case of minor version not present

alfonsosanchezbeato:
So `$Version: 15 $` is fine in the .data.ident section.

### DIFF
--- a/efi/secureboot_policy.go
+++ b/efi/secureboot_policy.go
@@ -159,7 +159,7 @@ func (s *shimImageHandle) hasSbatSection() bool {
 	return s.openSection(".sbat") != nil
 }
 
-var shimIdentVersionRE = regexp.MustCompile(`^\$Version:[[:blank:]]*([[:digit:]]+)\.([[:digit:]]+)[[:blank:]]*\$$`)
+var shimIdentVersionRE = regexp.MustCompile(`^\$Version:[[:blank:]]*([[:digit:]]+)\.?([[:digit:]]+)?[[:blank:]]*\$$`)
 
 type shimVersion struct {
 	Major uint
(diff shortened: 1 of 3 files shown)


efi/secureboot_policy.go
```diff
@@ -159,7 +159,7 @@ func (s *shimImageHandle) hasSbatSection() bool {
 	return s.openSection(".sbat") != nil
 }
 
-var shimIdentVersionRE = regexp.MustCompile(`^\$Version:[[:blank:]]*([[:digit:]]+)\.([[:digit:]]+)[[:blank:]]*\$$`)
+var shimIdentVersionRE = regexp.MustCompile(`^\$Version:[[:blank:]]*([[:digit:]]+)\.?([[:digit:]]+)?[[:blank:]]*\$$`)
```
pedronis, line 162: maybe the optional part can use a single (?:\\. ...)?

efi/secureboot_policy.go
```diff
@@ -198,14 +192,17 @@ func (s *shimImageHandle) version() (shimVersion, error) {
 	}
 	for scanner.Scan() {
 		m := shimIdentVersionRE.FindStringSubmatch(scanner.Text())
-		if len(m) == 3 {
+		if len(m) > 1 {
```
pedronis, line 195: this could be clearer as >= 2

efi/secureboot_policy.go
```diff
@@ -198,14 +192,17 @@ func (s *shimImageHandle) version() (shimVersion, error) {
...
 			major, err := strconv.ParseUint(m[1], 10, 0)
 			if err != nil {
 				return shimVersion{}, fmt.Errorf("invalid major version \"%s\"", m[1])
 			}
-			minor, err := strconv.ParseUint(m[2], 10, 0)
-			if err != nil {
-				return shimVersion{}, fmt.Errorf("invalid minor version \"%s\"", m[2])
+			var minor uint64
+			if len(m) > 2 && len(m[2]) > 0 {
```
pedronis, line 201: and this as len(m) == 3 as we know how many capturing groups we have?

pedronis (approved):
lgtm, some non-blocking minor comments

efi/secureboot_policy_test.go
```diff
@@ -1644,3 +1644,28 @@ func (s *securebootPolicySuite) TestAddSecureBootPolicyDualSignedShim2(c *C) {
...
+	for _, tc := range []struct {
+		inp               string
+		expectedErr       string
+		expectedShimMajor uint
+		expectedShimMinor uint
+	}{
+		{"", "empty .data.ident section", 0, 0},
+		{"UEFI SHIM", "cannot determine version - missing from .data.ident section", 0, 0},
+		{"UEFI SHIM\n$Version: 15.4 $", "", 15, 4},
+		{"UEFI SHIM\n$Version: 15 $", "", 15, 0},
```
pedronis, line 1658: do we need a test with 15. ?

pedronis: what happens with `$Version: $` ?

chrisccoulson (approved):
Thanks, this LGTM

efi/secureboot_policy.go
alfonsosanchezbeato: It looks like that would convert the whole of the optional part in non-capturing.

pedronis: I think capturing groups are allowed inside non-capturing ones? unless my memory is off

efi/secureboot_policy.go
alfonsosanchezbeato: Looking into more detail how `FindStringSubmatch` works, it seems that either it returns nil or it is successful and it returns an array with 1 + number of groups elements (there needs to be a way to know which group is an element of the array for), so the checks on len(m) are not needed, we just need to check `len(m[2])` here. But probably not worth changing this now as this code will not be used in master afaik.

PR #249, opened 2023-06-05
efi: AddPCRProfile support for PCRs 4 and 7 profiles

chrisccoulson:
This adds the remaining missing pieces so that AddPCRProfile can
generate boot manager code and secure boot policy profiles. This will
replace the existing AddBootManagerProfile and AddSecureBootPolicyProfile
APIs.


efi/image_rules_defs.go
```diff
@@ -0,0 +1,202 @@
...
+		),
+		// Grub
+		newImageRule(
+			"grub",
+			imageSectionExists(".mods"),
+			newGrubLoadHandler,
+		),
+		// TODO: add rules for Ubuntu Core UKI and Ubuntu grub that are not part of
+		// the MS UEFI CA?
+		//
```
pedronis, line 194: what's the state of things, do we need to fix this TODO now?

efi/secureboot.go
```diff
@@ -49,6 +49,38 @@ type SignatureDBUpdate struct {
...
+type signatureDBUpdatesOption []*SignatureDBUpdate
+
+func (u signatureDBUpdatesOption) applyOptionTo(gen *pcrProfileGenerator) {
+	gen.varModifiers = append(gen.varModifiers, func(rootVars *rootVarsCollector) error {
+		for _, root := range rootVars.PeekAll() {
+			for _, quirk := range []signatureDBUpdateFirmwareQuirk{
+				signatureDBUpdateNoFirmwareQuirk,
+				signatureDBUpdateFirmwareDedupIgnoresOwner} {
+				// create branch by copying the root varBranch
+				branch := *root
```
pedronis, line 61: I'm probably missing something, how does the change propagate out of this function if below we operate on a copy?

efi/uc_uki_load_handler.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+type ubuntuCoreUKILoadHandler struct{}
+
+func newUbuntuCoreUKILoadHandler(_ peImageHandle) (imageLoadHandler, error) {
+	return new(ubuntuCoreUKILoadHandler), nil
+}
+
+func (h *ubuntuCoreUKILoadHandler) MeasureImageStart(_ pcrBranchContext) error {
+	// TODO: Add stuff that the kernel measures here
```
pedronis, line 33: do we need to fix this before things can be used?

pedronis (commented):
did a first pass, some questions

efi/image_rules_defs.go
chrisccoulson: We don't, but I'll probably add a section to our UKI to allow it to be identified more easily

efi/uc_uki_load_handler.go
chrisccoulson: It doesn't - snapd can continue to use the `AddSystemdStubProfile` API in addition to this one, although that will be moved here in a follow up. That will be a fairly minor change though.

efi/secureboot.go
chrisccoulson: The copy maintains a copy of updates, and then each call to `WriteVar` propagates those updates to `rootVarsCollector` via its `registerUpdates` callback.

pedronis: I see maybe the comment should be:

// create a branch per quirk by copying the root varBranch

efi/uc_uki_load_handler.go
pedronis: maybe it's a good idea to leave a comment also about that:

// for now clients can continue using AddSystemdStubProfile

efi/secureboot.go
pedronis: for a moment I was confused and thought that root.updates was a slice, that is not case, otherwise the shallow copy would not be safe, as append to the same slice again can produce confusing results depending on the allocation history of the slice. I don't think we have code here that shallow copy structs with slices in them?

pedronis (approved):
thanks, some small final comments

PR #250, opened 2023-06-05
tpm2: Split up SealedKeyObject

chrisccoulson:
SealedKeyObject will be deprecated by a follow up PR, and a new API
will be introduced to replace it. This splits it into code that can be
reused by the new API (sealedKeyDataBase) and moves the existing
API to a new file.


tpm2/policy_v3.go
```diff
@@ -44,6 +44,10 @@ func computeV3PcrPolicyRefFromCounterContext(context tpm2.ResourceContext) tpm2.
 	return computeV1PcrPolicyRefFromCounterContext(context)
 }
 
+func computeV3PcrPolicyCounterAuthPolicies(alg tpm2.HashAlgorithmId, updateKeyName tpm2.Name) tpm2.DigestList {
```
pedronis, line 47: this is not used yet?

tpm2/update_test.go
```diff
@@ -1,153 +0,0 @@
...
-		PCRPolicyCounterHandle: s.NextAvailableHandle(c, 0x01810000)})
-	c.Check(err, ErrorMatches, "invalid key data: cannot complete authorization policy assertions: the PCR policy has been revoked")
-}
-
-func (s *updateSuite) TestRevokeOldPCRProtectionPoliciesWithoutPCRPolicyCounter(c *C) {
-	err := s.testRevokeOldPCRProtectionPolicies(c, &KeyCreationParams{
-		PCRProfile:             tpm2test.NewPCRProfileFromCurrentValues(tpm2.HashAlgorithmSHA256, []int{7, 23}),
-		PCRPolicyCounterHandle: tpm2.HandleNull})
-	c.Check(err, IsNil)
-}
```
pedronis, line 153: what happened to these tests?

tpm2/keydata.go
pedronis, line 1: A bit unrelated and it's true also on master but I notice this had and the split parts have relatively low coverage

pedronis (commented):
did a pass, as it's mostly a reorg just some small comments

tpm2/policy_v3.go
chrisccoulson: I've moved this to the follow up PR.

tpm2/update_test.go
chrisccoulson: Ah, I missed those when I was moving things around. I've added them back again.

pedronis (approved):
thanks

PR #252, opened 2023-06-07
luks2: use "aes-cbc-essiv:sha256" on 32bit arm HW

mvo5:
Most hardware should use aes-xts-plain64 for LUKS2 but on 32bit arm the XTS mode cannot be accelerated by the hardware which leads to very poor perfomance (less than half compared to cbc mode). Therefore this commit switches the default on 32bit ARM to aes-cbc-essiv:sha256.

Note that this is a simple approach, we could run "cryptsetup benchmark" but that seems over engineered (and easy enough to switch if we need in the future).

### DIFF
--- a/internal/luks2/cryptsetup.go
+++ b/internal/luks2/cryptsetup.go
@@ -27,6 +27,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"runtime"
 	"strconv"
 	"sync"
 	"time"
@@ -255,6 +256,27 @@ func (options *FormatOptions) appendArguments(args []string) []string {
 	return args
 }
 
+var runtimeGOARCH = runtime.GOARCH
+
+// selectCipher will return the cipher to use. This is aes-xts-plain64
+// everywhere except on armhf (32bit arm) hardware where the XTS mode
+// cannot be accelerated by the hardware.
+//
+// Note that this is a simple approach, we could run "cryptsetup
+// benchmark" but that seems over engineered (and easy enough to
+// switch if we need in the future).
+func selectCipherAndKeysize() (string, int) {
+	switch runtimeGOARCH {
+	case "arm":
+		// on 32bit arm CPUs xts cannot be accerlated in HW so
+		// cbc is used
+		return "aes-cbc-essiv:sha256", keySize * 4
+	default:
+		// use AES-256 with XTS block cipher mode (XTS requires 2 keys)
+		return "aes-xts-plain64", keySize * 8
+	}
+}
+
 // Format will initialize a LUKS2 container with the specified options and set the primary key to the
 // supplied key. The label for the new container will be set to the supplied label. This can only be
 // called on a device that is not mapped.
@@ -274,6 +296,7 @@ func Format(devicePath, label string, key []byte, opts *FormatOptions) error {
 		return err
 	}
 
+	cipher, ksize := selectCipherAndKeysize()
 	args := []string{
 		// batch processing, no password verification for formatting an existing LUKS container
 		"-q",
@@ -283,8 +306,8 @@ func Format(devicePath, label string, key []byte, opts *FormatOptions) error {
 		"--type", "luks2",
 		// read the key from stdin
 		"--key-file", "-",
-		// use AES-256 with XTS block cipher mode (XTS requires 2 keys)
-		"--cipher", "aes-xts-plain64", "--key-size", strconv.Itoa(keySize * 8),
+
+		"--cipher", cipher, "--key-size", strconv.Itoa(ksize),
 		// set LUKS2 label
 		"--label", label}
 
--- a/internal/luks2/cryptsetup.go
+++ b/internal/luks2/cryptsetup.go
@@ -55,8 +55,6 @@ var (
 
 	features     Features
 	featuresOnce sync.Once
-
-	keySize = 64
 )
 
 // Features indicates the set of features supported by this package,
@@ -209,7 +207,7 @@ type FormatOptions struct {
 	KDFOptions KDFOptions
 }
 
-func (options *FormatOptions) validate() error {
+func (options *FormatOptions) validate(cipher string) error {
 	if (options.MetadataKiBSize != 0 || options.KeyslotsAreaKiBSize != 0) &&
 		DetectCryptsetupFeatures()&FeatureHeaderSizeSetting == 0 {
 		return ErrMissingCryptsetupFeature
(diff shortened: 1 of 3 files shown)


mvo5:
I updated the tests and this is now ready for a proper review/merge.

setharnold:
I'm curious what the performance difference is between the two. I wouldn't have expected much one way or the other.

Thanks

internal/luks2/cryptsetup.go
```diff
@@ -255,6 +256,27 @@ func (options *FormatOptions) appendArguments(args []string) []string {
...
+//
+// Note that this is a simple approach, we could run "cryptsetup
+// benchmark" but that seems over engineered (and easy enough to
+// switch if we need in the future).
+func selectCipherAndKeysize() (string, int) {
+	switch runtimeGOARCH {
+	case "arm":
+		// on 32bit arm CPUs xts cannot be accerlated in HW so
+		// cbc is used
+		return "aes-cbc-essiv:sha256", keySize * 4
```
chrisccoulson, line 273: Whilst this creates the correct key size, use of the keySize variable here is a bit weird because that's currently 64, which is the key size for AES-256-XTS in bytes (hence the `keySize * 8` that exists to convert it to bits). It might be better to get rid of the keySize variable and have a function that returns the correct value based on the selected cipher.

internal/luks2/cryptsetup.go
```diff
@@ -255,6 +256,27 @@ func (options *FormatOptions) appendArguments(args []string) []string {
...
+// cannot be accelerated by the hardware.
+//
+// Note that this is a simple approach, we could run "cryptsetup
+// benchmark" but that seems over engineered (and easy enough to
+// switch if we need in the future).
+func selectCipherAndKeysize() (string, int) {
+	switch runtimeGOARCH {
+	case "arm":
+		// on 32bit arm CPUs xts cannot be accerlated in HW so
+		// cbc is used
```
chrisccoulson, line 272: The underlying Arm32 instruction set doesn't have dedicated instructions for AES at all - in this case we're relying on a feature specific to this SoC which can accelerate AES but IIUC relies on a proprietary kernel blob that doesn't expose an API for doing AES-XTS (so i don't think it's strictly a hardware limitation).

chrisccoulson (changes requested):
Thanks for looking at this - I've added a couple of comments.

internal/luks2/cryptsetup.go
setharnold: Should we / can we / may we switch to xchacha20 or some similar, newer, algorithm? AES implemented in software is usually susceptible to sidechannel data leaks.

internal/luks2/cryptsetup.go
mvo5: Thank you, yeah, my approach is a bit awkward indeed, I reworked it now based on your suggestion :)

internal/luks2/cryptsetup.go
chrisccoulson: > Should we / can we / may we switch to xchacha20 or some similar, newer, algorithm? AES implemented in software is usually susceptible to sidechannel data leaks.

Ah, you might be missing a bit of context here - this is for a specific enablement that is armv7 based where there is no accelerated CPU instructions but the SoC does contain a crypto module that's capable of accelerated AES. The issue that this PR addresses is that this hardware module is used by the kernel's crypto code via a proprietary blob supplied by the vendor which exposes high level APIs for performing AES encryption with CBC mode and a few others, but it doesn't support XTS and nor does it offer an API that's at a low enough level to implement this. So this PR enables hardware accelerated crypto for this specific enablement, which is the only way that the device is usable.

pedronis (approved):
looks reasonable to me

internal/luks2/cryptsetup_test.go
```diff
@@ -1009,3 +1015,41 @@ func (s *cryptsetupSuite) TestSetSlotPriority3(c *C) {
 		slotId:   1,
 		priority: SlotPriorityIgnore})
 }
+
+func (s *cryptsetupSuite) TestSelectCipherAndKeysize(c *C) {
```
pedronis, line 1019: shouldn't we run this only one of the suites or a different suite?

chrisccoulson (approved):
Thanks, this looks good to me.

internal/luks2/cryptsetup_test.go
mvo5: Thanks, I updated this and also improved/fixed the TestFormatOptionsValidateBadKeyslotsAreaSize test

internal/luks2/cryptsetup_test.go
```diff
@@ -180,13 +183,19 @@ func (s *cryptsetupSuite) TestFormatOptionsValidateBadMetadataSize(c *C) {
...
+		// smaller than the minimum
+		{KeyslotsAreaKiBSize: minSize},
+		// must be multiply of 1024
 		{KeyslotsAreaKiBSize: (4 * 1024) + 1},
+		// can't be more than 128KiB
 		{KeyslotsAreaKiBSize: 256 * 1024},
 	} {
 		c.Check(Format("/dev/null", "", make([]byte, 32), &opts), ErrorMatches,
-			fmt.Sprintf("cannot set keyslots area size to %v KiB", opts.KeyslotsAreaKiBSize))
+			fmt.Sprintf("cannot set keyslots area size to %v KiB", opts.KeyslotsAreaKiBSize), Commentf("%v", opts))
```
pedronis, line 198: I let @chrisccoulson review this

pedronis (approved):
other test changes are good

internal/luks2/cryptsetup_test.go
```diff
@@ -180,13 +183,19 @@ func (s *cryptsetupSuite) TestFormatOptionsValidateBadMetadataSize(c *C) {
 }
 
 func (s *cryptsetupSuite) TestFormatOptionsValidateBadKeyslotsAreaSize(c *C) {
+	// minimum size of a keyslot is keysize*4000 in KiB
+	// (250 KiB for 64bit keys, 125 for 32bit keys)
+	minSize := (KeySize(SelectCipher()) * 4000) / 1024
 	for _, opts := range []FormatOptions{
-		{KeyslotsAreaKiBSize: 128},
+		// smaller than the minimum
+		{KeyslotsAreaKiBSize: minSize},
```
chrisccoulson, line 191: I was a bit confused about why this works because this is actually just the minimum size, but it passes because it actually fails the "must be a multiple of 4KiB" check.

internal/luks2/cryptsetup_test.go
```diff
@@ -180,13 +183,19 @@ func (s *cryptsetupSuite) TestFormatOptionsValidateBadMetadataSize(c *C) {
...
+		// must be multiply of 1024
```
chrisccoulson, line 192: "multiple of 4KiB"

internal/luks2/cryptsetup_test.go
```diff
@@ -180,13 +183,19 @@ func (s *cryptsetupSuite) TestFormatOptionsValidateBadMetadataSize(c *C) {
...
 		{KeyslotsAreaKiBSize: (4 * 1024) + 1},
+		// can't be more than 128KiB
```
chrisccoulson, line 194: I think this should say "128MiB"

chrisccoulson (changes requested):
Thanks - I was just about to approve this with a couple of small comment nits when I realized that one of the test cases is passing when I expected it not to.

chrisccoulson (approved):

PR #253, opened 2023-06-08
efi: use go:embed for source testdata and recreate data during the test

chrisccoulson:
The tests in efi/ rely on a lot of testdata that is generated out of
band using tools/make-efi-testdata and loaded from files. It can be
burdonsome to update this in order to support additional test cases.

Start embedding source testdata with go:embed, and recreate additional
data from that when running the tests, which will make it easier to
add test cases in the future. This also updates the new tests added
as part of this sequence of PRs to use this new testdata. Existing
tests are left untouched - when this code is removed, some of the
existing test data can be cleaned up.

An advantage of this approach is that we can make use of the fact that
peImageHandle and shimImageHandle are interfaces now, which means that
they can be mocked in tests with real signatures copied from binaries
that we have published. This will allow us to have tests that create
actual production-like digests.


internal/efitest/vars.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+// VarEntry describes the contents of a mock EFI variable.
+type VarEntry struct {
+	Attrs   efi.VariableAttributes
+	Payload []byte
+}
+
+type VarPayloadWriter interface {
+	Write(w io.Writer) error
```
pedronis, line 39: nitpick: this is a naming issue not here but in the efi package itself, but usually methods writing to a io.Writer are called WriteTo and not just Write

pedronis (approved):
thanks, mostly one naming nitpick about the efi package actually

internal/efitest/vars.go
chrisccoulson: I've used this naming convention in a few packages - I did try adjusting it to `WriteTo` in go-efilib at one point and IIRC it triggered go vet failures because it checks that functions called `WriteTo` and which take a `io.Writer` exactly match the `io.WriterTo` interface, which would mean adjusting them all to return `(int64, error)` as well.

efi/embeds_test.go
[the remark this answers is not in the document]
chrisccoulson: The main issue is that some of them eventually end up being used in quite a few files.

PR #254, opened 2023-06-08
efi: Add imageLoadHandler implementation for measuring firmware events

chrisccoulson:
This adds an implementation of imageLoadHandler for measuring firmware
events, with support for both the boot manager code (PCR4) and secure
boot policy (PCR7) profiles.


efi/fw_load_handler.go
```diff
@@ -20,9 +20,251 @@
...
+type fwLoadHandler struct {
+	log *tcglog.Log
+}
+
 var newFwLoadHandler = func(log *tcglog.Log) imageLoadHandler {
-	panic("not implemented")
+	return &fwLoadHandler{log: log}
+}
+
+func (h *fwLoadHandler) measureSignatureDb(ctx pcrBranchContext, name efi.VariableDescriptor) ([]byte, error) {
```
pedronis, line 47: does this need a slightly different name as it also returns the db, most measure* methods return only errors, but some return more

efi/fw_load_handler.go
```diff
@@ -20,9 +20,251 @@
...
+	data, err := h.measureSignatureDb(ctx, Db)
+	if err != nil {
+		return err
+	}
+
+	db, err := efi.ReadSignatureDatabase(bytes.NewReader(data))
+	if err != nil {
+		return xerrors.Errorf("cannot decode signatures: %w", err)
+	}
+	ctx.FwContext().Db = &secureBootDB{Name: Db, Contents: db}
```
pedronis, line 67: would this kind of code be clearer as some kind of method on fwContext? also I notice that no method of fwContext actually uses Db, will the Db from the fwContext be used by later stages beyond the fw load handling ?

efi/fw_load_handler.go
```diff
@@ -20,9 +20,251 @@
...
+			if !foundSecureBootSeparator {
+				return errors.New("unexpected verification event")
+			}
+			digest := tpm2.Digest(e.Digests[ctx.PCRAlg()])
+			ctx.FwContext().AppendVerificationEvent(digest)
+			ctx.ExtendPCR(secureBootPCR, digest)
+		case e.PCRIndex == secureBootPCR && e.EventType == tcglog.EventTypeEFIVariableDriverConfig:
+			// ignore: part of the secure boot configuration
+		case e.PCRIndex == secureBootPCR:
+			return fmt.Errorf("unexpected event type (%v) found in log", e.EventType)
```
pedronis, line 125: should the confusion=>error checks here be tested?

pedronis (commented):
did a first pass, some comments/questions

efi/fw_load_handler.go
chrisccoulson: `fwContext.Db` is used in this PR by `fwImageLoadMeasurer.measureVerification`, but it will also be used by the next PR as well.

efi/fw_load_handler.go
chrisccoulson: I've renamed this to `readAndMeasureSignatureDb`

efi/fw_load_handler.go
chrisccoulson: I've added some extra tests here.

efi/fw_load_handler.go
pedronis: I'll try to keep an eye on Db usage in the next PRs, to see if we can have the fwContext have a more structured surface

pedronis (approved):
thanks

PR #255, opened 2023-06-08
efi: Add imageLoadHandler implementations for shim and grub

chrisccoulson:
This adds implementations of imageLoadHandler for shim and grub, with
support for both the boot manager code (PCR4) and secure boot policy
(PCR7) profiles.


efi/shim_load_handler.go
```diff
@@ -0,0 +1,341 @@
...
+	c.version = &ver
+	return c
+}
+
+func (c *shimLoadHandlerConstructor) WithSbatLevel(level shimSbatLevel) *shimLoadHandlerConstructor {
+	c.sbatLevel = &level
+	return c
+}
+
+func (c *shimLoadHandlerConstructor) New(rules imageRuleSet, image peImageHandle) (imageLoadHandler, error) {
```
pedronis, line 61: is this meant to implement newImageLoadHandlerFn ?

efi/image_rules.go
```diff
@@ -0,0 +1,33 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package efi
+
+import "fmt"
+
+type (
+	newImageLoadHandlerFn func(imageRuleSet, peImageHandle) (imageLoadHandler, error)
```
pedronis, line 25: this type is not really used yet except as a return type, will there be something that consumes it?

efi/shim_load_handler.go
```diff
@@ -0,0 +1,341 @@
...
+
+func newShimLoadHandler(rules imageRuleSet, image peImageHandle) (imageLoadHandler, error) {
+	return newShimLoadHandlerConstructor().New(rules, image)
+}
+
+// MeasureImageStart implements imageLoadHandler.MeasureImageStart.
+func (h *shimLoadHandler) MeasureImageStart(ctx pcrBranchContext) error {
+	// Update the context for this branch
+	ctx.ShimContext().Flags = h.Flags
+	ctx.ShimContext().VendorDb = h.VendorDb
```
pedronis, line 179: this is a bit strange, as I remarked the same for `ctx.FwContext().Db = &secureBootDB{Name: Db, Contents: db}` in fw_load_handler.go. I think the pcrBranchContext should start without ShimContext and FwContext and have way to set a full one of those, or the context need some sort of Init or similar function that take the values they need to be initialized all together instead of set one by one.  It's also unclear whether VendorDB should be public on shimContext or a method that panics more cleanly if the context/vendorDb has not been initialised yet..

efi/shim_load_handler.go
```diff
@@ -0,0 +1,341 @@
...
+			}
+
+			cert, err := x509.ParseCertificate(esl.Signatures[0].Data)
+			if err != nil {
+				return nil, xerrors.Errorf("cannot parse vendor cert: %w", err)
+			}
+			certs = append(certs, cert)
+		}
+		authoritySet.AddAuthorities(certs...)
+	}
```
pedronis, line 160: it's a bit unusual for a pure new/constructor function to take an argument to have only side-effects of it. Maybe the "new" functions taking rules should have naming conventions like newOnRules... or similar.

efi/image_rules.go
pedronis: and maybe as remarked its name need to be new...OnRules?

pedronis (commented):
looks reasonable, but a bit confused by the complexity of some things as this continues to introduce quite a few interfaces that are not consumed or respectively only implemented by mock test code. Some style/clarity remarks and questions

efi/shim_load_handler.go
pedronis: or the updating/processing rules part should be a separate method called after construction?

efi/shim_load_handler.go
chrisccoulson: It is.

efi/image_rules.go
chrisccoulson: > this type is not really used yet except as a return type, will there be something that consumes it?

It will be used in the next couple of PRs.

chrisccoulson: I've changed this PR a bit to remove this type - it will get added back in a subsequent PR, although it will look a bit different.

efi/shim_load_handler.go
chrisccoulson: I've refactored this so that it will look a bit different in subsequent PRs.

efi/shim_load_handler.go
```diff
@@ -0,0 +1,338 @@
...
+			Contents: vendorDb,
+		},
+		SbatLevel: sbatLevel}, nil
+}
+
+func newShimLoadHandler(image peImageHandle) (imageLoadHandler, error) {
+	return newShimLoadHandlerConstructor().New(image)
+}
+
+func (h *shimLoadHandler) VendorAuthorities() ([]*x509.Certificate, error) {
```
pedronis, line 153: will this now be called in code coming in the next PRs?

efi/secureboot_namespace_rules.go
```diff
@@ -0,0 +1,26 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package efi
+
+import "crypto/x509"
+
+type vendorAuthorityGetter interface {
+	VendorAuthorities() ([]*x509.Certificate, error)
```
pedronis, line 25: this probably merits doc comments

efi/shim_load_handler.go
pedronis: this should refer to the interface it implements in a doc comment

pedronis (approved):
lgtm, couple small remarks

PR #256, opened 2023-06-09
efi: Add primitives to specify rules for constructing imageLoadHandlers

chrisccoulson:
This is draft until https://github.com/snapcore/secboot/pull/255 has landed.

This adds some primitives for building rules which specify how to
construct an imageLoadHandler for an EFI image. Note that this doesn't
actually implement any of those rules yet - that will be in the next PR.

This is implemented in a way that allows some rules to be scoped to a
secure boot namespace. Some examples of where this is useful is that
there are some versions of shim where we can't inspect the built-in
SBAT payloads, but we can infer what they are for specific versions of
shim becasue we know exactly what was published under the Microsoft UEFI
CA namespace. Also, we have a hard requirement on shim >= 15.2, but we
know that Canonical had 2 builds of shim 15 signed by Microsoft (the
original versions shipped with UC20) that included the required patches
from 15.2 to make it work, so we can add quirks for those.

This is all intended to make it easy to adapt to future versions of boot
components that introduce breaking changes or temporary bugs without
having to rewrite lots of code.


efi/secureboot_namespace_rules.go
```diff
@@ -19,8 +19,120 @@
...
+	"bytes"
+	"crypto/x509"
+
+	"golang.org/x/xerrors"
+)
 
 type vendorAuthorityGetter interface {
 	VendorAuthorities() ([]*x509.Certificate, error)
 }
+
```
pedronis, line 32: this from previous PRs need doc comments

efi/secureboot_namespace_rules.go
```diff
@@ -19,8 +19,120 @@
...
+				continue
+			}
+
+			handler, err := r.imageRules.NewImageLoadHandler(image)
+			if err != nil {
+				return nil, err
+			}
+
+			if v, ok := handler.(vendorAuthorityGetter); ok {
+				certs, err := v.VendorAuthorities()
```
pedronis, line 126: are authorities coming from a handler from a rule a subset of the authorities in r.authorities? wouldn't it make more sense the always add rules with some authorities together instead of separately? I'm looking at the separate with* helpers above

efi/secureboot_namespace_rules_test.go
```diff
@@ -0,0 +1,209 @@
...
+	. "github.com/snapcore/secboot/efi"
+	"github.com/snapcore/secboot/internal/efitest"
+	"github.com/snapcore/secboot/internal/testutil"
+)
+
+type secureBootNamespaceRulesSuite struct{}
+
+var _ = Suite(&secureBootNamespaceRulesSuite{})
+
+func (s *secureBootNamespaceRulesSuite) TestRulesMatch1(c *C) {
```
pedronis, line 37: this Rules# tests might need some comments to explain what is tested in each, just skimming them is hard to spot the differences between each

efi/image_rules.go
```diff
@@ -0,0 +1,243 @@
...
+func (r *imageRules) String() string {
+	return r.name + " image rules"
+}
+
+// NewImageLoadHandler implements imageLoadHandlerConstructor
+func (r *imageRules) NewImageLoadHandler(image peImageHandle) (imageLoadHandler, error) {
+	for _, rule := range r.rules {
+		matches, err := rule.match.Matches(image)
+		if err != nil {
+			return nil, xerrors.Errorf("cannot run \"%s\" rule: %w", rule.name, err)
```
pedronis, line 113: should s/rule/image rule/ ? or will it be clear somehow from context?

efi/image_rules.go
```diff
@@ -0,0 +1,243 @@
...
+		}
+		if matches {
+			handler, err := rule.create(image)
+			if err != nil {
+				return nil, xerrors.Errorf("cannot create using \"%s\" rule: %w", rule.name, err)
```
pedronis, line 118: same question

efi/image_rules.go
```diff
@@ -0,0 +1,243 @@
...
+	case "==":
+		return res == 0, nil
+	case "!=":
+		return res != 0, nil
+	case "<=":
+		return res <= 0, nil
+	case "<":
+		return res < 0, nil
+	default:
+		return false, fmt.Errorf("invalid operator %s", p.operator)
```
pedronis, line 241: s/operator/shim version operator/ ?

pedronis (commented):
did a pass, some questions

efi/secureboot_namespace_rules.go
chrisccoulson: I've added a comment to this now.

efi/secureboot_namespace_rules_test.go
chrisccoulson: I've added some comments here now.

efi/image_rules.go
chrisccoulson: I've updated this.

efi/image_rules.go
chrisccoulson: I've updated this as suggested.

efi/secureboot_namespace_rules.go
chrisccoulson: I'm not sure I understand the question. The comment for vendorAuthorityGetter might help a bit, and it might be useful to look at how this is put together in the next PR as well:

https://github.com/snapcore/secboot/blob/e4c2c15fb92e8673e28d68343fcc59d868e08172/efi/image_rules_defs.go

pedronis (approved):
+1

PR #258, opened 2023-09-21
go.{mod,sum}: move to use native go modules

mvo5:
Instead of using govendor, move to the build-in standard go modules. This also needs to change the workflow.

Note that I did not enable the the 1.21.1 go tests because they fail in https://github.com/snapcore/secboot/actions/runs/6258644709/job/16993049291?pr=258 - the failures actually do not make too much sense to me so I wonder if it's something about GH instead of the tests but definitely worth checking at some point. But going with 1.18 for now means we are on par with what the current tests are doing.


chrisccoulson:
Without looking in detail, the stable failures are almost certainly because go 1.20 changed how `ecdsa.GenerateKey` works - it generates a different key for the same random input now. As secboot expects to be able to recreate the same key for a given sequence of input bytes, we'll need to fix this by reimplementing the original FIPS186-4 generation which is fairly simple to do.

chrisccoulson (approved):

PR #259, opened 2023-09-26
tpm2: implement deterministic ECDSA key generation

chrisccoulson:
go 1.20 changed the behaviour of ecdsa.GenerateKey so that it no longer produces the same key given the same sequence of random bytes, and so that it is no longer even deterministic.

Work around this by reimplementing the ECDSA key generation based on FIPS186-4 section B.4.1.

### DIFF
--- a/internal/crypto/ecdsa.go
+++ b/internal/crypto/ecdsa.go
@@ -29,7 +29,12 @@ import (
 var one = new(big.Int).SetInt64(1)
 
 // GenerateECDSAKey generates a new elliptic key pair using the method described by
-// FIPS186-4 section B.4.1.
+// FIPS186-4 section B.4.1. This method is deterministic (given the same sequence of
+// random bytes, it will generate the same key) and is not sensitive to changes in the
+// standard library between go releases. This is required because the tpm2 package needs
+// to be able to deterministically derive keys from a sequence of bytes. The method
+// used to generate keys using crypto/ecdsa package changed in go1.20 to one that is
+// non-deterministic.
 func GenerateECDSAKey(curve elliptic.Curve, rand io.Reader) (*ecdsa.PrivateKey, error) {
 	params := curve.Params()
 
--- a/internal/crypto/ecdsa.go
+++ b/internal/crypto/ecdsa.go
@@ -1,0 +1,66 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2021 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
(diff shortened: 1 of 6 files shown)


internal/crypto/ecdsa.go
```diff
@@ -0,0 +1,66 @@
...
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"io"
+	"math/big"
+)
+
+var one = new(big.Int).SetInt64(1)
+
+// GenerateECDSAKey generates a new elliptic key pair using the method described by
+// FIPS186-4 section B.4.1.
```
mvo5, line 32: This should probably describe why it re-implements the standard lib crypto

internal/crypto/ecdsa.go
```diff
@@ -0,0 +1,66 @@
...
+func GenerateECDSAKey(curve elliptic.Curve, rand io.Reader) (*ecdsa.PrivateKey, error) {
```
mvo5, line 33: I wonder if this should get it's own dedicated unit test? Like deterministically generating a key and then doing a basic smoke test on it? But maybe that is already part of the tpm testsuite so unneeded(?).

mvo5 (commented):
This looks fine, I compared the code that generates the key with the old code in go that generates the curve and it matches.

internal/crypto/ecdsa.go
```diff
@@ -0,0 +1,71 @@
...
+	params := curve.Params()
+
+	// 1. N=len(n)
+	N := params.N.BitLen() / 8
+
+	// 4. Obtain a string of N+64 bits from an RBG
+	//
+	//  For P-521, this is N+63 bits because of the rounding error, but the original
+	//  crypto/ecdsa had this quirk as well and this function has to be compabible with
```
sespiros, line 47: s/compabible/compatible/ :)

internal/crypto/ecdsa_test.go
```diff
@@ -0,0 +1,66 @@
...
+type ecdsaSuite struct{}
+
+var _ = Suite(&ecdsaSuite{})
+
+func (s *ecdsaSuite) TestGenerateECDSAKey(c *C) {
+	rand := testutil.DecodeHexString(c, "26be875535d98e705cbd60c34f068985fc808e0b83ad1c8ac467ec8294b622a39657d3b9207ba865")
+
+	key, err := GenerateECDSAKey(elliptic.P256(), bytes.NewReader(rand))
+	c.Assert(err, IsNil)
+	c.Check(key.D.Bytes(), DeepEquals, testutil.DecodeHexString(c, "9296ef32f26e73c106a834e0415fa32ee2a6dc84d4359431ad4abd21dea061d6"))
```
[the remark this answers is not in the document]
chrisccoulson: There aren't any test vectors - these are just generated randomly and then compared with the existing crypto/ecdsa package

mvo5 (approved):
Thank you!

PR #263, opened 2023-10-05
internal/luks2: avoid using a pipe to send key to cryptsetup

alfonsosanchezbeato:
This approach has been found to be problematic [1] and the code can be simplified by using stdin to pass down existing and new keys to cryptsetup [2]. This aligns secboot code with snapd code.

[1] https://github.com/snapcore/snapd/pull/12935
[2] https://github.com/snapcore/snapd/pull/13258


internal/luks2/cryptsetup.go
```diff
@@ -363,67 +344,40 @@ func AddKey(devicePath string, existingKey, key []byte, options *AddKeyOptions)
...
-		"--key-file", fifoPath}
+		// read existing key from stdin, specifying key size so
+		// cryptsetup knows where the existing key ends and the new key
+		// starts (we are passing both keys via stdin). Otherwise it
+		// would interpret new lines as separator for the keys, while
+		// we actually allow '\n' to be part of the keys.
+		"--key-file", "-",
+		"--keyfile-size", strconv.Itoa(len(existingKey)),
+		// remove warnings and confirmation questions
+		"--batch-mode"}
```
chrisccoulson, line 360: Just a minor nit, but this file also uses the short form `-q` option in `Format` - they should probably be consistent.

chrisccoulson (approved):
Thanks - this looks ok to me

internal/luks2/cryptsetup.go
alfonsosanchezbeato: Good point, I've changed to the long option in `Format`

PR #265, opened 2023-10-23
[1/3] Keydata v3 platform API changes

sespiros:
Original commit: https://github.com/chrisccoulson/secboot/commit/87d9aa34d1d80d64ffe82f0a008693799d633cd9

I will split the commit above in 3 separate cascading PRs and add tests for it:

1. This one
2. https://github.com/snapcore/secboot/pull/270
3. https://github.com/snapcore/secboot/pull/271

### DIFF
--- a/crypt_test.go
+++ b/crypt_test.go
@@ -45,12 +45,6 @@ import (
 
 type cryptTestBase struct{}
 
-func (ctb *cryptTestBase) newPrimaryKey() []byte {
-	key := make([]byte, 32)
-	rand.Read(key)
-	return key
-}
-
 func (ctb *cryptTestBase) newRecoveryKey() RecoveryKey {
 	var key RecoveryKey
 	rand.Read(key[:])
@@ -469,10 +463,10 @@ func (s *cryptSuite) checkKeyDataKeysInKeyring(c *C, prefix, path string, expect
 	c.Check(auxKey, DeepEquals, expectedAuxKey)
 }
 
-func (s *cryptSuite) newMultipleNamedKeyData(c *C, names ...string) (keyData []*KeyData, keys []DiskUnlockKey, auxKeys []PrimaryKey) {
+func (s *cryptSuite) newMultipleNamedKeyData(c *C, names ...string) (keyData []*KeyData, keys []DiskUnlockKey, primaryKeys []PrimaryKey) {
 	for _, name := range names {
-		key, auxKey := s.newKeyDataKeys(c, 32, 32)
-		protected := s.mockProtectKeys(c, key, auxKey, crypto.SHA256)
+		primaryKey := s.newPrimaryKey(c, 32)
+		protected, unlockKey := s.mockProtectKeys(c, primaryKey, crypto.SHA256)
 
 		kd, err := NewKeyData(protected)
 		c.Assert(err, IsNil)
@@ -485,16 +479,45 @@ func (s *cryptSuite) newMultipleNamedKeyData(c *C, names ...string) (keyData []*
 		c.Assert(err, IsNil)
 
 		keyData = append(keyData, kd)
-		keys = append(keys, key)
-		auxKeys = append(auxKeys, auxKey)
+		keys = append(keys, unlockKey)
+		primaryKeys = append(primaryKeys, primaryKey)
 	}
 
-	return keyData, keys, auxKeys
+	return keyData, keys, primaryKeys
 }
 
 func (s *cryptSuite) newNamedKeyData(c *C, name string) (*KeyData, DiskUnlockKey, PrimaryKey) {
-	keyData, keys, auxKeys := s.newMultipleNamedKeyData(c, name)
-	return keyData[0], keys[0], auxKeys[0]
+	keyData, unlockKeys, primaryKeys := s.newMultipleNamedKeyData(c, name)
+	return keyData[0], unlockKeys[0], primaryKeys[0]
+}
+
+func (s *cryptSuite) newMultipleNamedKeyDataWithPassphrases(c *C, passphrases []string, kdf KDF, names ...string) (keyData []*KeyData, keys []DiskUnlockKey, primaryKeys []PrimaryKey) {
+	for i, name := range names {
+		primaryKey := s.newPrimaryKey(c, 32)
+		protected, unlockKey := s.mockProtectKeysWithPassphrase(c, primaryKey, kdf, nil, 32, crypto.SHA256)
+
+		kd, err := NewKeyDataWithPassphrase(protected, passphrases[i], kdf)
+		c.Assert(err, IsNil)
+
+		w := makeMockKeyDataWriter()
+		c.Check(kd.WriteAtomic(w), IsNil)
+
+		r := &mockKeyDataReader{name, w.Reader()}
+		kd, err = ReadKeyData(r)
+		c.Assert(err, IsNil)
+
+		keyData = append(keyData, kd)
+		keys = append(keys, unlockKey)
+		primaryKeys = append(primaryKeys, primaryKey)
+	}
+
+	return keyData, keys, primaryKeys
+}
+
+func (s *cryptSuite) newNamedKeyDataWithPassphrase(c *C, passphrase string, kdf KDF, name string) (*KeyData, DiskUnlockKey, PrimaryKey) {
+	passphrases := []string{passphrase}
+	keyData, unlockKeys, primaryKeys := s.newMultipleNamedKeyDataWithPassphrases(c, passphrases, kdf, name)
+	return keyData[0], unlockKeys[0], primaryKeys[0]
 }
 
 type testActivateVolumeWithRecoveryKeyData struct {
@@ -804,17 +827,20 @@ type testActivateVolumeWithKeyDataData struct {
 
 func (s *cryptSuite) testActivateVolumeWithKeyData(c *C, data *testActivateVolumeWithKeyDataData) {
 	var err error
-	keyData, key, auxKey := s.newNamedKeyData(c, "")
-	slot := s.addMockKeyslot(data.sourceDevicePath, key)
-
-	c.Check(keyData.SetAuthorizedSnapModels(auxKey, data.authorizedModels...), IsNil)
-
-	authRequestor := &mockAuthRequestor{passphraseResponses: data.authResponses}
-
+	var unlockKey DiskUnlockKey
+	var primaryKey PrimaryKey
+	var keyData *KeyData
 	var kdf testutil.MockKDF
 	if data.passphrase != "" {
-		c.Check(keyData.SetPassphrase(data.passphrase, nil, &kdf), IsNil)
+		keyData, unlockKey, primaryKey = s.newNamedKeyDataWithPassphrase(c, data.passphrase, &kdf, "")
+	} else {
+		keyData, unlockKey, primaryKey = s.newNamedKeyData(c, "")
 	}
+	slot := s.addMockKeyslot(data.sourceDevicePath, unlockKey)
+
+	c.Check(keyData.SetAuthorizedSnapModels(primaryKey, data.authorizedModels...), IsNil)
+
+	authRequestor := &mockAuthRequestor{passphraseResponses: data.authResponses}
 
 	options := &ActivateVolumeOptions{
 		PassphraseTries: data.passphraseTries,
@@ -854,7 +880,7 @@ func (s *cryptSuite) testActivateVolumeWithKeyData(c *C, data *testActivateVolum
 	}
 
 	// This should be done last because it may fail in some circumstances.
-	s.checkKeyDataKeysInKeyring(c, data.keyringPrefix, data.sourceDevicePath, key, auxKey)
+	s.checkKeyDataKeysInKeyring(c, data.keyringPrefix, data.sourceDevicePath, unlockKey, primaryKey)
 }
 
 func (s *cryptSuite) TestActivateVolumeWithKeyData1(c *C) {
@@ -1259,11 +1285,9 @@ func (s *cryptSuite) TestActivateVolumeWithKeyDataErrorHandling9(c *C) {
 
 func (s *cryptSuite) TestActivateVolumeWithKeyDataErrorHandling10(c *C) {
 	// Test that recovery key fallback works if the wrong passphrase is supplied.
-	keyData, key, _ := s.newNamedKeyData(c, "foo")
-	recoveryKey := s.newRecoveryKey()
-
 	var kdf testutil.MockKDF
-	c.Check(keyData.SetPassphrase("1234", nil, &kdf), IsNil)
+	keyData, key, _ := s.newNamedKeyDataWithPassphrase(c, "1234", &kdf, "foo")
+	recoveryKey := s.newRecoveryKey()
 
 	s.testActivateVolumeWithKeyDataErrorHandling(c, &testActivateVolumeWithKeyDataErrorHandlingData{
 		primaryKey:  key,
@@ -1306,11 +1330,9 @@ func (s *cryptSuite) TestActivateVolumeWithKeyDataErrorHandling12(c *C) {
 
 func (s *cryptSuite) TestActivateVolumeWithKeyDataErrorHandling13(c *C) {
 	// Test that activation fails if the supplied passphrase and recovery key are incorrect
-	keyData, key, _ := s.newNamedKeyData(c, "bar")
-	recoveryKey := s.newRecoveryKey()
-
 	var kdf testutil.MockKDF
-	c.Check(keyData.SetPassphrase("1234", nil, &kdf), IsNil)
+	keyData, key, _ := s.newNamedKeyDataWithPassphrase(c, "1234", &kdf, "bar")
+	recoveryKey := s.newRecoveryKey()
 
 	c.Check(s.testActivateVolumeWithKeyDataErrorHandling(c, &testActivateVolumeWithKeyDataErrorHandlingData{
 		primaryKey:  key,
@@ -1364,11 +1386,9 @@ func (s *cryptSuite) TestActivateVolumeWithKeyDataErrorHandling15(c *C) {
 
 func (s *cryptSuite) TestActivateVolumeWithKeyDataErrorHandling16(c *C) {
 	// Test that error in authRequestor error surfaces
-	keyData, key, _ := s.newNamedKeyData(c, "bar")
-	recoveryKey := s.newRecoveryKey()
-
 	var kdf testutil.MockKDF
-	c.Check(keyData.SetPassphrase("1234", nil, &kdf), IsNil)
+	keyData, key, _ := s.newNamedKeyDataWithPassphrase(c, "1234", &kdf, "bar")
+	recoveryKey := s.newRecoveryKey()
 
 	c.Check(s.testActivateVolumeWithKeyDataErrorHandling(c, &testActivateVolumeWithKeyDataErrorHandlingData{
 		primaryKey:  key,
@@ -1513,11 +1533,9 @@ func (s *cryptSuite) TestActivateVolumeWithMultipleKeyData3(c *C) {
 
 func (s *cryptSuite) TestActivateVolumeWithMultipleKeyData4(c *C) {
 	// Test with 2 keys that have a passphrase set, using the first key for activation.
-	keyData, keys, auxKeys := s.newMultipleNamedKeyData(c, "", "")
-
 	var kdf testutil.MockKDF
-	c.Check(keyData[0].SetPassphrase("1234", nil, &kdf), IsNil)
-	c.Check(keyData[1].SetPassphrase("5678", nil, &kdf), IsNil)
+	passphrases := []string{"1234", "5678"}
+	keyData, keys, auxKeys := s.newMultipleNamedKeyDataWithPassphrases(c, passphrases, &kdf, "", "")
 
 	models := []SnapModel{
 		testutil.MakeMockCore20ModelAssertion(c, map[string]interface{}{
@@ -1545,11 +1563,9 @@ func (s *cryptSuite) TestActivateVolumeWithMultipleKeyData4(c *C) {
 
 func (s *cryptSuite) TestActivateVolumeWithMultipleKeyData5(c *C) {
 	// Test with 2 keys that have a passphrase set, using the second key for activation.
-	keyData, keys, auxKeys := s.newMultipleNamedKeyData(c, "", "")
-
 	var kdf testutil.MockKDF
-	c.Check(keyData[0].SetPassphrase("1234", nil, &kdf), IsNil)
-	c.Check(keyData[1].SetPassphrase("5678", nil, &kdf), IsNil)
+	passphrases := []string{"1234", "5678"}
+	keyData, keys, auxKeys := s.newMultipleNamedKeyDataWithPassphrases(c, passphrases, &kdf, "", "")
 
 	models := []SnapModel{
 		testutil.MakeMockCore20ModelAssertion(c, map[string]interface{}{
@@ -1578,10 +1594,13 @@ func (s *cryptSuite) TestActivateVolumeWithMultipleKeyData5(c *C) {
 func (s *cryptSuite) TestActivateVolumeWithMultipleKeyData6(c *C) {
 	// Test with 2 keys where one has a passphrase set. The one without the passphrase
 	// should be used first.
-	keyData, keys, auxKeys := s.newMultipleNamedKeyData(c, "", "")
-
 	var kdf testutil.MockKDF
-	c.Check(keyData[0].SetPassphrase("1234", nil, &kdf), IsNil)
+	keyData1, unlockKey1, primaryKey1 := s.newNamedKeyDataWithPassphrase(c, "1234", &kdf, "")
+	keyData2, unlockKey2, primaryKey2 := s.newNamedKeyData(c, "")
+
+	keyData := []*KeyData{keyData1, keyData2}
+	unlockKeys := []DiskUnlockKey{unlockKey1, unlockKey2}
+	primaryKeys := []PrimaryKey{primaryKey1, primaryKey2}
 
 	models := []SnapModel{
 		testutil.MakeMockCore20ModelAssertion(c, map[string]interface{}{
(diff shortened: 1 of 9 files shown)


crypt_test.go
```diff
@@ -2524,7 +2547,7 @@ func (s *cryptSuite) TestInitializeLUKS2ContainerWithCustomKDFIterations(c *C) {
 }
 
 func (s *cryptSuite) TestInitializeLUKS2ContainerInvalidKeySize(c *C) {
-	c.Check(InitializeLUKS2Container("/dev/sda1", "data", s.newPrimaryKey()[0:16], nil), ErrorMatches, "expected a key length of at least 256-bits \\(got 128\\)")
+	c.Check(InitializeLUKS2Container("/dev/sda1", "data", ([]byte)(s.keyDataTestBase.newPrimaryKey(c, 32))[0:16], nil), ErrorMatches, "expected a key length of at least 256-bits \\(got 128\\)")
```
chrisccoulson, line 2550: As `keyDataTestBase` takes a size now, you don't need to create a 32-byte key here and then truncate it to 16 bytes.

Also, is the type conversion to `[]byte` necessary here? That should happen implicitly because `[]byte` is not a named type. This is repeated in a few other places too.

For an idea of when explicit conversions are necessary when converting between types, take a look at this section of the language spec: [Assignability](https://go.dev/ref/spec#Assignability) (in this case, "V and T have identical [underlying types](https://go.dev/ref/spec#Underlying_types) but are not type parameters and at least one of V or T is not a [named type](https://go.dev/ref/spec#Types)).

keydata_legacy_test.go
```diff
@@ -0,0 +1,334 @@
+package secboot
```
chrisccoulson, line 1: This should be in package `secboot_test`. I think moving this will create a naming conflict though because there are 2 `keyDataTestBase` types.

keydata_test.go
```diff
@@ -244,28 +248,35 @@ func (s *keyDataTestBase) TearDownSuite(c *C) {
...
-	payload := MarshalKeys(key, auxKey)
+func (s *keyDataTestBase) mockProtectKeys(c *C, primaryKey PrimaryKey, modelAuthHash crypto.Hash) (out *KeyParams, unlockKey DiskUnlockKey) {
+	unique := make([]byte, len(primaryKey))
+	_, err := rand.Read(unique)
+	c.Assert(err, IsNil)
+
+	reader := new(bytes.Buffer)
+	reader.Write(unique)
+
+	unlockKey, payload, err := MakeDiskUnlockKey(reader, crypto.SHA256, primaryKey)
```
chrisccoulson, line 267: You should just be able to pass `bytes.NewReader(unique)` as the first argument here.

keydata_test.go
```diff
@@ -276,10 +287,31 @@ func (s *keyDataTestBase) mockProtectKeys(c *C, key DiskUnlockKey, auxKey Auxili
...
+
+func (s *keyDataTestBase) mockProtectKeysWithPassphrase(c *C, primaryKey PrimaryKey, kdf KDF, kdfOptions *KDFOptions, authKeySize int, modelAuthHash crypto.Hash) (out *KeyWithPassphraseParams, unlockKey DiskUnlockKey) {
+	kp, unlockKey := s.mockProtectKeys(c, primaryKey, modelAuthHash)
+
+	if kdfOptions == nil {
+		var defaultOptions KDFOptions
+		kdfOptions = &defaultOptions
+	}
+
+	kdfOptions.DeriveCostParams(len(primaryKey)+16, kdf)
```
chrisccoulson, line 306: Is this call necessary?

keydata_test.go
```diff
@@ -350,7 +382,7 @@ func (s *keyDataTestBase) checkKeyDataJSONDecodedAuthModeNone(c *C, j map[string
 	c.Check(err, IsNil)
 	c.Check(encryptedPayload, DeepEquals, creationParams.EncryptedPayload)
 
-	c.Check(j, Not(testutil.HasKey), "passphrase_protected_payload")
+	c.Check(j, Not(testutil.HasKey), "passphrase_params")
```
chrisccoulson, line 385: I think this is missing some checks of new fields (`version` and `kdf_alg`).

keydata_test.go
```diff
@@ -362,30 +394,43 @@ func (s *keyDataTestBase) checkKeyDataJSONFromReaderAuthModeNone(c *C, r io.Read
...
-func (s *keyDataTestBase) checkKeyDataJSONDecodedAuthModePassphrase(c *C, j map[string]interface{}, creationParams *KeyParams, nmodels int, passphrase string, kdfOpts *KDFOptions) {
+func (s *keyDataTestBase) checkKeyDataJSONDecodedAuthModePassphrase(c *C, j map[string]interface{}, creationParams *KeyWithPassphraseParams, nmodels int, passphrase string, kdfOpts *KDFOptions) {
 	if kdfOpts == nil {
 		var def KDFOptions
 		kdfOpts = &def
 	}
 	var kdf testutil.MockKDF
 
-	costParams, err := kdfOpts.DeriveCostParams(0, &kdf)
+	costParams, err := kdfOpts.DeriveCostParams(32+16, &kdf)
```
chrisccoulson, line 404: I passed zero here originally because the value isn't really used by the mock KDF used in tests.

keydata_test.go
```diff
@@ -362,30 +394,43 @@ func (s *keyDataTestBase) checkKeyDataJSONFromReaderAuthModeNone(c *C, r io.Read
...
-	s.checkKeyDataJSONCommon(c, j, creationParams, nmodels)
+	s.checkKeyDataJSONCommon(c, j, &creationParams.KeyParams, nmodels)
+
+	_, ok := j["kdf_alg"].(string)
+	c.Check(ok, testutil.IsTrue)
 
-	c.Check(j, Not(testutil.HasKey), "encrypted_payload")
+	version, ok := j["version"].(float64)
+	c.Check(ok, testutil.IsTrue)
+	c.Check(version, Equals, float64(2))
```
chrisccoulson, line 414: Aha, ignore my earlier comment about these fields not being checked, but they should probably move to `checkKeyDataJSONCommon`.

keydata_test.go
```diff
@@ -448,70 +526,111 @@ func (s *keyDataSuite) checkKeyDataJSONAuthModeNone(c *C, keyData *KeyData, crea
...
-	c.Check(key, DeepEquals, data.key)
-	c.Check(auxKey, DeepEquals, data.auxKey)
+
+	dj := new(bytes.Buffer)
+	json.NewEncoder(dj).Encode(data)
+
+	pkj := new(bytes.Buffer)
+	json.NewEncoder(pkj).Encode(pk)
+
+	c.Check(dj.Bytes(), DeepEquals, pkj.Bytes())
```
chrisccoulson, line 566: Rather than encoding `protectedKeys` to JSON in order to byte compare them, I would:
- Allow the members of `protectedKeys` to be exported (ie, rename them to `Primary` and `Unique`.
- Add a type alias to export_test.go (`type ProtectedKeys = protectedKeys`)
- Do `c.Check(pk, DeepEquals, &ProtectedKeys{Primary: data.primary, Unique: data.unique})`

keydata_test.go
```diff
@@ -1369,12 +1304,14 @@ func (s *keyDataSuite) TestReadAndWriteWithUnsaltedKeyDigest(c *C) {
 	c.Check(ok, testutil.IsTrue)
 }
 
+// Legacy tests
```
chrisccoulson, line 1307: Should the comment here be above `TestReadAndWriteWithUnsaltedKeyDigest`?

keydata_test.go
```diff
@@ -1322,6 +1256,7 @@ func (s *keyDataSuite) TestReadAndWriteWithUnsaltedKeyDigest(c *C) {
 	auxKey := testutil.DecodeHexString(c, "8107f1c65c58934f0d59245d1d94d312ea803e69c8599a7bac8c67fe253232f2")
 	j := []byte(
 		`{` +
+			`"version":0,` +
```
chrisccoulson, line 1259: As existing keys won't have this field, we should test that reading works for payloads that don't contain it.

platform.go
```diff
@@ -64,16 +66,19 @@ func (e *PlatformHandlerError) Unwrap() error {
 // PlatformKeyData represents the data exchanged between this package and
 // platform implementations via the PlatformKeyDataHandler.
 type PlatformKeyData struct {
-	EncodedHandle    []byte // The JSON encoded platform handle
-	EncryptedPayload []byte // The encrypted payload
+	Version       int
+	EncodedHandle []byte // The JSON encoded platform handle
+	KDFAlg        crypto.Hash
+
+	AuthMode AuthMode
```
chrisccoulson, line 73: There's nothing in the tests inside `mockPlatformKeyDataHandler` that checks we pass the expected values for these new fields.

keydata_legacy_test.go
```diff
@@ -0,0 +1,334 @@
...
+	stream.XORKeyStream(out.EncryptedPayload, payload)
+	return
+}
+
+type keyDataLegacySuite struct {
+	keyDataTestBase
+}
+
+func (s *keyDataLegacySuite) newKeyDataKeys(c *C, sz1, sz2 int) (DiskUnlockKey, PrimaryKey) {
+	key := make([]byte, sz1)
```
chrisccoulson, line 208: It looks like there's 2 implementations of this.

chrisccoulson (changes requested):
Thanks for working on this - I've left some comments inline.

crypt_test.go
```diff
@@ -485,16 +485,45 @@ func (s *cryptSuite) newMultipleNamedKeyData(c *C, names ...string) (keyData []*
...
-	keyData, keys, auxKeys := s.newMultipleNamedKeyData(c, name)
-	return keyData[0], keys[0], auxKeys[0]
+func (s *cryptSuite) newNamedKeyData(c *C, name string) (*KeyData, DiskUnlockKey, PrimaryKey) {
+	keyData, unlockKeys, primaryKeys := s.newMultipleNamedKeyData(c, name)
+	return keyData[0], unlockKeys[0], primaryKeys[0]
+}
+
+func (s *cryptSuite) newMultipleNamedKeyDataWithPassphrases(c *C, passphrases []string, kdf KDF, names ...string) (keyData []*KeyData, keys []DiskUnlockKey, primaryKeys []PrimaryKey) {
+	for i, name := range names {
+		primaryKey := s.keyDataTestBase.newPrimaryKey(c, 32)
```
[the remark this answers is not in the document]
sespiros: It is ambiguous, yes. `keyDataTestBase.newPrimaryKey()` replaced `keyDataTestBase.newKeyDataKeys()` while `cryptTestBase.newPrimaryKey()` is also defined. As a cleanup I added a fixup commit which removes it from `cryptTestBase`. 

`cryptSuiteUnmocked*` also used it from `cryptTestBase` but they don't inherit `keyDataTestBase` so I implemented it just for `cryptSuiteUnmockedBase` instead.

Any better cleanup idea here is welcome :)

crypt_test.go
sespiros: Removed the truncation, thanks. 

Regarding the type conversion, keyDataTestBase`s `newPrimaryKey` returns type `PrimaryKey` while all the LUKS related APIs expect type `DiskUnlockKey` so I believe this doesn't fall under the categories above so the type conversion is needed despite them having the same underlying type `[]byte`.

keydata_legacy_test.go
sespiros: I refactored the legacy keyData test suite.

keydata_test.go
sespiros: Thanks, fixed in 2 places (+1 in a later PR).

keydata_test.go
sespiros: Not really. Removed.

keydata_test.go
sespiros: aha, reverted.

keydata_test.go
sespiros: Moved.

keydata_test.go
sespiros: Thanks for the recommendation, I fixed it as suggested :) 

I was not sure what was your intention with making them private so I thought not to change them.

keydata_test.go
sespiros: Yes, moved it. Also I see that the `TestReadKeyData*` tests make use of `SetAuthorizedSnapModels` which will now be legacy behavior so I will move it above those tests too in the next PR.

keydata_test.go
sespiros: Correct, I got confused here. So I modified the version behavior to use 0 of 1 for the version field if the `Version` is less than 2 and by marking the Version field as omitempty but I am not sure if that is correct. Can you think of any other legacy scenario that might be weird here?

I also had to modify the tests in `keydata_legacy_test.go` to use 0 for legacy version. Is keyData with version 1 possible?

keydata_legacy_test.go
sespiros: Removed the redundant one from `keyDataLegacySuite`.

platform.go
sespiros: I added some validation for the `mockPlatformKeyDataHandler` fields. I found a missing field in the legacy tests because of it. Let me know how it looks. I hardcoded `crypto.SHA256` in `mockProtectKeys` for the legacy keyData test bases as there aren't any test cases that test for different KDF algorithms.

crypt_test.go
chrisccoulson: Ah, I think the problem here is that perhaps there should have originally been a `newUnlockKey` method for returning a key with the correct type.

keydata.go
```diff
@@ -42,7 +42,7 @@ import (
 )
 
 const (
-	KeyDataVersion int = 2
```
chrisccoulson: It's probably cleaner to just define `KeyDataVersion` as a variable here rather than having a mockable function, and then `MockKeyDataVersion` can just set this instead.

keydata_test.go
chrisccoulson: Thanks, that looks good.

The members are still private by default because the structure isn't exported - all renaming them does here is make them exportable so that they are accessible by adding an exported type alias just for testing. I generally keep members of unexported types unexportable unless it's useful to have direct access to them in unit tests.

keydata_test.go
```diff
@@ -76,7 +76,29 @@ func (h *mockPlatformKeyDataHandler) checkState() error {
...
+		if data.KDFAlg < crypto.MD4 || data.KDFAlg > crypto.BLAKE2b_512 {
+			return xerrors.Errorf("Unexpected KDFAlg: %d", data.KDFAlg)
+		}
+	}
+
+	if data.AuthMode < AuthModeNone || data.AuthMode > AuthModePassphrase {
+		return xerrors.Errorf("Invalid AuthMode: %d", data.AuthMode)
+	}
+
+	return nil
```
chrisccoulson, line 94: The validation should check that the supplied fields are the expected values rather than bounds checking them. I'm not sure of the best way to do this though - maybe adding the expected values as fields to `mockPlatformKeyDataHandle`?

keydata.go
```diff
@@ -482,110 +505,172 @@ func (d *KeyData) snapModelAuthKey(auxKey AuxiliaryKey) ([]byte, error) {
...
+			return nil, nil, &InvalidKeyDataError{xerrors.Errorf("cannot unmarshal cleartext key payload: %w", err)}
+		}
+		return pk.unlockKey(crypto.Hash(d.data.KDFAlg)), pk.Primary, nil
+	default:
+		return nil, nil, fmt.Errorf("invalid keydata version %d", d.Version())
+	}
+}
+
+func (d *KeyData) Version() int {
+	return d.data.Version
```
chrisccoulson, line 673: I think having this field as `omitempty` is fine, but I would probably still covert 0 -> 1 here.

chrisccoulson, line 673: I think having this field as omitempty is fine, but I would probably still covert 0 -> 1 here.

platform.go
chrisccoulson: Thanks, I added a comment for this separately before responding to this thread: https://github.com/snapcore/secboot/pull/265#discussion_r1450624305

keydata_test.go
chrisccoulson: I added a comment separately for this - I think the `omitempty` is fine, but I would still fake version 1 in `KeyData.Version` so that it returns 1 if the field is 0. I'm not sure if it matters either way to be honest, but it's a bit weird going from v0 -> v2.

chrisccoulson:
The test failures are occurring because the callback returned from `MockKeyDataVersion` isn't being called at the end of the test. It should be used something like this:
```
restore := MockKeyDataVersion(0)
defer restore()
```

keydata.go
```diff
@@ -482,110 +505,172 @@ func (d *KeyData) snapModelAuthKey(auxKey AuxiliaryKey) ([]byte, error) {
...
 	}
-	stream := cipher.NewCFBDecrypter(c, key[data.KeySize:])
-	stream.XORKeyStream(payload, data.EncryptedPayload)
+}
 
-	return payload, key, nil
+func (d *KeyData) recoverKeysCommon(data []byte) (DiskUnlockKey, PrimaryKey, error) {
+	switch d.Version() {
+	case 0:
+		unlockKey, primaryKey, err := unmarshalV1KeyPayload(data)
```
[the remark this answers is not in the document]
chrisccoulson: This is related to https://github.com/snapcore/secboot/pull/265#discussion_r1450633800 as well.

keydata_test.go
sespiros: Reverted the last attempt and pushed some changes to the logic if you could take a look. I added `Expected*` fields to the `mockPlatformKeyDataHandle` but tbh I find it a bit confusing to parse.

keydata_legacy_test.go
```diff
@@ -0,0 +1,195 @@
...
+	"crypto/cipher"
+	"crypto/hmac"
+	"hash"
+	"math/rand"
+
+	. "github.com/snapcore/secboot"
+	. "gopkg.in/check.v1"
+)
+
+type keyDataLegacyTestBase struct {
```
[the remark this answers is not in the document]
sespiros: So I had refactored and completely removed that file in the subsequent PR, check https://github.com/snapcore/secboot/pull/270/commits/4cea1c7b4ad07dfc536609ad9646c2938704fa6d

keydata.go
sespiros: I reverted it to the previous implementation which fakes 1 for version 0.

platform.go
```diff
@@ -64,16 +66,19 @@ func (e *PlatformHandlerError) Unwrap() error {
...
 }
 
 // PlatormKeyDataHandler is the interface that this go package uses to
 // interact with a platform's secure device for the purpose of recovering keys.
 type PlatformKeyDataHandler interface {
 	// RecoverKeys attempts to recover the cleartext keys from the supplied key
 	// data using this platform's secure device.
```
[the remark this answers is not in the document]
sespiros: Changed doc comments. @chrisccoulson could you also take a look?

keydata.go
```diff
@@ -34,19 +35,29 @@ import (
...
 	drbg "github.com/canonical/go-sp800.90a-drbg"
 
+	"golang.org/x/crypto/cryptobyte"
+	cryptobyte_asn1 "golang.org/x/crypto/cryptobyte/asn1"
 	"golang.org/x/crypto/hkdf"
 	"golang.org/x/xerrors"
 )
 
 const (
+	keyDataVersion int = 2
```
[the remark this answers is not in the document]
sespiros: I added a slightly more complete doc comment in `KeyData`'s `Version()`. Should we add something here too?

keydata.go
```diff
@@ -932,9 +948,118 @@ func NewKeyData(params *KeyParams) (*KeyData, error) {
...
+}
+
+func (k *protectedKeys) marshalASN1(builder *cryptobyte.Builder) {
+	builder.AddASN1(cryptobyte_asn1.SEQUENCE, func(b *cryptobyte.Builder) { // ProtectedKeys ::= SEQUENCE {
+		b.AddASN1OctetString(k.Primary) // primary OCTETSTRING
+		b.AddASN1OctetString(k.Unique)  // unique OCTETSTRING
+	})
+}
+
+func MakeDiskUnlockKey(rand io.Reader, alg crypto.Hash, primaryKey PrimaryKey) (unlockKey DiskUnlockKey, cleartextPayload []byte, err error) {
```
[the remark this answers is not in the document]
sespiros: Added.

keydata.go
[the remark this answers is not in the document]
sespiros: So I already changed it and replied to the comment above but summarizing it again to make sure I don't miss something. The following versions are possible:
- version 0. Key data already used are missing a version field. `Version()` called in that keydata will return 1 and `unmarshalV1KeyPayload` will be called.
- version 1 shouldn't exist (?), but even then the behavior would be the same as version 0 keys. 
- version 2 is the currently introduced format.

When I first saw V1, I assumed there was another V0. Isn't that the case? If not, maybe we could simply call old keys legacy keys?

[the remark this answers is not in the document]
sespiros: Maybe "generation"? So that we can refer to cross-platform keydata versions as generations and then each platform can refer to its internal version simply as "version". 

i.e TPM platform's key versions 0-2 use cross-platform keys of generation 1 (or 0, this is still a bit confusing) while TPM platform's version 3 keys use generation 2 cross-platform keys.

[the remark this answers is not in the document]
pedronis: "generation" would work for me

keydata.go
```diff
@@ -103,53 +113,16 @@ func (e *PlatformDeviceUnavailableError) Unwrap() error {
 // DiskUnlockKey is the key used to unlock a LUKS volume.
 type DiskUnlockKey []byte
 
-// AuxiliaryKey is an additional key used to modify properties of a KeyData
+// PrimaryKey is an additional key used to modify properties of a KeyData
 // object without having to create a new object.
-type AuxiliaryKey []byte
+type PrimaryKey []byte
```
pedronis, line 118: for future consideration: this kind of rename would have been best done has its own PR

keydata.go
```diff
@@ -932,9 +951,121 @@ func NewKeyData(params *KeyParams) (*KeyData, error) {
...
+	}
+
+	return pk, nil
+}
+
+func (k *protectedKeys) unlockKey(alg crypto.Hash) DiskUnlockKey {
+	if alg == crypto.Hash(nilHash) {
+		// This is to support the legacy TPM key data created
+		// via tpm2.NewKeyDataFromSealedKeyObjectFile.
+		return k.Unique
```
pedronis, line 1024: this path is not exercised

keydata.go
```diff
@@ -932,9 +951,121 @@ func NewKeyData(params *KeyParams) (*KeyData, error) {
...
+	}
+
+	if err := kd.updatePassphrase(kd.data.EncryptedPayload, make([]byte, params.AuthKeySize), passphrase, kdf); err != nil {
+		return nil, xerrors.Errorf("cannot set passphrase: %w", err)
+	}
+
+	return kd, nil
+}
+
+type protectedKeys struct {
```
pedronis, line 997: this probably could use a doc comment

keydata.go
```diff
@@ -932,9 +951,121 @@ func NewKeyData(params *KeyParams) (*KeyData, error) {
 	return kd, nil
 }
 
+func NewKeyDataWithPassphrase(params *KeyWithPassphraseParams, passphrase string, kdf KDF) (*KeyData, error) {
```
pedronis, line 954: needs a doc comment

keydata.go
```diff
@@ -482,110 +504,180 @@ func (d *KeyData) snapModelAuthKey(auxKey AuxiliaryKey) ([]byte, error) {
...
+	}
+	if params.DerivedKeySize < 0 {
+		return nil, nil, nil, fmt.Errorf("invalid derived key size (%d bytes)", params.DerivedKeySize)
+	}
+	if params.EncryptionKeySize < 0 || params.EncryptionKeySize > 32 {
+		// The key size can't be larger than 32 with the supported cipher
+		return nil, nil, nil, fmt.Errorf("invalid encryption key size (%d bytes)", params.EncryptionKeySize)
+	}
+	if params.AuthKeySize < 0 {
+		return nil, nil, nil, fmt.Errorf("invalid auth key size (%d bytes)", params.AuthKeySize)
```
pedronis, line 525: should/can some of these checks be tested?

pedronis (changes requested):
did another pass, a few comments, from a review POV my preference would be for the rename Auxiliary=>Primary to be a separate preparatory PR, it's hard to review the other changes mixed with that mechanical one

keydata.go
pedronis: as I wrote in the overall comment, this actual would be useful

keydata.go
sespiros: I added some tests that use `RecoverKeysWithPassphrase`.

keydata.go
sespiros: Added.

keydata.go
sespiros: Added.

keydata_file_test.go
```diff
@@ -59,12 +59,17 @@ func (s *keyDataFileSuite) TestWriter(c *C) {
 	c.Assert(err, IsNil)
 	defer f.Close()
 
-	s.checkKeyDataJSONFromReaderAuthModeNone(c, f, protected, 0)
+	var j map[string]interface{}
+
+	d := json.NewDecoder(f)
```
pedronis, line 64: need to import json

keydata.go
```diff
@@ -174,6 +148,27 @@ type KeyParams struct {
...
 	SnapModelAuthHash crypto.Hash
+	PlatformName      string // Name of the platform that produced this data
+
+	// KDFAlg is the digest algorithm used to derive additional keys during
+	// the use of the created KeyData. It must match the algorithm passed to
+	// [MakeDiskUnlockKey]. The zero value here has a special meaning which
+	// is reserved to support legacy TPM2 key data files, and tells the
+	// KeyData to use the unique key as the unlock key rather than using it
+	// to derive the unlock key.
+	KDFAlg crypto.Hash
```
pedronis, line 159: I'm confused by the comment about the legacy case, as this used for creating a new KeyData and I don't see any code that special case this KDFAlg when it's null, nor I understand what goes into EncryptedPayload in that case

keydata.go
pedronis: but if Version is 1 we don't arrive here,  as I said I'm confused by the role of KDFAlg vs Version here and the KeyParams, aynway if this is useful there should be test in keydata_test.go that shows how this is meant to happen

pedronis (commented):
thanks, I'm confused by the role of KDFAlg nilHash in a couple of places, maybe it's really a question for Chris

keydata.go
sespiros: So I gave it a second look and my guess would be that this is probably forgotten comment/code (maybe from an older implementation?), since we use the `version` field to differentiate the keys (which now is also marked as `omitempty`). I removed the unreachable block and the existing tests pass (as expected).

keydata.go
chrisccoulson: This is exercised by the tests in the `tpm2` package - specifically, the ones in `tpm2/platform_legacy_test.go` that call `NewKeyDataFromSealedKeyObjectFile`. This creates an ephemeral v2 key data but with 0 for the `KDFAlg` parameter. This path exists just for that case.

keydata.go
pedronis: those are files without the json wrapper?

pedronis: Thanks. I would mention tpm2.NewKeyDataFromSealedKeyObjectFile in the comment for clarity.

keydata.go
```diff
@@ -482,110 +504,180 @@ func (d *KeyData) snapModelAuthKey(auxKey PrimaryKey) ([]byte, error) {
...
+
+	iv = make([]byte, aes.BlockSize)
+	r = hkdf.Expand(func() hash.Hash { return kdfAlg.New() }, derived, []byte("PASSPHRASE-IV"))
+	if _, err := io.ReadFull(r, iv); err != nil {
+		return nil, nil, nil, xerrors.Errorf("cannot derive IV: %w", err)
 	}
 
-	key, err := kdf.Derive(passphrase, salt[:], params, uint32(keyLen))
+	auth = make([]byte, params.AuthKeySize)
+	r = hkdf.Expand(func() hash.Hash { return kdfAlg.New() }, derived, []byte("PASSPHRASE-AUTH"))
```
pedronis, line 574: if I change the constant  string no tests seem to fail, I wonder if and how this bit of the code is tested

pedronis (commented):
this looks reasonable to me, question about testing though. The question about what to do with "version" in the JSON is still open though

keydata.go
sespiros: I added a test case for that for good measure. The test hardcodes a keydata generated with the current valid info fields.

keydata_test.go
chrisccoulson: I think this is ok. One other way to implement this would be to make `mockKeyDataHandler` use GCM and have the extra properties as part of the AAD instead, but maybe we can change that in a follow-up.

keydata_legacy_test.go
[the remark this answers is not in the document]
chrisccoulson: It looks like this change doesn't exist in the most recently pushed branch.

keydata.go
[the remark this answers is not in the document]
chrisccoulson: Yes, generation is ok with me as well.

keydata.go
chrisccoulson: The added test looks ok - one note for future reference is that we have `testutil.DecodeHexString` for embedding bytes into tests which will fail the test if an invalid string is supplied without having to check an error.

chrisccoulson:
Thanks - I think the latest changes look good to me

keydata.go
```diff
@@ -50,7 +50,7 @@ const (
 )
 
 var (
-	keyDataVersion        int = 2
+	keyDataGeneration     int = 2
```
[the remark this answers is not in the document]
chrisccoulson: I've actually got a follow-up PR that makes this public (so that the version can be bound into the AAD)

chrisccoulson (approved):

keydata.go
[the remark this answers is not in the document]
sespiros: oh sorry about that, I thought I had that change in here already, I also make it public in the final PR because I needed it for a test for AAD https://github.com/snapcore/secboot/pull/271/commits/33445cf5c8ba0ebb5e609a0af9e9122232d8f5e9

PR #267, opened 2023-10-27
efi: Fix a potential grub matching bug

chrisccoulson:
Grub's module section has the name "mods" rather than ".mods"


pedronis (approved):
thanks

PR #269, opened 2023-10-27
luks2: run luksKillSlot in batch mode

chrisccoulson:
We currently don't run luksKillSlot in batch mode, which means that
we need to supply an existing key in order to delete a keyslot.
Cryptsetup requires this to be for another keyslot unless the last
keyslot is being deleted, so it doesn't provide any protections to
prevent deletion of the last keyslot. It is intended to ensure that
a user who deletes a keyslot still knows a passphrase for another
keyslot.

The fdemanager service will have an endpoint to delete a keyslot, but
the API does not require an unlock key because unlock keys aren't
something users will generally interact with. In this case, the service
will grab the key used to unlock the device on the current boot from
the kernel and pass this to secboot, so not running luksKillSlot in
batch mode actually doesn't offer any safety barriers. Secboot already
has its own check to prevent deletion of the last keyslot, something
which luksKillSlot doesn't prevent on its own anyway - it prompts the
user when stdin is a tty, but that's not the case for us.

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -822,10 +822,9 @@ func ListLUKS2ContainerRecoveryKeyNames(devicePath string) ([]string, error) {
 }
 
 // DeleteLUKS2ContainerKey deletes the keyslot with the specified name from the
-// LUKS2 container at the specified path. An existing key associated with a different
-// keyslot must be supplied. This will return an error if the container only has a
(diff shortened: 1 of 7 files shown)


pedronis (approved):
lgtm

PR #271, opened 2023-11-17
[3/3] Keydata v3 TPM platform changes

sespiros:
Original commit: https://github.com/chrisccoulson/secboot/commit/87d9aa34d1d80d64ffe82f0a008693799d633cd9

I will split the commit above in 3 separate cascading PRs and add tests for it:

1. https://github.com/snapcore/secboot/pull/265
2. https://github.com/snapcore/secboot/pull/270
2. This one

This PR is rebased on top of 2. It adds the changes to the TPM platform that use the new v3 cross-platform API + fixes existing tests and adds a few more.

### DIFF
--- a/keydata.go
+++ b/keydata.go
@@ -48,12 +48,12 @@ const (
 )
 
 var (
-	keyDataGeneration int = 2
-	sha1Oid               = asn1.ObjectIdentifier{1, 3, 14, 3, 2, 26}
-	sha224Oid             = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 4}
-	sha256Oid             = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 1}
-	sha384Oid             = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 2}
-	sha512Oid             = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 3}
+	KeyDataGeneration = 2
+	sha1Oid           = asn1.ObjectIdentifier{1, 3, 14, 3, 2, 26}
+	sha224Oid         = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 4}
+	sha256Oid         = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 1}
+	sha384Oid         = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 2}
+	sha512Oid         = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 3}
 )
 
 // ErrNoPlatformHandlerRegistered is returned from KeyData methods if no
@@ -715,7 +715,7 @@ func NewKeyData(params *KeyParams) (*KeyData, error) {
 
 	kd := &KeyData{
 		data: keyData{
-			Generation:       keyDataGeneration,
+			Generation:       KeyDataGeneration,
 			PlatformName:     params.PlatformName,
 			Role:             params.Role,
 			PlatformHandle:   json.RawMessage(encodedHandle),
--- a/keydata.go
+++ b/keydata.go
@@ -20,13 +20,11 @@
 package secboot
 
 import (
-	"bytes"
 	"crypto"
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/rand"
 	"encoding/asn1"
-	"encoding/binary"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -843,15 +841,4 @@ func MakeDiskUnlockKey(rand io.Reader, alg crypto.Hash, primaryKey PrimaryKey) (
 	}
 
 	return pk.unlockKey(alg), cleartextPayload, nil
-}
-
-// MarshalKeys serializes the supplied disk unlock key and auxiliary key in
-// to a format that is ready to be encrypted by a platform's secure device.
-func MarshalKeys(key DiskUnlockKey, auxKey PrimaryKey) []byte {
-	w := new(bytes.Buffer)
-	binary.Write(w, binary.BigEndian, uint16(len(key)))
-	w.Write(key)
-	binary.Write(w, binary.BigEndian, uint16(len(auxKey)))
-	w.Write(auxKey)
-	return w.Bytes()
 }
--- a/keydata.go
+++ b/keydata.go
@@ -46,22 +46,29 @@ const (
 )
 
 var (
-	KeyDataGeneration = 2
-	sha1Oid           = asn1.ObjectIdentifier{1, 3, 14, 3, 2, 26}
-	sha224Oid         = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 4}
-	sha256Oid         = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 1}
-	sha384Oid         = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 2}
-	sha512Oid         = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 3}
+	sha1Oid   = asn1.ObjectIdentifier{1, 3, 14, 3, 2, 26}
+	sha224Oid = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 4}
+	sha256Oid = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 1}
+	sha384Oid = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 2}
+	sha512Oid = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 3}
 )
 
-// ErrNoPlatformHandlerRegistered is returned from KeyData methods if no
-// appropriate platform handler is registered using the
-// RegisterPlatformKeyDataHandler API.
-var ErrNoPlatformHandlerRegistered = errors.New("no appropriate platform handler is registered")
+var (
+	// KeyDataGeneration describes the generation number of new keys created by NewKeyData.
+	// This will be supplied via PlatformKeyDataHandler and should generally be
+	// authenticated by the platform code in order to protect against future key format
+	// changes that might have security relevant implications.
+	KeyDataGeneration = 2
 
-// ErrInvalidPassphrase is returned from KeyData methods that require
-// knowledge of a passphrase is the supplied passphrase is incorrect.
-var ErrInvalidPassphrase = errors.New("the supplied passphrase is incorrect")
+	// ErrNoPlatformHandlerRegistered is returned from KeyData methods if no
+	// appropriate platform handler is registered using the
+	// RegisterPlatformKeyDataHandler API.
+	ErrNoPlatformHandlerRegistered = errors.New("no appropriate platform handler is registered")
+
+	// ErrInvalidPassphrase is returned from KeyData methods that require
+	// knowledge of a passphrase is the supplied passphrase is incorrect.
+	ErrInvalidPassphrase = errors.New("the supplied passphrase is incorrect")
+)
 
 // InvalidKeyDataError is returned from KeyData methods if the key data
 // is invalid in some way.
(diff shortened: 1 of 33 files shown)


tpm2/update.go
```diff
@@ -269,19 +245,28 @@ func (k *SealedKeyData) RevokeOldPCRProtectionPolicies(tpm *Connection, authKey
 // On success, each of the supplied KeyData objects will have an updated authorization policy that includes a
 // PCR policy computed from the supplied PCRProtectionProfile. They must be persisted using
 // secboot.KeyData.WriteAtomic.
-func UpdateKeyDataPCRProtectionPolicy(tpm *Connection, authKey secboot.PrimaryKey, pcrProfile *PCRProtectionProfile, keys ...*secboot.KeyData) error {
+func UpdateKeyDataPCRProtectionPolicy(tpm *Connection, authKey secboot.PrimaryKey, pcrProfile *PCRProtectionProfile, incrementPolicyVersion bool, keys ...*secboot.KeyData) error {
```
pedronis, line 248: this function is not tested afaict

tpm2/update.go
```diff
@@ -43,13 +42,33 @@ var skdbUpdatePCRProtectionPolicyImpl = (*sealedKeyDataBase).updatePCRProtection
 //
 // If k.data.policy().pcrPolicyCounterHandle() is not tpm2.HandleNull, then counterPub
 // must be supplied, and it must correspond to the public area associated with that handle.
-func (k *sealedKeyDataBase) updatePCRProtectionPolicyImpl(tpm *tpm2.TPMContext, key secboot.PrimaryKey,
-	counterPub *tpm2.NVPublic, profile *PCRProtectionProfile, session tpm2.SessionContext) error {
+func (k *sealedKeyDataBase) updatePCRProtectionPolicyNoValidate(tpm *tpm2.TPMContext, key secboot.PrimaryKey,
+	counterPub *tpm2.NVPublic, profile *PCRProtectionProfile, incrementPolicyVersion bool,
```
pedronis, line 46: using a bool for incrementPolicyVersion is not very readable, we usually work around that keeping the bool type but always passing a const in the invocations:

const incrementPolicyVersion = true|false
updatePCRProtectionPolicyNoValidate(..., incrementPolicyVersion, ...)

or we need to switch the param to an enum or pass an option struct, the latter is not a obvious change here though

tpm2/update.go
```diff
@@ -105,12 +124,34 @@ func (k *sealedKeyDataBase) updatePCRProtectionPolicyImpl(tpm *tpm2.TPMContext,
...
 		pcrDigests:        pcrDigests,
-		policyCounterName: counterName}
+		policyCounterName: counterName,
+		policySequence:    policySequence}
 	return k.data.Policy().UpdatePCRPolicy(alg, params)
 }
 
-func (k *sealedKeyDataBase) revokeOldPCRProtectionPoliciesImpl(tpm *tpm2.TPMContext, key secboot.PrimaryKey, session tpm2.SessionContext) error {
-	pcrPolicyCounterPub, err := k.validateData(tpm, session)
+func (k *sealedKeyDataBase) updatePCRProtectionPolicy(tpm *tpm2.TPMContext, authKey secboot.PrimaryKey, role string, pcrProfile *PCRProtectionProfile, incrementPolicyVersion bool, session tpm2.SessionContext) error {
```
pedronis, line 132: same issue with the bool

tpm2/seal.go
```diff
@@ -54,204 +58,163 @@ type ProtectKeyParams struct {
...
-	payload := secboot.MarshalKeys(key, authKey)
+type keyDataConstructor func(skd *SealedKeyData, role string, encryptedPayload []byte, kdfAlg crypto.Hash) (*secboot.KeyData, error)
 
-	b, err := aes.NewCipher(symKey[:32])
-	if err != nil {
-		return nil, xerrors.Errorf("cannot create new cipher: %w", err)
-	}
-	stream := cipher.NewCFBEncrypter(b, symKey[32:])
-	stream.XORKeyStream(payload, payload)
+func (fn keyDataConstructor) NewKeyData(skd *SealedKeyData, role string, encryptedPayload []byte, kdfAlg crypto.Hash) (*secboot.KeyData, error) {
```
pedronis, line 72: I don't find the indirection of a method on a func type more readable or idiomatic of simply using the func type alone but maybe I'm missing something

tpm2/seal.go
```diff
@@ -54,204 +58,163 @@ type ProtectKeyParams struct {
...
+	return fn(skd, role, encryptedPayload, kdfAlg)
+}
 
-	kd, err := secbootNewKeyData(&secboot.KeyParams{
+var makeKeyDataNoAuth keyDataConstructor = func(skd *SealedKeyData, role string, encryptedPayload []byte, kdfAlg crypto.Hash) (*secboot.KeyData, error) {
```
pedronis, line 76: same for the var def vs just func definition

tpm2/policy.go
```diff
@@ -178,37 +180,92 @@ func createPcrPolicyCounterImpl(tpm *tpm2.TPMContext, handle tpm2.Handle, update
...
+		// ok, need to create
+		index, err = tpm.NVDefineSpace(tpm.OwnerHandleContext(), nil, public, hmacSession)
+		if err != nil {
+			return nil, err
+		}
+		defer func() {
+			if err == nil {
+				return
+			}
+			tpm.NVUndefineSpace(tpm.OwnerHandleContext(), index, hmacSession)
```
pedronis, line 228: this bit of undo logic does not seem tested

tpm2/seal_test.go
```diff
@@ -310,248 +256,17 @@ func (s *sealSuite) TestProtectKeyWithTPMNoPCRPolicyCounterHandle(c *C) {
...
-		c.Check(err, IsNil)
-
-		for _, k := range protectedKeys {
-			_, _, err = k.RecoverKeys()
-			c.Check(err, ErrorMatches, "invalid key data: cannot complete authorization policy assertions: cannot execute PCR assertions: "+
-				"cannot execute PolicyOR assertions: current session digest not found in policy data")
-		}
-	}
-}
-
```
pedronis, line 419: explanation PR comments  or pointers to relevant commit messages if they exist of why the various tests here have been dropped would be good? some are obviously tied to model auth which is gone, but for some is less obvious why they are gone

pedronis (commented):
did a first pass, seal.go is nicely streamlined

keydata.go
```diff
@@ -48,12 +46,12 @@ const (
 )
 
 var (
-	keyDataGeneration int = 2
-	sha1Oid               = asn1.ObjectIdentifier{1, 3, 14, 3, 2, 26}
-	sha224Oid             = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 4}
-	sha256Oid             = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 1}
-	sha384Oid             = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 2}
-	sha512Oid             = asn1.ObjectIdentifier{2, 16, 840, 1, 101, 3, 4, 2, 3}
+	KeyDataGeneration = 2
```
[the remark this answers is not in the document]
chrisccoulson: I've added a comment to this

tpm2/update_legacy.go
```diff
@@ -144,10 +145,37 @@ func UpdateKeyPCRProtectionPolicyMultiple(tpm *Connection, keys []*SealedKeyObje
...
+	for i, k := range keys[1:] {
+		if k.data.Version() != primaryKey.data.Version() {
+			return fmt.Errorf("key data at index %d has a different metadata version compared to the primary key data", i+1)
+		}
+
+		if _, err := k.validateData(tpm.TPMContext, "", session); err != nil {
+			if isKeyDataError(err) {
+				return InvalidKeyDataError{fmt.Sprintf("%v (%d)", err.Error(), i+1)}
+			}
+			return xerrors.Errorf("cannot validate related key data: %w", err)
```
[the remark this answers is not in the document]
chrisccoulson: This code was copied from `updateKeyPCRProtectionPoliciesCommon` where it was also not tested. It's probably worth adding a case for this in a follow-up.

tpm2/update.go
chrisccoulson: I've added a test for this one now.

tpm2/update.go
chrisccoulson: I've cleaned this up a bit with an enumerated type. I also ended up restoring the old behaviour for the `SealedKeyObject` type where every update increments the counter as opposed to changing the behaviour in this PR, so this function has 3 options now.

tpm2/update.go
chrisccoulson: I've added a public enumerated type for the new `SealedKeyData` API to replace this.

tpm2/seal.go
chrisccoulson: I've removed the indirections here.

tpm2/seal.go
chrisccoulson: Ditto - this is just a normal function definition now.

tpm2/policy.go
chrisccoulson: This code was copied from `createPcrPolicyCounter` where this path was also untested. It's a bit tricky to test this case without mocking the calls to go-tpm2 and there's a bunch of other paths that depend on TPM errors that aren't tested.

I think it's worth addressing these in follow-ups, but by:
- adding an API to github.com/canonical/go-tpm2/mssim to put the simulator into failure mode.
- adding an API to github.com/canonical/go-tpm2/testutil to call the new API after a specific number of commands are executed successfully.
- using the new APIs here to trigger failures at specific points.

tpm2/seal_test.go
chrisccoulson: It looks like these are dropped because there's not an equivalent API that creates multiple keys in a single call, as we aren't sharing PCR policies anymore.

chrisccoulson (commented):
I've responded to some of your comments now (I had an in-progress review that I had to finish to release them)

tpm2/update.go
```diff
@@ -32,6 +32,23 @@ import (
...
+	// to the current counter value.
+	resetPcrPolicyVersion pcrPolicyVersionOption = iota
+
+	// newPcrPolicyVersion indicates that the new policy version should be the current
+	// counter value plus 1.
+	newPcrPolicyVersion
+
+	// incrementPolicyVersion indicates that the new policy version should be the
+	// previous policy version plus 1.
+	incrementPcrPolicyVersion
```
sespiros, line 49: I got slightly confused by this and how it differs from the `newPcrPolicyVersion` option. Should we expand its comment a bit to indicate that this is used only by legacy/v0,v1,v2 TPM keydata versions? (I believe the rest of the difference in logic can be inferred by following the corresponding flows and reading other doc comments)

tpm2/policy_v0.go
```diff
@@ -374,7 +374,7 @@ func (p *keyDataPolicy_v0) UpdatePCRPolicy(alg tpm2.HashAlgorithmId, params *pcr
 		return xerrors.Errorf("cannot compute base PCR policy: %w", err)
 	}
 
-	pcrData.addRevocationCheck(trial, params.policyCounterName, params.policySequence+1)
+	pcrData.addRevocationCheck(trial, params.policyCounterName, params.policySequence)
```
pedronis, line 377: why this change?

tpm2/policy_v1.go
```diff
@@ -147,7 +147,7 @@ func (p *keyDataPolicy_v1) UpdatePCRPolicy(alg tpm2.HashAlgorithmId, params *pcr
 	}
 
 	if params.policyCounterName != nil {
-		pcrData.addRevocationCheck(trial, params.policyCounterName, params.policySequence+1)
+		pcrData.addRevocationCheck(trial, params.policyCounterName, params.policySequence)
```
pedronis, line 150: same question

tpm2/update.go
pedronis: I agree that some further clarification about the 3 modes would be good

pedronis (commented):
some questions

tpm2/policy_v0.go
chrisccoulson: This PR moves incrementing the value earlier so this avoids a double increment.

pedronis: thanks, I followed the code a bit more and makes sense

pedronis (approved):
+1, reminder about the discussion/suggestion here though: https://github.com/snapcore/secboot/pull/271#pullrequestreview-1919142705

chrisccoulson:
> +1, reminder about the discussion/suggestion here though: [#271 (review)](https://github.com/snapcore/secboot/pull/271#pullrequestreview-1919142705)

Thanks, I've added a comment here.

I think I've had an idea for how to clean this up and remove these values (moving the logic to the tpm2/policy_v* implementations and supplying the counter value as an argument instead), but I'll leave that for a follow-up.

PR #273, opened 2023-11-19
tpm2: remove SecureConnectToDefaultTPM

chrisccoulson:
This was intended to address some attacks against the bus between the TPM and the host CPU which can be leveraged to extract secrets, but it isn't currently used, it has some issues, and it doesn't provide any benefit over simpler alternatives.

For context, there are 3 classes of attacks against the communications between the TPM and the host CPU:

 - Passive person-in-the-middle attacks where an adversary is able to monitor communications. This can be mitigated by using parameter encryption with a symmetric key that is derived from key material agreed using a TPM backed asymmetric key. This is what we're already doing today.

 - Active person-in-the-middle attacks where an adversary is able to both monitor and modify communications. This may be able to be mitigated in some circumstances by the use of parameter encryption and integrity protection using keys derived from key material agreed using a certified TPM backed asymmetric key, such as the endorsement key.

 - Attacks where an adversary physically removes the TPM and inserts it into a platform that can spoof the original host platform. There are no software mitigations against this.

Note that firmware TPMs don't exhibit this type of weakness.

SecureConnectoToDefaultTPM tries to mitigate active person-in-the-middle attacks by using a session that is salted with the endorsement key in order to integrity protect commands and responses, and encrypt secrets that are communicated between the TPM and the host CPU. There are several problems here though:

 - Integrity protections must be employed throughout the entire chain of trust in order to be effective, as the adversary can othewise modify PCR extends from the firmware in order to spoof the host platform. This doesn't happen today, and as long as this is the case, there is nothing we can do to mitigate active attacks.

 - The certificate verification code relies on go's crypto/x509 package which doesn't provide a mechanism to ignore the validity period of certificates. Hardware certificates will often expire during the lifetime of a device. UEFI secure boot ignores the validity period.

 - Although endorsement key certificates can be obtained from the TPM and secboot can embed TPM manufacturer root certificates, verification generally requires us to download intermediate certificates. What happens if we can't download an intermediate? Should booting fail hard? If not, we still then have to communicate with the TPM in order to perform PCR extends but these won't have any integrity protection. In this case, an adversary can attack communications by just removing the intermediate certificates from a device and preventing them from being downloaded again.

 - The current code enables parameter encryption for TPM2_Unseal by supplying an additional HMAC session with parameter encryption enabled in addition to the policy session used for authorization. An adversary who can modify communications can remove this extra session so that the TPM returns the unsealed key in cleartext. There is no mitigation for this, even if the policy session is salted with a certified key (removing the extra encryption session invalidates the policy session HMAC), as the same adversary can create an unsalted policy session in order to remove the parameter encryption and use this instead. This will be detected by code running on the host CPU when it processes the response, but it's too late at this point. This is a weakness of TPMs and there is no mitigation for it.

It's clear that nothing we do in secboot can mitigate the active attack scenario, so we should instead focus on mitigating the simpler passive attack case instead. If we assume that communications can be monitored but not modified by an adversary, then we don't need integrity protections for commands and parameter encryption doesn't need to use a certified TPM key (it just needs to use any TPM protected asymmetric key that we create).

Note, since I worked on this, there has been a change proposed to the kernel to protect the kernel's TPM communications: https://lore.kernel.org/linux-integrity/20231127190854.13310-1-James.Bottomley@HansenPartnership.com/T/#mb3220ef6ce12824a1427b596b9bab6013fc484b8. The approach here though is to provide a way to prove that TPM communications haven't been compromised during the current boot. This doesn't offer any protection whilst firmware is running, and cannot protect against the stealing of secrets


xnox:
separately i am considering to package all the tpm certs..... but there are a lot of them

pedronis (commented):
thank you, after 6f8a19e668cd6afb1f99af44c72d0ca011d9fb82 there are very places in production code that use HmacSession() and pass it, maybe they could grow comments explaining why it is used case by case

chrisccoulson:
> thank you, after [6f8a19e](https://github.com/snapcore/secboot/commit/6f8a19e668cd6afb1f99af44c72d0ca011d9fb82) there are very places in production code that use HmacSession() and pass it, maybe they could grow comments explaining why it is used case by case



> thank you, after [6f8a19e](https://github.com/snapcore/secboot/commit/6f8a19e668cd6afb1f99af44c72d0ca011d9fb82) there are very places in production code that use HmacSession() and pass it, maybe they could grow comments explaining why it is used case by case

Sure, I can add some comments where appropriate.

I'm going to get back to working on this because as part of testing the new `efi/preinstall` code, I want to get the TPM connection code and `internal/tpm2test` using the `tpm2.TPMDevice` interface (and the corresponding `github.com/canonical/go-tpm2/linux.DefaultTPM2Device` and associated methods for accessing the in-kernel resource managed device and physical presence interface), and it's going to be a lot easier starting with less code :)

pedronis (approved):
thank you

PR #274, opened 2023-11-21
efi: remove deprecated APIs

chrisccoulson:
This removes deprecated APIs and test data, and will make it easier to
generate new mock binaries without having to adjust digests in associated
unit tests. I want to generate new mock grub binaries in another PR to add
the EFI prefix.

Note that this removes some data that is used by tools/gen-compattest-data,
but it turns out that a previous PR (https://github.com/snapcore/secboot/pull/156)
already removed test data required by it and it hasn't worked since then. I need
to rework how we test that changes are backwards compatible, so I'll fix that in
another PR.


chrisccoulson:
Ugh, of course, this also makes tools/gen-compatdata unusable, so I'll need to fix that first

pedronis (approved):
lgtm

> Ugh, of course, this also makes tools/gen-compatdata unusable, so I'll need to fix that first

if the data is checked in as we discussed and the tests pass and do not want to block on this, you can leave a TODO there and come back in a different PR

You can also comment out code at the same time to help landing things

PR #275, opened 2023-11-21
luks2: don't time tests to verify KDF settings

chrisccoulson:
There are a few tests where we time the KDF operation because there
isn't another way to determine if it was configured correctly by
cryptsetup. These tests have always been a little bit flaky, and they
have been significantly more so recently.
    
Stop timing them, and instead:
- in internal/luks2, just rely on the fact that we verify the expected
  commandline arguments were passed to cryptsetup. We have to assume
  that these options are correct, which is what the existing tests
  attempted to confirm.
- in crypt_test.go, ensure that we pass the correct target time to
  internal/luks2.


pedronis (approved):
+1

PR #280, opened 2024-01-17
Keydata v3 auxiliary primary rename and legacy keydata test

sespiros:
First 2 commits taken from https://github.com/chrisccoulson/secboot/tree/keydata-format-improvements-v2 and rebased on top of recent master.

This adds the following changes:
1) Renaming AuxiliaryKey to PrimaryKey
2) adds a basic legacy behavior test to ensure correct behavior for the upcoming changes related to the new keydata format.

### DIFF
--- a/keydata_test.go
+++ b/keydata_test.go
@@ -244,7 +244,7 @@ func (s *keyDataTestBase) TearDownSuite(c *C) {
 	RegisterPlatformKeyDataHandler(mockPlatformName, nil)
 }
 
-func (s *keyDataTestBase) newKeyDataKeys(c *C, sz1, sz2 int) (DiskUnlockKey, AuxiliaryKey) {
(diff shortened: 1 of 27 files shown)


keydata_test.go
```diff
@@ -1427,3 +1427,88 @@ func (s *keyDataSuite) TestReadAndWriteWithLegacySnapModelAuthKey(c *C) {
 	c.Check(err, IsNil)
 	c.Check(ok, testutil.IsTrue)
 }
+
+func (s *keyDataSuite) TestLegacyKeyData(c *C) {
```
pedronis, line 1431: this needs to be mentioned in the description and summary, as it is not quite related with the rename:

Keydata v3 auxilary primary rename and legady keydata test

etc

pedronis (approved):
thanks you for splitting it out, small remark about the new test vs the summary/description of this

keydata_test.go
sespiros: Changed the title and summary, thanks for pointing it out.

chrisccoulson (approved):

PR #282, opened 2024-02-01
Add plainkey platform

chrisccoulson:
This platform provides a way to unlock an encrypted container using a
key recovered from unlocking another container, and is suitable when
access to one encrypted container implies access to another one.

The intention of this is to replace the existing mechanism for unlocking
the save partition in run mode with something that uses a proper keyslot.

Keys are protected with a "platform key" - this is the key that is
stored in the first encrypted container. For the intended use case,
this key is stored inside the data partition, and then loaded by
snap-bootstrap, which should call plainkey.SetPlatformKeys before
unlocking the save partition.

Although normally there will only be a single key supplied to
SetPlatformKeys, it supports setting multiple keys and then matching
a key data object to the correct one. This is to support the possibility
of changing this key during reprovisioning (ie, as part of factory
reset), where there would need to exist 2 keys (the old and new one)
temporarily - this would prevent against loss of data if the process
was interrupted at any point.

### DIFF
--- a/plainkey/keydata.go
+++ b/plainkey/keydata.go
@@ -182,9 +182,8 @@ type platformKeyId struct {
 type keyData struct {
 	Version int `json:"version"`
 
-	// Nonce has a dual purpose - it provides a salt that is used to derive
-	// an encryption key from the platform key, and it provides the GCM nonce.
-	Nonce []byte `json:"nonce"`
+	Salt  []byte `json:"salt"`  // Used to derive the symmetric key from the platform key
+	Nonce []byte `json:"nonce"` // the GCM nonce
 
 	// PlatformKeyID is used to identify the loaded platform key to
 	// use for key recovery.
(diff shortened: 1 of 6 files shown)


plainkey/platform.go
```diff
@@ -0,0 +1,156 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package plainkey
```
pedronis, line 20: it would be good to have a doc comment for the package here.

probably a good idea to add one also for tpm2 at some point

plainkey/platform.go
```diff
@@ -0,0 +1,156 @@
...
+func SetPlatformKeys(keys ...[]byte) {
+	platformKeysMu.Lock()
+	platformKeys = keys
+	platformKeysMu.Unlock()
+}
+
+// AddPlatformKeys adds keys that will be used by this platform to recover other
+// keys. These are typically stored in and loaded from an encrypted container that is
+// unlocked via some other mechanism.
+func AddPlatformKeys(keys ...[]byte) {
```
pedronis, line 58: do we need this? anyway it seems not to be tested atm

plainkey/keydata.go
```diff
@@ -0,0 +1,264 @@
...
+	Alg    hashAlg `json:"alg"`
+	Salt   []byte  `json:"salt"`
+	Digest []byte  `json:"digest"`
+}
+
+type keyData struct {
+	Version int    `json:"version"`
+	Nonce   []byte `json:"nonce"`
+
+	PlatformKeyID platformKeyId `json:"platform-key-id"`
```
pedronis, line 183: it would be good to have a doc comment on this one

pedronis (commented):
thanks, some small comments

plainkey/platform.go
chrisccoulson: I've added a comment here now.

plainkey/platform.go
chrisccoulson: I've removed this.

plainkey/keydata.go
chrisccoulson: I added a comment here.

I added a comment for the Nonce field as well and then realized it didn't make sense because the field was holding 2 unrelated values. I've pushed another change that splits this into 2 distinct fields.

pedronis (approved):
thank you

PR #283, opened 2024-02-20
keydata_test.go: fix platform name test

sespiros:
newKeyDataKeys and mockProtectKeys was changed slightly after the recent key format changes and this test failed after the rebase.


chrisccoulson (approved):
Thanks, this looks ok to me

PR #284, opened 2024-02-20
Keydata testsuite minor refactors and improvements and hashAlg

sespiros:
This addresses:
- keyDataLegacyTestBase not used from https://github.com/snapcore/secboot/pull/265#discussion_r1450626357
- use DecodeHexString from https://github.com/snapcore/secboot/pull/265#discussion_r1487572075

Also makes hashAlg public as it will be used in a follow-up PR which will slightly rework the mockPlatformKeyDataHandler to use GCM and associated data to check expected values.

### DIFF
--- a/keydata.go
+++ b/keydata.go
@@ -185,7 +185,7 @@ type KeyDataReader interface {
 	ReadableName() string
 }
 
-// HashAlg corresponds to a digest algorithm.
+// HashAlg provides an abstraction for crypto.Hash that can be serialized to JSON and DER.
 type HashAlg crypto.Hash
 
 var hashAlgAvailable = (*HashAlg).Available
--- a/keydata.go
+++ b/keydata.go
@@ -39,7 +39,7 @@ import (
 
 const (
 	kdfType                            = "argon2i"
-	nilHash                    hashAlg = 0
+	nilHash                    HashAlg = 0
 	passphraseKeyLen                   = 32
 	passphraseEncryptionKeyLen         = 32
 	passphraseEncryption               = "aes-cfb"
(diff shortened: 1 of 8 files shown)


keydata.go
```diff
@@ -189,24 +189,24 @@ type KeyDataReader interface {
 	ReadableName() string
 }
 
-// hashAlg corresponds to a digest algorithm.
-type hashAlg crypto.Hash
+// HashAlg corresponds to a digest algorithm.
```
chrisccoulson, line 192: It might be worth just adding a note as to why this exists (it's just a crypto.Hash that can be serialed to JSON and DER)

chrisccoulson (changes requested):
Thanks - this looks ok, although it does need a rebase on master now to resolve some merge conflicts.

I notice that `xerrors.Errorf` is replaced with `fmt.Errorf` in a few places - that's ok, but I think it's probably worth another PR to drop the xerrors dependency entirely given that we depend on go1.18 which has the required functionality in the standard library.

sespiros:
> Thanks - this looks ok, although it does need a rebase on master now to resolve some merge conflicts.

Rebased on latest master.

> I notice that `xerrors.Errorf` is replaced with `fmt.Errorf` in a few places - that's ok, but I think it's probably worth another PR to drop the xerrors dependency entirely given that we depend on go1.18 which has the required functionality in the standard library.

I did a cleanup for xerrors in secboot (that I will push in another PR) but it is still added as an indirect dependency. I can do the same for efilib (which also uses go 1.18). go-tpm2 also has it in its go.sum although it's not using it. Any reason why go-tpm2 is still in go 1.9?

sespiros:
@chrisccoulson I also opened https://github.com/canonical/go-efilib/pull/13

chrisccoulson (approved):
Thanks, this looks fine to me.

PR #286, opened 2024-03-08
use PBKDF2 for the LUKS2 keyslots

chrisccoulson:
We currently use a minimal Argon2i configuration for the main unlock
keyslots and then Argon2i benchmarked to 2s by default for recovery
keyslots. Note that the cryptsetup default now is Argon2id, but Argon2
is overkill in this context anyway.

For the main unlock keyslots, hardcode a minimal PBKDF2 configuration.
Also hardcode a PBKDF2 configuration for the recovery keyslots, but
use 600,000 iterations which is the current OWASP recommendations for
ordinary passphrases
(https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html).
This is still overkill for a 128bit recovery key, and if we implement a
256bit recovery key then we could justify using a minimal configuration
here as well. Note that systemd is using minimal PBKDF2 configurations
for both normal and recovery keyslots.

This simplifies the API by not requiring KDFs to be configured, and
means that the configuration is suitable where FIPS140 compliance is
required.

### DIFF
--- a/argon2.go
+++ b/argon2.go
@@ -28,7 +28,6 @@ import (
 	"golang.org/x/xerrors"
 
 	"github.com/snapcore/secboot/internal/argon2"
-	"github.com/snapcore/secboot/internal/luks2"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


pedronis (approved):
thanks

PR #287, opened 2024-03-09
add support for the hybrid mode of Argon2 for passphrases

chrisccoulson:
Passphrase support currently hardcodes the use of argon2i, which is
the data-independent version of Argon2. Whilst this has better
side-channel resistance than data-dependent versions, it requires
more iterations to protect against time-memory tradeoff attacks.

Cryptsetup recently switch to argon2id by default for passphrase
hashing, which is the hybrid version that provides a good balance
between side-channel resistance and time-memory tradeoff.

This introduces support for selecting between the 2 versions for
passphrase support.

There are other versions of Argon2, such as the data-dependent version
which sacrifices side-channel resistance for better resistance against
time-memory tradeoff attacks, but this isn't supported by the argon2 go
package.

### DIFF
--- a/argon2_test.go
+++ b/argon2_test.go
@@ -80,6 +80,12 @@ func (s *argon2Suite) checkParams(c *C, opts *Argon2Options, ncpus uint8, params
 			expectedMem = 1 * 1024 * 1024
 		}
 		c.Check(params.MemoryKiB, Equals, expectedMem)
+
+		expectedThreads := opts.Parallel
+		if expectedThreads == 0 {
+			expectedThreads = ncpus
+		}
+		c.Check(params.Threads, Equals, expectedThreads)
 	} else {
 		targetDuration := opts.TargetDuration
 		if targetDuration == 0 {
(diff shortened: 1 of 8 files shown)


pedronis (commented):
this has conflicts now

argon2_test.go
```diff
@@ -96,97 +112,188 @@ func (s *argon2Suite) checkParams(c *C, opts *Argon2Options, ncpus uint8, params
...
+
+func (s *argon2Suite) TestInProcessKDFTimeInvalidThreads(c *C) {
+	_, err := InProcessArgon2KDF.Time(Argon2id, &Argon2CostParams{Time: 4, MemoryKiB: 32, Threads: 0})
+	c.Check(err, ErrorMatches, `invalid number of threads`)
+}
+
+func (s *argon2Suite) TestModeConstants(c *C) {
+	c.Check(Argon2i, Equals, Argon2Mode(argon2.ModeI))
+	c.Check(Argon2id, Equals, Argon2Mode(argon2.ModeID))
+}
```
pedronis, line 295: I see we have this test that enforces the property but still a bit odd overall

pedronis (approved):
thanks, the assumption around Mode values and using conversions is slightly odd. As one could have also declared type Argon2Mode = argon2.Mode perhaps?

chrisccoulson:
> thanks, the assumption around Mode values and using conversions is slightly odd. As one could have also declared type Argon2Mode = argon2.Mode perhaps?

I've just updated this to make `Argon2Mode` a type alias of `argon2.Mode` instead.

PR #288, opened 2024-03-09
add support for PBKDF2 for passphrases

chrisccoulson:


internal/pbkdf2/pbkdf2.go
```diff
@@ -0,0 +1,145 @@
...
+	benchmarkSalt = []byte("0123456789abcdefghijklmnopqrstuv")
+)
+
+var timeExecution = func(params *Params) time.Duration {
+	start := time.Now()
+	if _, err := Key(benchmarkPassword, benchmarkSalt, params, uint(params.HashAlg.Size())); err != nil {
+		panic(err)
+	}
+	return time.Now().Sub(start)
+}
```
pedronis, line 45: this bit of code in itself is not reached by tests

internal/pbkdf2/pbkdf2.go
```diff
@@ -0,0 +1,145 @@
...
+func Key(passphrase string, salt []byte, params *Params, keyLen uint) ([]byte, error) {
+	switch {
+	case params == nil:
+		return nil, errors.New("nil params")
+	case params.Iterations > math.MaxInt:
+		return nil, errors.New("too many iterations")
+	case !params.HashAlg.Available():
+		return nil, errors.New("unavailable digest algorithm")
+	case keyLen > math.MaxInt:
+		return nil, errors.New("invalid key length")
```
pedronis, line 142: isn't there anything to check between keyLen and HashAlg?

pedronis (commented):
did a pass, couple of remarks/questions

internal/pbkdf2/pbkdf2.go
chrisccoulson: I've added a test that doesn't mock this now.

internal/pbkdf2/pbkdf2.go
chrisccoulson: There isn't really any need to do that - one could use any hash algorithm to produce a key of any length, eg, if you request a key of 64 bytes with SHA-256 as the digest and 10000 iterations, it runs 10000 iterations twice to produce 2 blocks of 32-bytes and there's nothing in the design of PBKDF2 that disallows this.

pedronis (approved):
thanks

PR #289, opened 2024-03-09
add PIN support to KeyData

chrisccoulson:


ZeyadYasser (commented):
Are we missing `NewTPMPINProtectedKey` helper or is it planned to land along with  unlocking support in a future PR?

keydata.go
```diff
@@ -512,6 +594,26 @@ func (d *KeyData) updatePassphrase(payload, oldAuthKey []byte, passphrase string
...
+	handle, err := handlerInfo.handler.ChangeAuthKey(d.platformKeyData(), oldAuthKey, authKey, platformContext)
+	if err != nil {
+		return err
+	}
+
+	d.data.PlatformHandle = handle
+	return nil
+}
+
 func (d *KeyData) openWithPassphrase(passphrase string) (payload []byte, authKey []byte, err error) {
```
pedronis, line 617: there is no openWithPIN because there is no extra encryption with PINs, right?

pedronis (commented):
did a pass, couple of comments/questions

keydata.go
chrisccoulson: That's right - the PIN derived key is only used for authentication.

pin.go
```diff
@@ -0,0 +1,152 @@
...
+			}
+			dst[dstI] = b
+			dstI -= 1 // Move on to the next destination byte.
+			b = 0
+		}
+
+		srcI -= 1 // Move on to the next source byte.
+	}
+
+	if dst[0] == 0x80 {
```
pedronis, line 147: can you add a comment about this case

pedronis (commented):
did another pass, couple of comments

pedronis (approved):
+1

PR #290, opened 2024-03-10
argon2: use a fixed key length for benchmarking

chrisccoulson:
The key length has a negligible effect on the execution time, so use
a fixed key length of 32 bytes for benchmarking.

Note that this keeps the keyLen argument to KDFOptions.deriveCostParams
as this is going to become an interface in a future PR, where the key
length will be used to select the default PBKDF2 hash.


pedronis (approved):
thanks

PR #291, opened 2024-03-13
argon2: refactor API to prepare for PBKDF2 support for passphrases

chrisccoulson:
This renames the existing KDFOptions, KDFCostParams and KDF types to
Argon2Options, Argon2CostParams and Argon2KDF in preparation for
the introduction of PBKDF2 support for passphrases in another PR, which
will use a new options implementation. Both options implementations will
implement a new interface (KDFOptions).

As PBKDF2 won't be abstracted like Argon2 is because it doesn't use lots
of memory, also remove the kdf argument from functions that take them.
Instead, have a process wide Argon2KDF implementation, set by calling
SetArgon2KDF.

### DIFF
--- a/argon2.go
+++ b/argon2.go
@@ -21,8 +21,8 @@ package secboot
 
 import (
 	"errors"
-	"math"
 	"runtime"
+	"sync"
 	"time"
 
 	"golang.org/x/xerrors"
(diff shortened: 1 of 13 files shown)


pedronis (approved):
some of this in itself is a bit strange but I suppose it will clean itself up in the next steps

PR #293, opened 2024-03-15
argon2: define test expectations explicitly

chrisccoulson:
This makes it clearer that we are actually testing something valid.

### DIFF
--- a/argon2_test.go
+++ b/argon2_test.go
@@ -23,7 +23,6 @@ import (
 	"math"
 	"os"
 	"runtime"
-	"time"
 
 	snapd_testutil "github.com/snapcore/snapd/testutil"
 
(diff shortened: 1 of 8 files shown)


argon2_test.go
```diff
@@ -41,25 +40,25 @@ type argon2Suite struct {
...
 	c.Check(unix.Sysinfo(&sysInfo), IsNil)
 
 	halfTotalRamKiB := uint64(sysInfo.Totalram) * uint64(sysInfo.Unit) / 2048
-	if halfTotalRamKiB > math.MaxUint32 {
-		halfTotalRamKiB = math.MaxUint32
+	if halfTotalRamKiB > math.MaxInt32 {
+		halfTotalRamKiB = math.MaxInt32
 	}
-	s.halfTotalRamKiB = uint32(halfTotalRamKiB)
+	s.maxMemoryCostKiBAuto = int(halfTotalRamKiB)
```
pedronis, line 55: I'm probably confused, we set this but we don't seem to read it anywhere ?

pedronis (commented):
question

argon2_test.go
chrisccoulson: You're right, this value appears to be unused.

pedronis (approved):
thank you

PR #294, opened 2024-03-16
internal/argon2: return errors

chrisccoulson:
We're currently wrapping x/crypto/argon2 which panics on some invalid
arguments. Instead, move the checking of these arguments in order to
avoid some panics.


pedronis (approved):
thanks

PR #297, opened 2024-04-04
Add hooks platform

chrisccoulson:
This adds a KeyData platform that permits sealing and recovery of keys
via hooks provided via a device's gadget and kernel snaps, for platforms
where we don't have a native secboot platform.


chrisccoulson:
One thing I've just noticed here is that I've missed some integration tests between the new platform and `secboot.KeyData`. I'll push another update with those in, although it doesn't stop anyone reviewing what's already here.

hooks/platform.go
```diff
@@ -0,0 +1,158 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+	"sync"
+
+	"github.com/snapcore/secboot"
+)
+
+const (
+	platformName = "fde-hooks-v3"
```
pedronis, line 42: I wonder if for consistency with the other naming and the context of all of this we need the "fde-" of the name here?

hooks/keydata.go
```diff
@@ -0,0 +1,284 @@
...
+		aead, err := cipher.NewGCM(b)
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("cannot create AEAD for AEAD compat: %w", err)
+		}
+		ciphertext = aead.Seal(nil, nonce, payload, aad)
+
+		var encryptedKey []byte
+		encryptedKey, handle, err = keyProtector.ProtectKey(rand, symKey, nil)
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("cannot protect symmetric key for AEAD compat: %w", err)
```
pedronis, line 200: it would be good to have a test for the error path coming from the Protector

hooks/keydata.go
```diff
@@ -0,0 +1,284 @@
...
+		}
+
+		aeadCompat = &aeadCompatData{
+			Nonce:        nonce,
+			EncryptedKey: encryptedKey,
+		}
+	default:
+		ciphertext, handle, err = keyProtector.ProtectKey(rand, payload, aad)
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("cannot protect key: %w", err)
```
pedronis, line 210: same here

hooks/keydata.go
```diff
@@ -0,0 +1,284 @@
...
+// key that was originally supplied to [NewProtectedKey].
+//
+// On success, this will automatically update the corresponding *[secboot.KeyData]
+// that this key data was created from using [NewKeyData].
+func (d *KeyData) SetAuthorizedBootModes(key secboot.PrimaryKey, modes ...string) error {
+	if err := d.data.Scope.SetAuthorizedBootModes(key, d.k.Role(), modes...); err != nil {
+		return err
+	}
+	return d.k.MarshalAndUpdatePlatformHandle(d)
+}
```
pedronis, line 284: these methods don't seem to be exercised in tests

hooks/platform.go
```diff
@@ -0,0 +1,158 @@
...
+		}
+	}
+
+	keyRevealerMu.Lock()
+	defer keyRevealerMu.Unlock()
+
+	switch {
+	case kd.data.AEADCompat != nil:
+		symKey, err := keyRevealer.RevealKey(kd.data.Handle, kd.data.AEADCompat.EncryptedKey, nil)
+		if err != nil {
```
pedronis, line 82: same for the revealer

hooks/platform.go
```diff
@@ -0,0 +1,158 @@
...
+			return nil, &secboot.PlatformHandlerError{
+				Type: secboot.PlatformHandlerErrorInvalidData,
+				Err:  fmt.Errorf("cannot recover key: %w", err),
+			}
+		}
+		return payload, nil
+	default:
+		payload, err := keyRevealer.RevealKey(kd.data.Handle, encryptedPayload, aad)
+		if err != nil {
```
pedronis, line 114: and here

pedronis (commented):
thank you, did a first pass

hooks/platform.go
chrisccoulson: I'm not sure - I just based this on what's currently in snapd (`fde-hooks-v2`)

pedronis: ah, I forgot about that. This is not unreasonable then

hooks/keydata.go
chrisccoulson: I've added a test for this now.

hooks/keydata.go
chrisccoulson: And here

hooks/platform.go
chrisccoulson: I've added tests for key revealer errors, plus tests for several other error paths.

hooks/platform.go
chrisccoulson: There's a test for this one as well.

chrisccoulson:
I've added most tests now with the exception of tests for `KeyData.SetAuthorizedSnapModels` and `KeyData.SetAuthorizedBootModes`,which I'll do after the weekend. I've also added a bunch of extra tests for error paths in `hooksPlatform.RecoverKeys`.

I'm going to add some tests for the JSON serialization next week as well. In order to avoid non-determinism in the outputs here (because of the ECDSA signatures in `bootscope` which consume randomness), I've added an `io.Reader` for functions that consume random bytes so that unit tests can override the source of randomness with fixed sets of bytes where they need determinism, and plumbed this into `bootscope`. I've had to add an additional helper to `internal/testutil` called `BypassMaybeReadByte` to bypass the usage of go's internal `randutil.MaybeReadByte` which is used in various go crypto packages in order to make the results of APIs non-deterministic even for the same sequence of input bytes.

hooks/keydata.go
chrisccoulson: I've added some tests for both of these methods now.

chrisccoulson:
The 1.22 tests are failing unfortunately due to change in nonce generation when creating ECDSA signatures, although it's still deterministic (well, it is with my `BypassMaybeReadByte` hack. I'm re-working them to try to keep the hardcoded byte sequences in the tests as I think it's important for detecting changes that might have backwards compatibility implications (particularly with the marshalling / unmarshalling tests - and I'd like to use this same approach for testing in other packages), but this is a spanner in the works I could have done without tbh :(

pedronis (approved):
thanks

PR #298, opened 2024-04-10
efi: add sbatlevel for shim from 24.04

valentindavid:
Note, this is for the uc22 branch.

### DIFF
--- a/efi/secureboot_policy.go
+++ b/efi/secureboot_policy.go
@@ -1271,7 +1271,13 @@ func AddSecureBootPolicyProfile(profile *secboot_tpm2.PCRProtectionProfile, para
 	// XXX: Work around the lack of support for handling SBAT revocations by just generating
 	// a profile for the current values under the MS UEFI CA (minus the latest, which requires
 	// an explicit opt-in via SbatPolicy).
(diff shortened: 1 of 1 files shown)


julian-klode:
I suppose also add `2024010900` level `shim,4\ngrub,3\ngrub.debian,4\n`; I just bumped the latest level in shim to a newer one, so this will popup eventually in shim builds, albeit in the shim we have now it is the latest policy.

But then again why not actually also support the latest policy. I updated it yesterday upstream to 2024040900
"shim,4\ngrub,4\ngrub.peimage,2\n"

valentindavid:
It is going to be a big policy generated...

To be sure I understood, I need to add those?

* `sbat,1,2024010900\nshim,4\ngrub,3\ngrub.debian,4\n`
* `sbat,1,2024040900\nshim,4\ngrub,4\ngrub.peimage,2\n`

valentindavid:
@julian-klode I have added the expected values you said. Could you verify them?

pedronis (approved):
thanks

valentindavid:
I cannot merge. Whoever has that power should do it. Thank you.

chrisccoulson (approved):

chrisccoulson:
I've approved this. I'm very glad we implemented support for nested PolicyORs because we're definitely going to be using them with this :)

PR #299, opened 2024-04-29
efi: Add profiles for PCRs 0 and 2

chrisccoulson:
This adds profiles for PCR0 (platform firmware) and PCR2 (host firmware that runs from adapter cards or firmware that runs on embedded controllers)

### DIFF
--- a/efi/efi.go
+++ b/efi/efi.go
@@ -20,7 +20,9 @@
 package efi
 
 const (
-	bootManagerCodePCR = 4 // Boot Manager Code and Boot Attempts PCR
-	secureBootPCR      = 7 // Secure Boot Policy Measurements PCR
-	kernelConfigPCR    = 12
+	platformFirmwarePCR = 0 // SRTM, POST BIOS, and Embedded Drivers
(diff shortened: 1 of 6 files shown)


pedronis (commented):
this looks good, Is the plan that we should start including these in profiles going forward? a bit more motivation in the summary would be useful

chrisccoulson:
> this looks good, Is the plan that we should start including these in profiles going forward? a bit more motivation in the summary would be useful

The goal is that an eventual PR that implements the pre-install checks will return an options that contains the optimal set of TCG defined PCRs, which may or many not include these new profiles depending on the device configuration. There will be some additional options to control this, such as selecting "optimal" configuration or "most secure".

pedronis (approved):
+1  but please add some of 

```
The goal is that an eventual PR that implements the pre-install checks will return an options that contains the optimal set of TCG defined PCRs, which may or many not include these new profiles depending on the device configuration. There will be some additional options to control this, such as selecting "optimal" configuration or "most secure".
```

to the description

PR #301, opened 2024-05-09
efi: add Microsoft UEFI CA 2023 data

chrisccoulson:
Note that there aren't any tests for this yet because as far as we're aware, it hasn't been used to sign anything and so there isn't a signing certificate in the wild from which we can create mock binaries in the way that we test image matching with the 2011 CA.


pedronis (approved):
thanks

PR #302, opened 2024-05-22
efi: Don't create profiles with invalid separators

chrisccoulson:
We copy the EV_SEPARATOR events from the firmware log because it can
be one of two valid values. But we should be aborting early if any of
these separators indicate that an error occurred in firmware rather
than continuing to build a profile.


pedronis (approved):
thank you

PR #303, opened 2024-05-22
efi: PCR4: Add support for certain firmware applications loaded during OS-present

chrisccoulson:
Some newer laptops are loading endpoint management applications that are
shipped with firmware, and these are being loaded as part of the OS-present
environment with the LoadImage API, resulting in them being measured to PCR4.
This currently causes us to mis-predict values for PCR4 because everything
after the separator (the pre-OS to OS-present transition) is assumed to
belong to the OS and gets dropped from the predicted measurements.

A future PR will provide support for detecting compatibility for FDE,
with an option for detecting and disallowing these firmware applications.
In most cases, they should be disabled (and it's possible to disable
Absolute from userspace on Dell laptops), but we should add support here
for specific, known endpoint management applications for any edge cases where
these agents cannot be disabled.

This adds support for any applications where the firmware volume
filename GUID is well known and maps to "AbsoluteAbtInstaller" or
"AbsoluteComputraceInstaller".

### DIFF
--- a/efi/fw_load_handler.go
+++ b/efi/fw_load_handler.go
@@ -26,6 +26,7 @@ import (
 	"fmt"
 
 	efi "github.com/canonical/go-efilib"
+	"github.com/canonical/go-efilib/guids"
 	"github.com/canonical/go-tpm2"
 	"github.com/canonical/tcglog-parser"
 	"golang.org/x/xerrors"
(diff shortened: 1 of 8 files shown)


efi/fw_load_handler.go
```diff
@@ -236,22 +237,90 @@ func (h *fwLoadHandler) measureBootManagerCodePreOS(ctx pcrBranchContext) error
...
-	return errors.New("missing separator")
+	if !measuredSeparator {
+		return errors.New("missing separator")
+	}
+
+	// Some newer laptops including those from Dell and Lenovo execute code from a firmware volume as part
+	// of the OS-present environment, before shim runs, and using the LoadImage API which results in an
+	// additional measurement to PCR4. Copy this into the profile if it's part of a well-known endpoint
+	// management application, else return an error. Anything else here will be picked up by the pre-install
+	// checks.
```
pedronis, line 271: I suppose it doesn't make sense to make this behavior of supporting these payload a policy option?

efi/fw_load_handler.go
```diff
@@ -236,22 +237,90 @@ func (h *fwLoadHandler) measureBootManagerCodePreOS(ctx pcrBranchContext) error
...
+
+		name, known := guids.FileNameString(efi.GUID(fvf))
+		if !known {
+			return fmt.Errorf("unknown image loaded from firmware volume during OS-present: %v", data.DevicePath)
+		}
+		switch name {
+		case "AbsoluteAbtInstaller", "AbsoluteComputraceInstaller":
+			ctx.ExtendPCR(bootManagerCodePCR, tpm2.Digest(event.Digests[ctx.PCRAlg()]))
+		default:
+			return fmt.Errorf("unexpected image loaded from firmware volume during OS-present: %v", data.DevicePath)
```
pedronis, line 314: we would hit the errors here already when installing, right?

pedronis (commented):
lgtm, some questions to double check things

efi/fw_load_handler.go
chrisccoulson: Yes, in fact the check in my current branch uses a shared function `isAbsoluteAgentLaunch(*tcglog.Event)` - I was hoping to push an update to this before you got to reviewing it that would introduce the shared function. I'm currently in the process of splitting the new checks into a `efi/preinstall` sub-package that will have a main `RunChecks` API, but untangling things is a bit of a challenge (not impossible) and there'll need to be another internal package with a few helpers that's shared by both packages.

efi/fw_load_handler.go
chrisccoulson: I'd rather not make this an option for now - as far as we can tell, Absolute seems to be the only problematic firmware component we know about so far and the checks that will be added later will fail if anything else ends up being measured to PCR4 before the OS. I think we're probably better off adding additional accommodations for other components as and when we encounter them, where appropriate.

pedronis (approved):
thank you

efi/fw_load_handler.go
pedronis: ok

PR #304, opened 2024-05-30
efi: update go-efilib to v0.9.6

chrisccoulson:
(Draft for now because it depends on https://github.com/snapcore/secboot/pull/303)

This adds a new function for mocking the variable backend, which will
be required in a future PR which uses new functions from go-efilib that
wrap around ReadVariable but where we want to be able to override
the HostEnvironment in production code.


chrisccoulson:
Closing this because it gets updated as part of https://github.com/snapcore/secboot/pull/303

PR #309, opened 2024-06-06
efi: Prepare to add pre-install detection

chrisccoulson:
This changes a few types, replaces pcrProfileFlags with pcrFlags
and introduces a PCRProfileEnablePCRsOption interface which is
implemented by the existing profile options and will be implemented by
a new composite option returned by the detection code.


efi/efi.go
```diff
@@ -19,10 +19,50 @@
...
-	platformFirmwarePCR = 0 // SRTM, POST BIOS, and Embedded Drivers
-	driversAndAppsPCR   = 2 // UEFI Drivers and UEFI Applications
-	bootManagerCodePCR  = 4 // Boot Manager Code and Boot Attempts PCR
-	secureBootPCR       = 7 // Secure Boot Policy Measurements PCR
-	kernelConfigPCR     = 12
+	platformFirmwarePCR tpm2.Handle = 0 // SRTM, POST BIOS, and Embedded Drivers
+	driversAndAppsPCR   tpm2.Handle = 2 // UEFI Drivers and UEFI Applications
+	bootManagerCodePCR  tpm2.Handle = 4 // Boot Manager Code and Boot Attempts PCR
+	secureBootPolicyPCR tpm2.Handle = 7 // Secure Boot Policy Measurements PCR
+	kernelConfigPCR     tpm2.Handle = 12
```
pedronis, line 29: is this just convience or does the spec considers PCR number references also Handles?

pedronis (commented):
I have a question about the reuse of the Handle type

efi/efi.go
chrisccoulson: Yes, PCRs are just considered like any other resource in the TPM and are addressed by a handle, in this case, with `0x00` being the first octet for all PCRs, which is why they can often just be referenced by an integer, although at some point that has to get converted to a type that's based on uint32 (which is what `tpm2.Handle` is.

See, eg, how go-tpm2 takes a `ResourceContext` which is used to reference some TPM resource with a name and a handle (and contains a few other bits) for commands that operate on PCRs, such as https://pkg.go.dev/github.com/canonical/go-tpm2#TPMContext.PCRExtend, just like any other command that operates on a TPM resource.

And here's the first octet for the handle of all resource types: https://pkg.go.dev/github.com/canonical/go-tpm2@master#HandleType

efi/image_rules_defs.go
```diff
@@ -50,7 +50,7 @@ func makeMicrosoftUEFICASecureBootNamespaceRules() *secureBootNamespaceRules {
 				0x13, 0xad, 0xbf, 0x43, 0x09, 0xbd, 0x82, 0x70, 0x9c, 0x8c,
 				0xd5, 0x4f, 0x31, 0x6e, 0xd5, 0x22, 0x98, 0x8a, 0x1b, 0xd4,
 			},
-			// pubkey alg
+			// pubkey a;g
```
pedronis, line 53: typo?

efi/pcr_profile.go
```diff
@@ -28,15 +28,36 @@ import (
 	"golang.org/x/xerrors"
 )
 
+// PCRProfileEnablePCRsOption is an option for AddPCRProfile that adds one or more PCRs.
+type PCRProfileEnablePCRsOption interface {
+	PCRProfileOption
+	PCRs() tpm2.HandleList
+}
+
```
pedronis, line 36: nitpick: shouldn't this definition come after the PCRProfileOption one?

efi/fw_load_handler.go
```diff
@@ -101,32 +99,32 @@ func (h *fwLoadHandler) measureSecureBootPolicyPreOS(ctx pcrBranchContext) error
 		events = events[1:]
 
 		switch {
-		case e.PCRIndex < secureBootPCR && e.EventType == tcglog.EventTypeSeparator:
+		case e.PCRIndex < tcglog.PCRIndex(secureBootPolicyPCR) && e.EventType == tcglog.EventTypeSeparator:
```
pedronis, line 102: would it make sense to change the type in Event?

pedronis (approved):
+1

efi/image_rules_defs.go
chrisccoulson: It is

efi/pcr_profile.go
chrisccoulson: I've moved this now

efi/fw_load_handler.go
chrisccoulson: Possibly. I think at one point I tried to implement the tcglog-parser library without a dependency on go-tpm2, with it's own types and constants for digest algorithms as well, but it's had a dependency on it for a while now so it doesn't make much sense to keep the `tcglog.PCRIndex` type. I'll fix that up separately.

PR #312, opened 2024-06-21
efi: change how HostEnvironment overrides EFI variables

chrisccoulson:
The `HostEnvironment` interface contains a `ReadVar` method for abstracting
reads of EFI variables. The default environment just calls
`efi.ReadVariable`, with other implementations providing mocked variables.

The `github.com/canonical/go-efilib` package provides a lot more APIs now
that read from (and some that write to) EFI variables and some of these
will be used from the new `efi/preinstall` package (these APIs are mostly
boot and secure boot related). I wanted a way to be able to override the
environment for the `preinstall.RunChecks` call (particularly as this is
very useful for testing), but the existing `HostEnvironment` approach doesn't
work well for this.

I initially considered a couple of options:
- Make it possible to mock each of the individual calls to go-efilib
  functions via the `HostEnvironment` interface.
- Export `varsBackend` from go-efilib and provide a function in to mock
  the backend by overriding the system one.

The first approach scales poorly, so my initial approach was going to be
the second one. The problem with this though is that `HostEnvironment`
will be part of a public, production API, and mocking the backend is global
to the whole process. I don't think this is appropriate in production code
where it might be possible that there may be one goroutine reading EFI
variables from the system when another goroutine decides to mock the backend
based on a supplied `HostEnvironment`.

I've taken a different approach instead, by modifying every go-efilib
function that interacts with EFI variables to accept a `context.Context`
argument. The context contains a `VarsBackend` keyed from it, and the package
exports a `DefaultVarContext` symbol corresponding to the default system
backend (`efivarfs` if it is detected, or a null backend if it isn't
detected). I've used a context rather than passing the `VarsBackend` directly
because the default efivarfs backend may make use of the optional
cancellation and deadline during writing, as it runs a retry loop that potentially
races with other applications.

This approach permits individual call sites to override the backend to
use for individual functions that interact with EFI variables (not just
`ReadVariable`, `WriteVariable` and `ListVariables`, but there's a whole set
of extra functions related to secure boot and boot configuration that
make use of reading and writing EFI variables).

The `ReadVar` method on the `HostEnvironment` interface has gone, and a new
method that returns a context has been added. The `internal.DefaultEnv`
implementation just returns `efi.DefaultVarContext`.

The `MockVars` type in `internal/efitest` has been updated to implement
`efi.VarsBackend`, and the `MockHostEnvironment` implementation in
`internal/efitest` returns a context that keys to this implementation.

This should hopefully allow `preinstall.RunChecks` to be able to take a
`HostEnvironment` argument and benefit from mocked variables even when
interacting with variables indirectly using higher level functions
exported from go-efilib.


efi/env.go
```diff
@@ -37,8 +38,10 @@ import (
 // consumers of the API can provide a custom mechanism to read EFI variables or parse
 // the TCG event log.
 type HostEnvironment interface {
-	// ReadVar reads the specified EFI variable
-	ReadVar(name string, guid efi.GUID) ([]byte, efi.VariableAttributes, error)
+	// VarContext returns a context containing a VarsBackend, keyed by efi.VarsBackendKey,
+	// for interacting with EFI variables via go-efilib. This context can be passed to any
+	// go-efilib function that interacts with EFI variables.
+	VarContext() context.Context
```
pedronis, line 44: this is a bit a strange API for contexts, if a caller gets a context ti usually should pass it along, and not start with a fresh one, so a more expected API would be:

```
VartContext(parent context.Context) context.Context
```

which would return something derived from parent if parent is not nil, instead of returning a fresh context in all cases

efi/internal/default_env.go
```diff
@@ -28,15 +29,16 @@ import (
...
-	readVar      = efi.ReadVariable
 )
 
 type defaultEnvImpl struct{}
 
-func (e defaultEnvImpl) ReadVar(name string, guid efi.GUID) ([]byte, efi.VariableAttributes, error) {
-	return readVar(name, guid)
+// VarContext implements [HostEnvironment.VarContext].
+func (e defaultEnvImpl) VarContext() context.Context {
+	return efi.DefaultVarContext
```
pedronis, line 38: so this really should really be some form of WithDefaultVarContext from efi

pedronis (commented):
not sure the new APIs fully match expectations about context usage

efi/env.go
chrisccoulson: Funnily enough, I'd already added an API for this, so I've just pulled in the go-efilib version with it and used it in the default implementation.

efi/internal/default_env.go
chrisccoulson: Yes, I've updated this now.

efi/env.go
```diff
@@ -40,8 +40,13 @@ import (
...
 	// VarContext returns a context containing a VarsBackend, keyed by efi.VarsBackendKey,
 	// for interacting with EFI variables via go-efilib. This context can be passed to any
-	// go-efilib function that interacts with EFI variables.
-	VarContext() context.Context
+	// go-efilib function that interacts with EFI variables. Right now, go-efilib doesn't
+	// support any other uses of the context such as cancelation or deadlines. The efivarfs
+	// backend will support this eventually for variable writes because it currently implements
+	// a retry loop that has a potential to race with other processes. Cancelation and / or
+	// deadlines for sections of code that performs multiple reads using efivarfs may be useful
+	// in the future because the kernel rate-limits reads per-user.
```
pedronis, line 48: should the doc say something about parent?

efi/internal/default_env_test.go
```diff
@@ -39,7 +40,11 @@ type defaultEnvSuite struct{}
 var _ = Suite(&defaultEnvSuite{})
 
 func (s *defaultEnvSuite) TestVarContext(c *C) {
-	c.Check(DefaultEnv.VarContext(), Equals, efi.DefaultVarContext)
+	ctx := DefaultEnv.VarContext(context.Background())
+	c.Assert(ctx, NotNil)
+
+	expected := efi.WithDefaultVarsBackend(context.Background())
```
pedronis, line 46: maybe a good idea to pass something with an attached value already and check it can be retrieved?

efi/internal/env.go
```diff
@@ -31,8 +31,13 @@ import (
...
 	// VarContext returns a context containing a VarsBackend, keyed by efi.VarsBackendKey,
 	// for interacting with EFI variables via go-efilib. This context can be passed to any
-	// go-efilib function that interacts with EFI variables.
-	VarContext() context.Context
+	// go-efilib function that interacts with EFI variables. Right now, go-efilib doesn't
+	// support any other uses of the context such as cancelation or deadlines. The efivarfs
+	// backend will support this eventually for variable writes because it currently implements
+	// a retry loop that has a potential to race with other processes. Cancelation and / or
+	// deadlines for sections of code that performs multiple reads using efivarfs may be useful
+	// in the future because the kernel rate-limits reads per-user.
```
pedronis, line 39: same about parent

pedronis (commented):
thanks for the change, couple small remarks

efi/internal/default_env_test.go
chrisccoulson: I've added this to the test

pedronis (approved):
thank you

PR #313, opened 2024-07-10
efi/internal: move to internal/efi

chrisccoulson:
This will allow internal/efi to be used by internal/efitest in a follow-up PR, which will extend HostEnvironment with some methods that are only intended for use by efi/preinstall, and which I don't want to immediately make public - the extended HostEnvironment will be used only in unit tests, as it requires the ability to mock some low-level parts of the system, such as sysfs, cpuid and MSRs and I'm not sure if there will be a need for this to be part of the public API yet.


pedronis (approved):
this looks alright

PR #314, opened 2024-07-10
internal/efi: Add HostEnvironment bits for efi/preinstall

chrisccoulson:
This extends the internal `HostEnvironment` interface to accommodate
`efi/preinstall`, to make it a superset of the one that's already exposed by the
`efi` package. The intention is that this superset won't be exposed for now,
other than for unit testing - the `efi` package still only exposes the existing
subset with methods for overriding EFI variables and the TCG log.

The defaultEnvImpl in has been updated to provide a default implementation of
it, and the `internal/efitest` package has been updated to be able to mock the
new parts of it.


internal/efi/env.go
```diff
@@ -42,3 +49,90 @@ type HostEnvironment interface {
...
+type HostEnvironmentAMD64 interface {
+	// CPUVendorIdentificator returns the CPU vendor.
+	CPUVendorIdentificator() string
+
+	// HasCPUIDFeature returns if feature from FeatureNames map in the
+	// github.com/intel-go/cpuid package is available.
+	HasCPUIDFeature(feature uint64) bool
+
+	// ReadMSRs reads the value of the specified MSR for all CPUs,
+	// returning a map of the result for all CPU numbers
```
pedronis, line 76: maybe "a map of the results  keyed by the CPU numbers" would be clearer here?

internal/efi/default_env.go
```diff
@@ -49,6 +61,146 @@ func (e defaultEnvImpl) ReadEventLog() (*tcglog.Log, error) {
...
+// AttributeReader implements [SysfsDevice.AttributeReader].
+func (d *defaultEnvSysfsDevice) AttributeReader(attr string) (rc io.ReadCloser, err error) {
+	if attr == "uevent" {
+		return nil, ErrNoDeviceAttribute
+	}
+
+	f, err := os.Open(filepath.Join(d.path, attr))
+	switch {
+	case os.IsNotExist(err):
+		return nil, ErrNoDeviceAttribute
```
pedronis, line 145: these error path are no tested

internal/efi/default_env.go
```diff
@@ -49,6 +61,146 @@ func (e defaultEnvImpl) ReadEventLog() (*tcglog.Log, error) {
...
+		}
+		f.Close()
+	}()
+
+	fi, err := f.Stat()
+	if err != nil {
+		return nil, err
+	}
+	if !fi.Mode().IsRegular() {
+		return nil, ErrNoDeviceAttribute
```
pedronis, line 161: same for this

internal/efi/default_env.go
```diff
@@ -49,6 +61,146 @@ func (e defaultEnvImpl) ReadEventLog() (*tcglog.Log, error) {
...
+}
+
+// DeviceForClass implements [HostEnvironment.DevicesForClass].
+func (defaultEnvImpl) DevicesForClass(class string) ([]SysfsDevice, error) {
+	classPath := filepath.Join(sysfsPath, "class", class)
+	f, err := os.Open(classPath)
+	switch {
+	case os.IsNotExist(err):
+		// it's ok to have no devices for the specified class
+		return nil, nil
```
pedronis, line 174: this case is not tested

efi/shim_test.go
```diff
@@ -400,7 +400,7 @@ func (s *shimSuite) TestShimSbatPolicyLatestUnset(c *C) {
 
 	c.Assert(visitor.varModifiers, HasLen, 1)
 
-	collector := NewVariableSetCollector(efitest.NewMockHostEnvironment(nil, nil))
+	collector := NewVariableSetCollector(efitest.NewMockHostEnvironment(efitest.MockVars{}, nil))
```
pedronis, line 403: what's the reason for the need of this change and other similar ones?

pedronis (commented):
thanks, did a pass

efi/shim_test.go
chrisccoulson: Ah, it was because I made a change to `MockHostEnvironment.VarContext` to not attach a `efi.VarsBackend` if the supplied `efitest.MockVars` map is nil, which resulted in anything that ends up calling `efi.ReadVariable` crashing. I reverted the change to how it worked before - `MockHostEnvironment.VarContext` will always add a `efi.VarsBackend` even if the supplied `efitest.MockVars` is nil. This does mean that anything which calls `efi.ReadVariable` ends up reading keys from a nil map, but that seems to be ok in go (it's only adding entries to a nil map that causes a panic)

internal/efi/default_env.go
chrisccoulson: I've added a test for this

internal/efi/default_env.go
chrisccoulson: I've added a test for this

internal/efi/default_env.go
chrisccoulson: I've added a test for this

pedronis (approved):
thank you

PR #315, opened 2024-07-10
preinstall: Add checks for virtualization mode

chrisccoulson:
If a container environment is detected, this always return an error.
Detecting a VM returns a value which will be interpreted by the caller
(`RunChecks`) and will also return an error unless a flag is specified
by snapd to suppress this, in which case the subsequent firmware
protection tests will be skipped because these don't make sense in
virtualized environments.

### DIFF
--- a/efi/preinstall/check_virt.go
+++ b/efi/preinstall/check_virt.go
@@ -39,7 +39,7 @@ func detectVirtualization(env internal_efi.HostEnvironment) (detectVirtResult, e
 	// First check for any virtualization
 	virt, err := env.DetectVirtMode(internal_efi.DetectVirtModeAll)
 	if err != nil {
-		return 0, fmt.Errorf("cannot detect if environment if virtualized: %w", err)
+		return 0, fmt.Errorf("cannot detect if environment is virtualized: %w", err)
 	}
 	if virt == internal_efi.VirtModeNone {
 		// we're not in a containerized or virtualized environment
--- a/efi/preinstall/check_virt.go
+++ b/efi/preinstall/check_virt.go
@@ -64,7 +64,7 @@ func detectVirtualization(env internal_efi.HostEnvironment) (detectVirtResult, e
 		return 0, fmt.Errorf("cannot detect if environment is a VM: %w", err)
 	}
 	if vmVirt != virt {
-		return 0, fmt.Errorf("unexpected return value from HostEnvironment.DetectVirtMode(DetectVirtModeVM) (got:%q, expected:%q)", vmVirt, virt)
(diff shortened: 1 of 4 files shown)


efi/preinstall/check_virt.go
```diff
@@ -0,0 +1,71 @@
...
+	detectVirtVM
+)
+
+// detectVirtualization detects if the supplied environment is virtualized,
+// returning the type of virtualization - container or VM.
+func detectVirtualization(env internal_efi.HostEnvironment) (detectVirtResult, error) {
+	// First check for any virtualization
+	virt, err := env.DetectVirtMode(internal_efi.DetectVirtModeAll)
+	if err != nil {
+		return 0, fmt.Errorf("cannot detect if environment if virtualized: %w", err)
```
pedronis, line 42: typo: 2nd /if/is/

efi/preinstall/check_virt.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	// We're in a virtualized (VM) environment. We expect the following call to return
+	// the same value as our original call here.
+	vmVirt, err := env.DetectVirtMode(internal_efi.DetectVirtModeVM)
+	if err != nil {
+		return 0, fmt.Errorf("cannot detect if environment is a VM: %w", err)
+	}
+	if vmVirt != virt {
+		return 0, fmt.Errorf("unexpected return value from HostEnvironment.DetectVirtMode(DetectVirtModeVM) (got:%q, expected:%q)", vmVirt, virt)
```
pedronis, line 67: s/unexpected/inconsistent/ perhaps?

efi/preinstall/check_virt.go
```diff
@@ -0,0 +1,71 @@
...
+package preinstall
+
+import (
+	"errors"
+	"fmt"
+
+	internal_efi "github.com/snapcore/secboot/internal/efi"
+)
+
+type detectVirtResult int
```
pedronis, line 29: why not reuse the type for internal_efi.DetectVirtModeVM ?

pedronis (commented):
question, and couple smaller comments

efi/preinstall/check_virt.go
chrisccoulson: I've fixed this

efi/preinstall/check_virt.go
chrisccoulson: I've updated this error string

efi/preinstall/check_virt.go
chrisccoulson: I added a distinct type here to omit the container value (`internal_efi.DetectVirtModeContainer`) so that it's explicit to the caller that it can't return this value.

pedronis (approved):
thank you

PR #317, opened 2024-07-22
plainkey: rename platform key to protector key

chrisccoulson:
"Platform" is a bit overloaded in this package, eg, there is a
`secboot.PlatformKeyData` type already and EFI has a "Platform Key" as
well.


pedronis (approved):
thank you

PR #319, opened 2024-07-31
preinstall: Add initial checks for TCG log consistency with PCR values

chrisccoulson:
This adds a function that verifies that the TCG log is well formed and
is consistent with the TPM's PCR values for at least one supported PCR
bank. On success, it returns the best PCR bank to use and the startup
locality, and any errors that occurred for non-mandatory PCRs. A PCR
might not be mandatory because we don't have a way to generate profiles
for it, or because the input flags to `RunChecks` marked it as optional.
The startup locality is used later on with the discrete TPM indication
returned from the earlier TPM check to customize the PCR profile.

It performs some other tests, notably that:
- the TCG log contains a sequence of expected phases.
- the TCG defined PCRs contain a EV_SEPARATOR event between the pre-OS
  and OS-present environment (although the one in PCR7 separates secure
  boot policy from secure boot authentication).
- none of the EV_SEPARATORs in the TCG defined PCRs indicated that an
  error occurred.
- there are no pre-OS measurements to non-TCG defined PCRs (8-).

Future PRs will add more specific checks for each supported PCR.

This pulls in a newer go-tpm2 for `TPM2_PCR_Allocate` support in the
library and test harness. This is a function that's normally ever
called from firmware, but we need to be able to do what the firmware
does to enable/disable PCR banks here for unit testing to emulate
different firmware configurations.

### DIFF
--- a/efi/fw_load_handler_test.go
+++ b/efi/fw_load_handler_test.go
@@ -285,7 +285,8 @@ func (s *fwLoadHandlerSuite) TestMeasureImageStartPlatformFirmwareProfile(c *C) 
 		pcrs:       MakePcrFlags(internal_efi.PlatformFirmwarePCR),
 		expectedEvents: []*mockPcrBranchEvent{
 			{pcr: 0, eventType: mockPcrBranchResetEvent},
-			{pcr: 0, eventType: mockPcrBranchExtendEvent, digest: testutil.DecodeHexString(c, "d0ff5974b6aa52cf562bea5921840c032a860a91a3512f7fe8f768f6bbe005f6")},
+			{pcr: 0, eventType: mockPcrBranchExtendEvent, digest: testutil.DecodeHexString(c, "cd1137dfa2bfba51973100d73d78d9f496e089fd246fe980fadc668b4efc9443")},
+			{pcr: 0, eventType: mockPcrBranchExtendEvent, digest: testutil.DecodeHexString(c, "5ca5e6acb83d626a42b53ddc5a2fe04d6a4b2f045bb07f6d9baf0e82900d7bbe")},
 			{pcr: 0, eventType: mockPcrBranchExtendEvent, digest: testutil.DecodeHexString(c, "aef237d4703e8936530141636186a9f249fa39e194f02f668cd328bd5902cf03")},
 			{pcr: 0, eventType: mockPcrBranchExtendEvent, digest: testutil.DecodeHexString(c, "8b0eec99d3cccc081edb98c3a2aa74b99a02b785bd74513e1cf7401e99121e80")},
 			{pcr: 0, eventType: mockPcrBranchExtendEvent, digest: testutil.DecodeHexString(c, "df3f619804a92fdb4057192dc43dd748ea778adc52bc498ce80524c014b81119")},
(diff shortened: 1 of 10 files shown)


efi/preinstall/check_tcglog.go
```diff
@@ -0,0 +1,610 @@
...
+			// can also be initialized before TPM2_Startup by a H-CRTM event sequence from
+			// locality 4, which will be reflected in this event in this case by setting the
+			// startup locality to 4, regardless of what locality TPM2_Startup is subsequently
+			// called from.
+			if ev.PCRIndex != internal_efi.PlatformFirmwarePCR {
+				// This event should only ever appear in PCR0
+				results.Lookup(ev.PCRIndex).setErr(errors.New("unexpected StartupLocality event (should be in PCR0)"))
+				continue
+			}
+			if !bytes.Equal(results.Lookup(0).logValue, results.Lookup(0).initialValue) {
```
pedronis, line 295: shouldn't this check be a predicate on the struct?

efi/preinstall/check_tcglog.go
```diff
@@ -0,0 +1,610 @@
...
+	numSeparatorsSeen int
+	pcrSeparatorsSeen map[tpm2.Handle]struct{}
+}
+
+// newTcgLogPhaseTracker returns a new tcglogPhaseTracker
+func newTcgLogPhaseTracker() *tcglogPhaseTracker {
+	return &tcglogPhaseTracker{
+		phase:             tcglogPhasePreOSBeforeMeasureSecureBootConfig,
+		numSeparatorsSeen: 0,
+		pcrSeparatorsSeen: make(map[tpm2.Handle]struct{}),
```
pedronis, line 419: nitpick: we tend to just use bool and not empty struct for this kind of set structures

efi/preinstall/check_tcglog.go
```diff
@@ -0,0 +1,610 @@
...
+		default:
+			// No change for any other event
+		}
+	case t.phase == tcglogPhasePreOSMeasuringSecureBootConfig:
+		switch {
+		case ev.PCRIndex != internal_efi.SecureBootPolicyPCR && ev.EventType == tcglog.EventTypeSeparator:
+			// An EV_SEPARATOR event in PCRs 0-6 after measuring the secure boot configuration begins the
+			// transition to OS-present, skipping the part of the pre-OS phase after the secure boot config
+			// has been measured.
+			t.phase = tcglogPhaseTransitioningToOSPresent
```
pedronis, line 497: this case is not tested

efi/preinstall/check_tcglog.go
```diff
@@ -0,0 +1,610 @@
...
+			return nil, err
+		}
+		if phase == tcglogPhaseOSPresent {
+			// We've hit OS-present - we don't care about the rest of the log at
+			// this stage
+			break
+		}
+	}
+	if !phaseTracker.reachedOSPresent() {
+		return nil, errors.New("reached the end of the log without seeing EV_SEPARATOR events in all TCG defined PCRs")
```
pedronis, line 603: this error case is not tested

pedronis (commented):
did an initial pass

efi/preinstall/check_tcglog.go
chrisccoulson: Yes, they could be

efi/preinstall/check_tcglog.go
chrisccoulson: I can change that.

efi/preinstall/check_tcglog.go
chrisccoulson: Aha, I'll see if there's something I can do to test that. At the moment I'm just doing ad-hoc modifications to logs to break things.

Note that the next PR currently has quite a low coverage because it's got a switch statement with a lot of branches to cover all possible event types. We have tests for the most common types that we're likelly to see, but testing all of the others is going to require significant work in `internal/efitest` to customize the log and generate events with different types. I'm not sure whether that will be a blocker for the first iteration, especially as Oliver is keen to get this landed.

efi/preinstall/check_tcglog.go
chrisccoulson: This one should be easy to figure out how to test.

internal/efitest/log.go
```diff
@@ -153,32 +154,67 @@ func NewLog(c *C, opts *LogOptions) *tcglog.Log {
 	}
 
 	// Mock S-CRTM measurements
-	{
-		data := tcglog.StringEventData("1.0")
-		builder.hashLogExtendEvent(c, data, &logEvent{
+	if opts.StartupLocality == 4 {
+		ev := &tcglog.Event{
```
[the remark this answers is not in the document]
chrisccoulson: I've added a comment to this.

efi/fw_load_handler_test.go
```diff
@@ -285,7 +285,8 @@ func (s *fwLoadHandlerSuite) TestMeasureImageStartPlatformFirmwareProfile(c *C)
 		pcrs:       MakePcrFlags(internal_efi.PlatformFirmwarePCR),
 		expectedEvents: []*mockPcrBranchEvent{
 			{pcr: 0, eventType: mockPcrBranchResetEvent},
-			{pcr: 0, eventType: mockPcrBranchExtendEvent, digest: testutil.DecodeHexString(c, "d0ff5974b6aa52cf562bea5921840c032a860a91a3512f7fe8f768f6bbe005f6")},
+			{pcr: 0, eventType: mockPcrBranchExtendEvent, digest: testutil.DecodeHexString(c, "cd1137dfa2bfba51973100d73d78d9f496e089fd246fe980fadc668b4efc9443")},
+			{pcr: 0, eventType: mockPcrBranchExtendEvent, digest: testutil.DecodeHexString(c, "5ca5e6acb83d626a42b53ddc5a2fe04d6a4b2f045bb07f6d9baf0e82900d7bbe")},
```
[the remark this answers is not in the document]
chrisccoulson: They are.

efi/preinstall/errors.go
```diff
@@ -128,3 +131,76 @@ var (
...
+			if err, isErr := pcrErrs[pcr]; isErr {
+				// We have an error for this PCR
+				fmt.Fprintf(w, "- %v(PCR%d): %v.\n", alg, pcr, err)
+			}
+		}
+	}
+	return w.String()
+}
+
+// UNwrapBankError returns the error associated with the specified PCR bank if one
```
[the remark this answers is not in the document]
chrisccoulson: Fixed.

efi/preinstall/check_tcglog.go
chrisccoulson: I've added a test for this one now.

pedronis (approved):
thank you, should Mate or somebody else familiar with boot give a look a this?

chrisccoulson:
> thank you, should Mate or somebody else familiar with boot give a look a this?

Sure, I can ask @kukrimate if he would like to take a look at this too before we land it.

PR #320, opened 2024-08-04
various: update github.com/canonical/tcglog-parser

chrisccoulson:
github.com/canonical/tcglog-parser defined its own types for PCR indides
and digests, which resulted in a bunch of pointless conversions between
these and go-tpm2 equivalent types in packages that depended on both of
these, despite the fact that tcglog-parser already depends on go-tpm2.

I've removed the tcglog-parser type definitions in master and updated it
here which allows us to remove a fair few type conversions.


pedronis (approved):
thank you for the clean up, I think we discussed this some PRs ago

PR #321, opened 2024-08-07
preinstall: Add simple checks for PCRs 0 and 2

chrisccoulson:
(Draft until https://github.com/snapcore/secboot/pull/319 has landed).

This adds some trivial to checks to both PCRs 0 and 2:
- that they contain events of the expected type.
- that there are no measurements to them from the OS-present phase.

As we don't do any sort of prediction for these PCRs (yet, but never say
never), it doesn't perform additional checking such as ensuring that the
event data is consistent with the event digests when the event digests are
a tagged hash of the event data (ie, not purely informational).

The PCR2 check also returns an indication as to whether drivers from
value-added-retailer hardware is running, which `RunChecks` will make it
possible to opt out of.


chrisccoulson:
I'm closing this PR and opting for a simpler approach for now (https://github.com/canonical/secboot/pull/334) so we can get things landed faster.

PR #324, opened 2024-08-13
preinstall: Add checks for PCR4

chrisccoulson:
This adds checks some checks for PCR4.

The caller supplies a `context.Context` to which an EFI variable backend
is attached, a `internal_efi.HostEnvironment` implementation, a TCG log, a
PCR digest algorithm (the optimum for this is computed earlier by
another function that does a more general check of the TCG log) and a
list of boot images related to the current boot. These are used to
verify the digests in the `EV_EFI_BOOT_SERVICES_APPLICATION` events.

It reads the `BootCurrent` EFI variable and matches this to the
`EFI_LOAD_OPTION` associated with the current boot from the TCG log - it
uses the log as `BootXXXX` EFI variables can updated at runtime and might
be out of date when this code runs.

In the pre-OS environment, it checks that the log only contains either
a single `EV_OMIT_BOOT_DEVICE_EVENTS` event or a single `EV_EFI_ACTION`
"Calling EFI Application from Boot Option" event (after which, it expects
to see the `EV_SEPARATOR` event to signal the transition to OS-present). It
also permits `EV_EFI_BOOT_SERVICES_APPLICATION` events in the pre-OS
environment if the `BootOptionSupport` EFI variable indicates that system
preparation applications are supported, although these must be before the
previously mentioned `EV_EFI_ACTION` event. The profile generating code in
efi/fw_load_handler.go copies these pre-OS events (including sysprep
applications to the final policy).

Note that firmware that implements v1.06 of the TCG PC-Client PFP spec
implements another `EV_EFI_ACTION` event that indicates the number of the
Boot option that the load occurred from - this code will need to include
support for that quite soon. There is a corresponding github issue for
this (https://github.com/canonical/secboot/issues/308).

When processing the initial part of the OS-present section of the log,
it expects to find the `EV_EFI_BOOT_SERVICES_APPLICATION` for the initial
boot loader (IBL), and it expects the path of this to match the previously
discovered load option associated with the current boot - note that it
doesn't expect a complete match. See how path matching works in
`efi.DevicePath.Matches`. Any other event type at this stage is rejected as
an error, as it would result in an invalid policy being generated by the
code in efi/fw_load_handler.go. If the first `EV_EFI_BOOT_SERVICES_APPLICATION`
event is not the IBL, it checks to see if it is part of Absolute, in which
case it skips it and looks for the next event. Note that Absolute will be
copied to the computed policy by efi/fw_load_handler.go, if it is present,
although it's recommended that this is disabled instead. The next event
must be the IBL, else an error is returned. It verifies that the digest in
the log matches the Authenticode digest of the first boot file supplied to
the function.

It then continues to process `EV_EFI_BOOT_SERVICE_APPLICATION` events in
the log (permitting, but ignoring other event types - it's expected that
if other OS boot code uses other event types in the log, the policy
generation code for that OS component in the secboot efi package emits
these. For each launch event (`EV_EFI_BOOT_SERVICES_APPLICATION`), it
verifies that the digest in the log matches the Authenticode digest of
the next boot file supplied. Note that it is not mandatory to verify every
one of these events, so it's not mandatory to supply every boot
component to the function.

Some OS components may not rely on the boot services LoadImage API for
loading subsequent stages (eg, shim). For the
`EV_EFI_BOOT_SERVICES_APPLICATION` digests to be correct, these components
rely on the existence of the `EFI_TCG2_PROTOCOL` with the `PE_COFF_IMAGE`
flag. There isn't a direct way to test for this, so for this reason, the
function does have to verify the digest of the
`EV_EFI_BOOT_SERVICES_APPLICATION` event associated with the secondary
boot loader (SBL), if there is one. This image must be supplied to this
function, else an error will be returned.

On success, the function will return a set of flags indicating whether
system preparation applications were detected to be running, whether
Absolute was detected to be running, or whether it wasn't able to verify
all `EV_EFI_BOOT_SERVICES_APPLICATION` event digests because not all paths
to every component for the current boot were supplied. What happens with
these flags will be customizable with flags supplied to the higher level
`RunChecks` API.

### DIFF
--- a/efi/preinstall/check_pcr4.go
+++ b/efi/preinstall/check_pcr4.go
@@ -1,0 +1,451 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2024 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package preinstall
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+
+	efi "github.com/canonical/go-efilib"
+	"github.com/canonical/go-tpm2"
+	"github.com/canonical/tcglog-parser"
+	secboot_efi "github.com/snapcore/secboot/efi"
+	internal_efi "github.com/snapcore/secboot/internal/efi"
+)
+
+var (
+	efiComputePeImageDigest = efi.ComputePeImageDigest
+)
+
+// readLoadOptionFromLog reads the corresponding Boot#### load option from the log,
+// which reflects the value of it at boot time, as opposed to reading it from an
+// EFI variable which may have been modified since booting.
+func readLoadOptionFromLog(log *tcglog.Log, n uint16) (*efi.LoadOption, error) {
+	events := log.Events
+	for len(events) > 0 {
+		ev := events[0]
+		events = events[1:]
+
+		if ev.PCRIndex != internal_efi.PlatformConfigPCR {
+			continue
+		}
+
+		if ev.EventType != tcglog.EventTypeEFIVariableBoot && ev.EventType != tcglog.EventTypeEFIVariableBoot2 {
+			// not a boot variable
+			continue
+		}
+
+		data, ok := ev.Data.(*tcglog.EFIVariableData)
+		if !ok {
+			// decode error data is guaranteed to implement the error interface
+			return nil, fmt.Errorf("boot variable measurement has wrong data format: %w", ev.Data.(error))
+		}
+		if data.VariableName != efi.GlobalVariable {
+			// not a global variable
+			continue
+		}
+		if !strings.HasPrefix(data.UnicodeName, "Boot") || len(data.UnicodeName) != 8 {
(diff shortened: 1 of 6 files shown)


efi/preinstall/check_pcr4.go
```diff
@@ -0,0 +1,451 @@
...
+		}
+
+		data, ok := ev.Data.(*tcglog.EFIVariableData)
+		if !ok {
+			// decode error data is guaranteed to implement the error interface
+			return nil, fmt.Errorf("boot variable measurement has wrong data format: %w", ev.Data.(error))
+		}
+		if data.VariableName != efi.GlobalVariable {
+			// not a global variable
+			continue
```
pedronis, line 66: we don't hit here

efi/preinstall/check_pcr4.go
```diff
@@ -0,0 +1,451 @@
...
+		}
+		if !strings.HasPrefix(data.UnicodeName, "Boot") || len(data.UnicodeName) != 8 {
+			// name has unexpected prefix or length
+			continue
+		}
+
+		var x uint16
+		if y, err := fmt.Sscanf(data.UnicodeName, "Boot%x", &x); err != nil || y != 1 {
```
pedronis, line 74: y is a bit of strange name here, as x and y are not the same kind of index. maybe s/y/c/ for count ?

efi/preinstall/check_pcr4.go
```diff
@@ -0,0 +1,451 @@
...
+
+			if ev.EventType != tcglog.EventTypeEFIBootServicesApplication {
+				// Only care about EV_EFI_BOOT_SERVICES_APPLICATION events for checking
+				if seenOSComponentLaunches == 0 {
+					// The only events we're expecting in OS-present for now is EV_EFI_BOOT_SERVICES_APPLICATION.
+					return 0, fmt.Errorf("unexpected OS-present log event type %v (expected EV_EFI_BOOT_SERVICES_APPLICATION)", ev.EventType)
+				}
+				// Once the IBL has launched, other event types are acceptable as long as the policy generation
+				// code associated with the component in the secboot efi package emits them.
+				continue NextEvent
```
pedronis, line 348: we never reach this case in the tests

pedronis (approved):
did a pass

PR #328, opened 2024-09-03
argon2: Add helpers for running the KDF remotely

chrisccoulson:
As Argon2 is memory intensive, it's not suitable for multiple invocations in long-lived
garbage collected processes where multiple invocations may result in the process being
unable to perform new allocations, or even worse, triggering the kernel’s OOM killer.
For this reason, Argon2 is currently abstracted with an interface (`Argon2KDF`), of which
the application sets a global version of this which is intended to proxy KDF requests to
a short-lived remote process which uses the real `InProcessArgon2KDF`.

This adds some functionality to facilitate this.

Then there are JSON serializable types `Argon2OutOfProcessRequest` and
`Argon2OutOfProcessResponse`. The request can be fed directly to `RunArgon2OutOfProcessRequest`
on the remote side, but this is a fairly low-level API - the application still has to
deal with the transport.

There is a higher level API - `NewOutOfProcessArgon2KDF`, for use in the application process,
and which returns an implementation of `Argon2KDF` which proxies requests to a short-lived
remote handler process. The caller supplies a function to construct an appropriate `exec.Cmd`
instance for this. This function configures the `exec.Cmd` so that the handler process receives
a request on stdin and it expects a response on stdout. The handler process is expected to pass
both `os.Stdin` and `os.Stdout` to `WaitForAndRunArgon2OutOfProcessRequest`, although it doesn't
hardcode these descriptors for implementations that want to construct their own processes with
transports that don't rely on stdin and stdout.

Once a handler process has completed a request, it should exit cleanly. Neither
`RunArgon2OutOfProcessRequest` or `WaitForAndRunArgon2OutOfProcessRequest` support being called
more than once in the same process.

The code in cmd/run_argon2 provides an example handler process, although this is mainly useful for
unit testing (where it is currently used). It is envisaged that the handler process will be a
special mode of snapd and snap-bootstrap in order to avoid adding an additional go binary just for
this.


argon2_remote_support.go
```diff
@@ -0,0 +1,460 @@
...
+	"fmt"
+	"io"
+	"os/exec"
+	"sync/atomic"
+	"time"
+
+	"github.com/snapcore/secboot/internal/argon2"
+)
+
+type nullInProcessArgon2KDFImpl struct{}
```
pedronis, line 35: null is a bit of a strange name for something that actually fails

argon2_remote_support.go
```diff
@@ -0,0 +1,460 @@
...
+		return nil, fmt.Errorf("cannot create stdout pipe: %w", err)
+	}
+
+	if err := cmd.Start(); err != nil {
+		return nil, fmt.Errorf("cannot start remote process: %w", err)
+	}
+	defer func() {
+		// Run Cmd.Wait in a defer so that we shut down on error paths too,
+		// and we capture the Wait error if there was no other error.
+		waitErr := cmd.Wait()
```
pedronis, line 387: do we need a timeout mechanism around this?

argon2_remote_support.go
```diff
@@ -0,0 +1,460 @@
...
+	}
+}
+
+// WaitAndRunArgon2RequestInRemoteProcess waits for a [Argon2RemoteInput] request on the
+// supplied io.Reader before running it and sending a [Argon2RemoteOutput] response back via
+// the supplied io.Writer. These will generally be connected to the process's os.Stdin and
+// os.Stdout when using - certainly when using [NewRemoteArgon2KDF] on the parent side.
+// This function can only be called once in a process. Subsequent calls in the same
+// process will result in an error response being returned.
+func WaitAndRunArgon2RequestInRemoteProcess(in io.Reader, out io.Writer) error {
```
pedronis, line 338: this function code does not seem to be exercised

export_test.go
```diff
@@ -148,6 +150,11 @@ func MockHashAlgAvailable() (restore func()) {
 	}
 }
 
+func ClearIsArgon2RemoteProcess() {
+	atomic.StoreUint32(&argon2RemoteProcessStatus, notArgon2RemoteProcess)
+	runtime.GC()
```
pedronis, line 155: why the GC here?

argon2_remote_support.go
```diff
@@ -0,0 +1,460 @@
...
+	}
+	if costParams.Threads == 0 {
+		return &Argon2RemoteOutput{
+			Command:     input.Command,
+			ErrorType:   Argon2RemoteErrorInvalidThreads,
+			ErrorString: "invalid threads: cannot be zero",
+		}
+	}
+
+	switch input.Command {
```
pedronis, line 273: maybe a bit more of the error cases below should be tested

pedronis (commented):
did a first pass

pedronis:
@chrisccoulson Also OutOfProcess might be a clearer term than Remote, if a bit longer

argon2_remote_support.go
chrisccoulson: This type has gone, although there is still a `nullArgon2KDFImpl` that existed previously.

argon2_remote_support.go
chrisccoulson: It gets executed in a sub-process (`cmd/run_argon2`) by any test that makes use of `NewOutOfProcessArgon2KDF`, which might be why it doesn't appear in the test coverage.

export_test.go
chrisccoulson: I've added a comment to this function to explain why.

chrisccoulson:
> @chrisccoulson Also OutOfProcess might be a clearer term than Remote, if a bit longer

I've renamed things to `OutOfProcess`.

argon2_remote_support.go
chrisccoulson: I've added some more test cases.

chrisccoulson:
> thanks for the changes, maybe I missed it but it doesn't seem you replied to my question about what to do if the subprocess takes way too long/gets stuck for some reason

Aha, I'll think of a way to address that. As the process is meant to block for some time, I wonder whether creating some sort of regular watchdog or ping would be more appropriate here. It shouldn't be that hard to add to the protocol.

argon2_remote_support.go
chrisccoulson: I've refactored this quite a bit now. Not only is there a timeout mechanism when waiting for the remote process to exit (and go 1.20 introduces a nicer way to do this), there is also an optional watchdog mechanism that exercises the command processing / response sending code on the remote side, so that the parent knows that it's eventually going to get a response and can terminate the operation if this fails.

export_test.go
chrisccoulson: I've removed this function (and the corresponding `SetIsArgon2RemoteProcess` entirely now.

argon2.go
```diff
@@ -119,14 +130,15 @@ func (o *Argon2Options) kdfParams(keyLen uint32) (*kdfParams, error) {
 
 	mode := o.Mode
 	if mode == Argon2Default {
+		//
```
pedronis, line 133: forgot the comment?

argon2_out_of_process_support_sync.go
```diff
@@ -0,0 +1,247 @@
...
+
+		// Attempt to acquire an exclusive, non-blocking, advisory lock.
+		if err := unix.Flock(int(lockFile.Fd()), unix.LOCK_EX|unix.LOCK_NB); err != nil {
+			// We failed to acquire the lock.
+			if os.IsTimeout(err) {
+				// The EWOULDBLOCK case. Someone else already has a lock on the
+				// file we have opened. Try again with a 10ms backoff time,
+				// unless the request timeout fires first.
+				select {
+				case <-time.NewTimer(10 * time.Millisecond).C:
```
pedronis, line 152: isn't this a bit too aggressive? how long does it take usually to compute the actual KDF?

argon2_out_of_process_support_sync.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// Run a loop to try to acquire the lock.
+	for {
+		if triedOnce {
+			// If the loop has executed at least once, make sure that
+			// the timeout hasn't expired.
+			select {
+			case <-timeoutTimer.C:
+				// The timeout has expired.
+				return nil, errArgon2OutOfProcessHandlerSystemLockTimeout
```
pedronis, line 112: this case is not reached by tests

argon2_out_of_process_support.go
```diff
@@ -0,0 +1,999 @@
...
+				// Treat everything else as a KDF request. We don't actually check the
+				// command here - RunArgon2OutOfProcessRequest does this already and will
+				// return an error response if the command is invalid.
+
+				// Spin up a new routine to handle the request, as it blocks and is long running,
+				// and we still want to be able to service watchdog requests on this routine whilst
+				// it's running. Block the current routine until we know the new routine has started
+				// so that the watchdog handler will fail if the new routine doesn't begin properly.
+				var startupWg sync.WaitGroup
+				startupWg.Add(1) // Mark the WaitGroup as waiting for a single event.
```
pedronis, line 544: any reason not to use a channel for this?

pedronis (commented):
did a pass, there is quite a few moving parts here, probably a good idea for @valentindavid @bboozzoo and @ZeyadYasser to take a look for awareness and possibly review as well

argon2_out_of_process_support_sync.go
chrisccoulson: It's a bit hit and miss whether the timeout on line 154 gets tested, or this one gets tested - it depends where the code lands before the timer expires. I'm not sure of a better way of doing it but I'll try and think of a way to make it a bit more reproducible.

argon2_out_of_process_support.go
chrisccoulson: I don't mind, I just picked `sync.WaitGroup` because it seems like it's designed for this use case.

argon2_out_of_process_support_sync.go
chrisccoulson: At this point, we have no idea how long it will be before we can acquire the lock, so it's a bit of a trade-off between being aggressive and waiting longer than is necessary, where the lock ends up unheld for longer for a period. I'm not sure what time to use for the backoff here tbh - the other option is to make the call to `flock` blocking, but then there's no way for it to be able to timeout.

argon2.go
chrisccoulson: I did. I've added a comment now.

argon2_out_of_process_support.go
chrisccoulson: I've changed this to a channel now.

argon2_out_of_process_support_sync.go
chrisccoulson: I've refactored this code to make it more reproducible where the code lands during tests.

argon2_out_of_process_support.go
```diff
@@ -0,0 +1,990 @@
...
+}
+
+// Error implements the error interface.
+func (e *Argon2OutOfProcessError) Error() string {
+	str := "cannot process request: " + string(e.ErrorType)
+	if e.ErrorString != "" {
+		str += " ("
+		str += e.ErrorString
+		str += ")"
+	}
```
bboozzoo, line 137: you can use strings.Builder(), it's available from 1.10+

argon2_out_of_process_support.go
```diff
@@ -0,0 +1,990 @@
...
+	if err := tmb.Wait(); err != nil {
+		// Don't wrap this error - this will be the first non-nil error passed
+		// to Tomb.Kill. There's no benefit to adding additional context here.
+		return nil, err
+	}
+
+	return actualRsp, nil
+}
+
+func (k *outOfProcessArgon2KDFImpl) Derive(passphrase string, salt []byte, mode Argon2Mode, params *Argon2CostParams, keyLen uint32) (key []byte, err error) {
```
bboozzoo, line 911: I know it's part of the existing interface, but if params is always expected to be non nil, this could pass value not a pointer.

argon2_out_of_process_support.go
```diff
@@ -0,0 +1,990 @@
...
+	if err != nil {
+		// This doesn't fail once the OS pipe is created, so there's no
+		// cleanup to do on failure paths other than closing the stdinPipe
+		stdinPipe.Close()
+		return nil, fmt.Errorf("cannot create stdout pipe: %w", err)
+	}
+
+	// Start the remote process.
+	if err := cmd.Start(); err != nil {
+		return nil, fmt.Errorf("cannot start handler process: %w", err)
```
bboozzoo, line 734: do the pipes need to be closed?

argon2_out_of_process_support.go
```diff
@@ -0,0 +1,990 @@
...
+			select {
+			case <-exitWaitCh:
+				// The command closed cleanly, so there's nothing to do.
+				return tomb.ErrDying
+			case <-time.NewTimer(5 * time.Second).C:
+				// We've waited 5 seconds - kill the child process instead. Go 1.20
+				// has a new feature (WaitDelay) which might make things a bit better
+				// here because I don't know how racey things are here - exec.Cmd is
+				// quite complicated.
+				if err := cmd.Process.Kill(); err != nil {
```
bboozzoo, line 758: Depends on how much we trust the caller to have implemented things correctly, but if they implemented the runner their own binary it may be worth to at least leave a comment somewhere that the command should likely be run in a new process group and kill should be done through syscall.Kill(-pgid, syscall.SIGKILL). In which case, some part of the code should set `Setpgid: true` in `cmd.SysProcAttr`.

argon2_out_of_process_support.go
```diff
@@ -0,0 +1,990 @@
...
+		// long as the tomb remains alive.
+		for tmb.Alive() {
+			// Wait for a request from the io.Reader. The only way to unblock this is
+			// if the parent sends something or closes its end of the OS pipe. If it's
+			// closed before we've received a KDF request, then we terminate this routine
+			// with an error to begin the process of the tomb dying and the function
+			// eventually returning an error to the caller.
+			var req *Argon2OutOfProcessRequest
+			dec := json.NewDecoder(in)
+			dec.DisallowUnknownFields()
```
bboozzoo, line 481: means that the both ends need to be in sync now, which hopefully is true most of the time, but should there be some `format`, `version` field in requests/responses to indicate protocol version?

Just to be clear, I would not want to introduce any form to a handshake, as this would complicate the code even more, but a clear version check would help with debugging.

argon2_out_of_process_support.go
```diff
@@ -0,0 +1,990 @@
...
+				// return an error response if the command is invalid.
+
+				// Spin up a new routine to handle the request, as it blocks and is long running,
+				// and we still want to be able to service watchdog requests on this routine whilst
+				// it's running. Block the current routine until we know the new routine has started
+				// so that the watchdog handler will fail if the new routine doesn't begin properly.
+				startupCh := make(chan struct{})
+
+				tmb.Go(func() error {
+					startupCh <- struct{}{} // Unblock the main routine.
```
bboozzoo, line 545: should have the same effect
```suggestion
					close(startupCh)
```

argon2_out_of_process_support.go
```diff
@@ -0,0 +1,990 @@
...
+		select {
+		case reqChan <- req: // Unbuffered channel, but read end is always there unless the tomb is dying.
+		case <-tmb.Dying():
+			// The tomb has begun dying before we had a chance to send the initial request.
+			return tomb.ErrDying
+		}
+
+		// Spin up another routine to run the watchdog
+		tmb.Go(func() error {
+			err := k.watchdog(tmb, reqChan, wdRspChan)
```
bboozzoo, line 805: Watchdog adds quite a bit of complexity. I don't know the spec, but what are we trying to guard against here? if the watchdog handler needs to run a goroutine it doesn't prove anything more than that the child process is still running, but we would get that information from a process dying on its own. Maybe we could split the PR into a part without watchdog and then one adding it separately?

argon2_out_of_process_support.go
chrisccoulson: They do. Both of them get closed at the end of the function with comments explaining that it's required to unblock the remote side so that it can shut down cleanly.

bboozzoo: What I mean is that the pipes were created for stdin/stdout by now. However, the code is exiting early here and there was no defer closing them added before. Actually, the Go documentation isn't exactly clear on what happens. Looking at source code for 1.23.4 the pipes should be closed automatically if Start() failed? https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/os/exec/exec.go;l=631-635 makes me wonder if this behavior is consistent with earlier releases.

argon2_out_of_process_support.go
chrisccoulson: It tests the loop in the remote process that encodes and sends responses back to the parent, so we have some confidence that we're going to receive a response back eventually (considering that this will be programmed to run for ~2s by default) - as long as we don't hit a bug in the crypto library that causes things to stall, of course - there's not much we can do for that. There might be in the future, eg, I'd like to make it interruptible (so you could pass around a `context.Context`), which would almost certainly mean forking the existing library. It wouldn't be difficult to work on, given that it's well specced. It's possible in this case that we could extend the API to add a progress channel to it, which would mean the watchdog could actually directly test that the crypto part is still executing.

argon2_out_of_process_support.go
chrisccoulson: I've made this change now.

argon2_out_of_process_support.go
chrisccoulson: Ooh, good catch - yes, I'm not sure what happens in the case where `Start` fails. I'll take a look at go 1.18 and see if the behaviour is the same.

argon2.go
```diff
@@ -157,12 +169,15 @@ func (o *Argon2Options) kdfParams(keyLen uint32) (*kdfParams, error) {
...
 	default:
+		// The benchmarked path, where we determing what cost paramters to
+		// use in order to obtain the desired execution time.
 		benchmarkParams := &argon2.BenchmarkParams{
 			MaxMemoryCostKiB: 1 * 1024 * 1024, // the default maximum memory cost is 1GiB.
 			TargetDuration:   2 * time.Second, // the default target duration is 2s.
 		}
 
 		if o.MemoryKiB != 0 {
+			// The memory cost has been spcified expli
```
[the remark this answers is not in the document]
chrisccoulson: I've fixed this.

argon2_out_of_process_support.go
chrisccoulson: That seems reasonable, but given that this interface has been around for a while, shall we leave this one for another PR?

argon2_out_of_process_support.go
chrisccoulson: Having a look at go 1.18 and the latest stable, it looks like both ends of each pipes we created get closed by `exec.Cmd.Start` if it fails, which is good. Where there was a bug in this code though is that if we get an error back from `exec.Cmd.StdoutPipe` after already calling `exec.Cmd.StdinPipe`, we closed only the parent side of the pipe. I've fixed this to make sure that both ends get closed now.

argon2_out_of_process_support.go
chrisccoulson: I've removed the calls to `DisallowUnknownFields` everywhere it was used.

argon2_out_of_process_support.go
chrisccoulson: I'm using that now.

argon2_out_of_process_support.go
```diff
@@ -0,0 +1,1042 @@
...
+
+// RunArgon2OutOfProcessRequest runs the specified Argon2 request, and returns a response.
+//
+// In general, this is intended to be executed once in a short-lived process, before the process
+// is discarded. It could be executed more than once in the same process, as long as the caller
+// takes steps to ensure that memory consumed by previous calls has been reclaimed by the GC
+// before calling this function again, but this isn't advised.
+//
+// Note that Argon2 requests are serialized using a system-wide lock, which this function does not
+// explicitly release. If the lock is acquired, it returns a callback that the caller may choose
```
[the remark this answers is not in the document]
bboozzoo: per discussion with @chrisccoulson Conceptually follows the same approach as cryptsetup, trying to keep the lock as close to the process as possible.

argon2_out_of_process_support.go
```diff
@@ -0,0 +1,1042 @@
...
+		return &Argon2OutOfProcessResponse{
+			Command:     request.Command,
+			ErrorType:   Argon2OutOfProcessErrorInvalidMode,
+			ErrorString: fmt.Sprintf("mode cannot be %q", string(request.Mode)),
+		}, nil
+	}
+
+	costParams := &Argon2CostParams{
+		Time:      request.Time,
+		MemoryKiB: request.MemoryKiB,
```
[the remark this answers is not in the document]
bboozzoo: per discussion with @chrisccoulson 0 is valid, the code will use a minimum of 32kB.

@chrisccoulson Can you add a comment?

argon2_out_of_process_support.go
```diff
@@ -0,0 +1,1042 @@
...
+				ErrorType:   Argon2OutOfProcessErrorUnexpected,
+				ErrorString: fmt.Sprintf("cannot run time command: %v", err),
+			}, lockRelease
+		}
+		return &Argon2OutOfProcessResponse{
+			Command:  request.Command,
+			Duration: duration,
+		}, lockRelease
+	default:
+		panic("not reachable")
```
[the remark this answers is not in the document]
bboozzoo: this is checked earlier

argon2_out_of_process_support.go
bboozzoo: Also ok to leave a comment that we only kill the main process, hoping that whoever implemented the thing does not do anything silly like spawning other child processes.

argon2_out_of_process_support.go
bboozzoo: per discussion with @chrisccoulson there's another angle here. the tomb's packge Go() function does not handle panics, and so it's possible that the key derivation could panic, tomb will exit, but the watchdog will carry on. The kdf call wrapper likely needs to recover from a panic and return an error.

argon2_out_of_process_support_sync.go
```diff
@@ -0,0 +1,262 @@
...
+		// happen inside of an exclusive lock without breaking the locking
+		// contract anyway.
+		if lockFile != nil {
+			if err := lockFile.Close(); err != nil {
+				return nil, fmt.Errorf("cannot close lock file from previous attempt before starting new attempt: %w", err)
+			}
+		}
+
+		// Attempt to open the lock file for writing.
+		lockFile, err = os.OpenFile(paths.Argon2OutOfProcessHandlerSystemLockPath, os.O_RDWR|os.O_CREATE, 0600)
```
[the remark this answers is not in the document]
bboozzoo: CLOEXEC may be set by default, let's double check

internal/paths/paths.go
```diff
@@ -19,4 +19,13 @@
...
-var RunDir = "/run"
+import "path/filepath"
+
+var (
+	RunDir = "/run"
+
+	// Argon2OutOfProcessHandlerSystemLockPath is the lock file path used to
+	// serialize KDF requests system-wide. All process's that use the system-wide
+	// lock participate in the lock/unlock contract described above.
+	Argon2OutOfProcessHandlerSystemLockPath = filepath.Join(RunDir, "snapd/argon2.lock")
```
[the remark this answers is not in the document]
bboozzoo: per discussion with @chrisccoulson we'll use a snapd agnostic path, or consider making the lock file location an argument

argon2_out_of_process_support.go
chrisccoulson: I've added a comment here now.

argon2_out_of_process_support.go
chrisccoulson: I've made the KDF goroutine recover from panics so that it can return an error now.

argon2_out_of_process_support.go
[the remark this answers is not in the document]
chrisccoulson: I've added a comment explaining why we don't test `MemoryKiB` here.

argon2_out_of_process_support.go
```diff
@@ -0,0 +1,1042 @@
...
+// as an argument to this function. The [Argon2OutOfProcessResponse] type has a
+// WatchdogResponse field (which the response of this function is used for). It's up
+// to the implementation how they choose to use these fields.
+//
+// If the implementation returns an error, it begins the shutdown of the processing
+// of commands and the eventual return of [WaitForAndRunArgon2OutOfProcessRequest].
+//
+// The implementation is expected to be paired with an equivalent implementation of
+// [Argon2OutOfProcessWatchdogMonitor] in the parent process.
+type Argon2OutOfProcessWatchdogHandler = func(challenge []byte) (response []byte, err error)
```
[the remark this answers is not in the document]
chrisccoulson: These are no longer type aliases

argon2_out_of_process_support_sync.go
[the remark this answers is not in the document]
chrisccoulson: I've added O_NOFOLLOW now, and verified that O_CLOEXEC is set b the go runtime.

argon2_out_of_process_support_sync.go
```diff
@@ -0,0 +1,262 @@
...
+		if err != nil {
+			// No error is expected here.
+			return nil, fmt.Errorf("cannot open lock file for writing: %w", err)
+		}
+
+		// Grab information about the lock file we just opened, via its descriptor.
+		var lockFileSt unix.Stat_t
+		if err := unix.Fstat(int(lockFile.Fd()), &lockFileSt); err != nil {
```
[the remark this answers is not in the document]
chrisccoulson: I've added a test that this is a regular file now.

argon2_out_of_process_support.go
```diff
@@ -0,0 +1,1042 @@
...
+// use of other go routines, noting that the watchdog handler should test that the input request and
+// output response processing continues to function. [WaitForAndRunArgon2OutOfProcessRequest] already
+// does this correctly, and most implementations should just use this.
+//
+// Unfortunately, there is no way to interrupt this function once the key derivation is in progress,
+// because the low-level crypto library does not support this. This feature may be desired in the
+// future, which might require replacing the existing library we use for Argon2.
+func RunArgon2OutOfProcessRequest(request *Argon2OutOfProcessRequest) (response *Argon2OutOfProcessResponse, lockRelease func()) {
+	// Perform checks of arguments that are common to call requests
+	switch request.Mode {
```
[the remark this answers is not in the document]
chrisccoulson: I might look at this in a follow-up

chrisccoulson:
This is ready for (hopefully the final 🙂) review again, although the unit tests currently fail because https://github.com/canonical/secboot/pull/363 needs to land

bboozzoo (approved):
LGTM

pedronis (approved):
thank you, one comment about the lock filename

nitpicks for possibly later:
* I think remote is strange terminology for this, I'm happy we fixed the code name, but the comments should probably use external/separate more instead of remote
* maybe the _support.go file should be split into invocation and external code, with common bits in the invocation .go file

internal/paths/paths.go
```diff
@@ -19,4 +19,13 @@
...
-var RunDir = "/run"
+import "path/filepath"
+
+var (
+	RunDir = "/run"
+
+	// Argon2OutOfProcessHandlerSystemLockPath is the lock file path used to
+	// serialize KDF requests system-wide. All process's that use the system-wide
+	// lock participate in the lock/unlock contract described above.
+	Argon2OutOfProcessHandlerSystemLockPath = filepath.Join(RunDir, "secboot_argon2.lock")
```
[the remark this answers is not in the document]
chrisccoulson: That's fine with me - I've pushed an update to rename this.

PR #332, opened 2024-09-19
crypt: Remove the Model field from ActivateVolumeOptions

chrisccoulson:
This field is not used for generation 2 key metadata where the
checking is happening in the platform-specific code transparently
to this API (snap-bootstrap is expected to call `bootscope.SetModel`
before activating devices).

To simplify how snap-bootstrap uses this API so that it doesn't have
to call it differently depending on whether it is dealing with a gen 1
vs gen 2 key, use the same model that is passed by snap-bootstrap to
`bootscope.SetModel` for gen 1 keys.

This involves creating a small `internal/bootscope` support package
which is consumed by `bootscope` and the core package in order to
avoid a circular dependency.


pedronis:
@valentindavid for your awareness

pedronis (approved):
thanks, this is ab it messier than I would like but  I don't have immediate suggestions to avoid that

PR #334, opened 2024-09-20
preinstall: Add minimal checks for PCR 2

chrisccoulson:
This adds a minimal check for PCR 2 which determines whether any
code from value-added-retailer components is executing or not. This
will be used by `RunChecks` because there will be an option to reject
systems that include code from value-added-retailer components.

As `efi.WithDriversAndAppsProfile` just copies events from the log and
does no prediction for this PCR, this function doesn't do any more
extensive testing, such as ensuring that the PCR only contains events
of the expected type, the event data for each event is of an expected
type, and for events where the digest is a tagged hash of the event
data, that the digest is consistent with the event data.

### DIFF
--- a/efi/preinstall/check_pcr2.go
+++ b/efi/preinstall/check_pcr2.go
@@ -1,0 +1,73 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2024 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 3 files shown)


efi/preinstall/check_pcr2.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+
+		switch ev.EventType {
+		case tcglog.EventTypeEFIBootServicesApplication, tcglog.EventTypeEFIBootServicesDriver, tcglog.EventTypeEFIRuntimeServicesDriver,
+			tcglog.EventTypeEFIPlatformFirmwareBlob, tcglog.EventTypeEFIPlatformFirmwareBlob2, tcglog.EventTypeEFISPDMFirmwareBlob:
+			return driversAndAppsPresent, nil
+		}
+	}
+
+	panic("not reached")
```
pedronis, line 72: couldn't we hit here if for some reason the phase never changed?

pedronis (commented):
question

efi/preinstall/check_pcr2.go
chrisccoulson: In isolation, yes, although the code in `check_tcglog.go` would have already picked this up by the time this code runs and we'll have bailed out. I could add a comment here, or just return an error just in case.

chrisccoulson: I think I'll just add an error in this case, although it should be one that we'll never hit in production code once everything it put together.

pedronis (approved):
thank you

PR #336, opened 2024-10-02
go.mod: switch to canonical/cpuid

valentindavid:


chrisccoulson (approved):
This looks fine to me - would this be the same approach you'd take in snapd for a forked project like this (updating the import paths rather than having the `replace` stanza in `go.mod`? @pedronis

pedronis:
@chrisccoulson we have used replace in the past but more for things where the upstream is legitimate, we just want a bit more of safety and control, but if the upstream github org is potentially not legitimate anymore then maybe is better indeed to just switch things and not just use replace?

PR #337, opened 2024-10-04
preinstall: Add checks for PCR7

chrisccoulson:
This adds some checks for PCR7.

The caller supplies a `context.Context` to which an EFI variable backend
is attached, a `internal_efi.HostEnvironment` implementation, a TCG log, a
PCR digest algorith (the optimum for this is computed earlier by
another function that does a more general check of the TCG log) and the
image of the initial boot loader for the current boot.

As we're testing the firmware, this only checks the log up to the launch
of the initial boot loader. It ignores events after this, as these are
under the control of the OS and don't rely on any special firmware
features such as `EFI_TCG2_PROTOCOL` + `PE_COFF_IMAGE` vs `EFI_TCG_PROTOCOL`
in the same way that OS components measuring to PCR4 do.

This ensures that secure boot is enabled, else an error is returned, as
`WithSecureBootPolicyProfile` only generates profiles compatible with
secure boot being enabled.

If the version of UEFI is >= 2.5, it also makes sure that the secure
boot mode is "deployed mode". If the secure boot mode is "user mode",
then the "AuditMode" and "DeployedMode" values are measured to PCR7,
something that `WithSecureBootPolicyProfile` doesn't support today.
Support for "user mode" will be added in the future, although the public
`RunChecks` API will probably require a flag to opt in to supporting user
mode, as it is the less secure mode of the 2 (see the documentation for
`SecureBootMode` in github.com/canonical/go-efilib).

It also reads the "OsIndicationsSupported" variable to test for features
that are not supported by `WithSecureBootPolicyProfile`. These are
timestamp revocation (which requires an extra signature database -
"dbt") and OS recovery (which requires an extra signature database -
"dbr", used to control access to OsRecoveryOrder and OsRecover####
variables). Of the 2, it's likely that we might need to add support for
timestamp revocation at some point.

It reads the "BootCurrent" EFI variable and matches this to the
`EFI_LOAD_OPTION` associated with the current boot from the TCG log - it
uses the log as "BootXXXX" EFI variables can be updated at runtime and
might be out of data when this code runs. It uses this to detect the
launch of the initial boot loader, which might not necessarily be the
first `EV_EFI_BOOT_SERVICES_APPLICATION` event in the OS-present
environment in PCR4 (eg, if Absolute is active).

First of all, it iterates over the secure boot configuration in the log,
making sure that the configuration is measured in the correct order,
that the event data is valid, and that the measured digest is the tagged
hash of the event data. It makes sure that the value of "SecureBoot" in
the log is consistent with the "SecureBoot" variable (which is read-only
at runtime), and it verifies that all of the signature databases are
formatted correctly. It will return an error if any of these checks
fail.

If the pre-OS environment contains events other than
`EV_EFI_VARIABLE_DRIVER_CONFIG`, it will return an error. This can happen
if a firmware debugger is enabled, in which case PCR7 will begin with a
`EV_EFI_ACTION` "UEFI Debug Mode" event. This case is detected by earlier
firmware protection checks.

If not all of the expected secure boot configuration is measured, an
error is returned.

Once the secure boot configuration has been measured, it looks for
`EV_EFI_VARIABLE_AUTHORITY` events in PCR7, until it detects the launch of
the initial boot loader. It verifies that each of these come from db,
and if the log is in the OS-present environment, it ensures that the
measured digest is the tagged hash of the event data. It doesn't do this
for events in the pre-OS environment because `WithSecureBootPolicyProfile`
just copies these to the profile. It verifies that the firmware doesn't
measure a signature more than once. For each `EV_EFI_VARIABLE_AUTHORITY`
event, it also matches the measured signature to a `EFI_SIGNATURE_LIST`
structure in db. If the matched ESL is a X.509 certificate, it records
the use of this CA in the return value. If the CA is an RSA certificate
with a public modulus of < 256 bytes, it sets a flag in the return
value indicating a weak algorithm. If the matched ESL is a Authenticode
digest, it sets a flag in the return value indicating that pre-OS
components were verified using digests rather than signatures. This
makes PCR7 fragile wrt firmware updates, because it means db needs to
be updated to reflect the new components each time. If the digest being
matched is SHA-1, it sets the flag in the return value indicating a weak
algorithm. If any of these checks fail, an error is returned. If an
event type other than `EV_EFI_VARIABLE_AUTHORITY` is detected, an error is
returned.

Upon detecting the launch of the initial boot loader in PCR4, it
extracts the authenticode signatures from the supplied image, and
matches these to a previously measured CA. If no match is found, an
error is returned. If a match is found, it ensures that the signing
certificate has an RSA public key with a modulus at least as large as
256 bytes, else it sets a flag in the return value indicating a weak algorithm.
Once the event for the initial boot loader is complete, the function
returns.

If the end of the log is reached without encountering the launch of the
initial boot loader, an error is returned.


efi/preinstall/check_pcr7.go
```diff
@@ -0,0 +1,529 @@
...
+
+				// Compute the expected digest from the event data in the log and make
+				// sure it's consistent with the measured digest.
+				expectedDigest := tcglog.ComputeEFIVariableDataDigest(pcrAlg.GetHash(), data.UnicodeName, data.VariableName, data.VariableData)
+				if !bytes.Equal(ev.Digests[pcrAlg], expectedDigest) {
+					return nil, fmt.Errorf("event data inconsistent with measured digest for EV_EFI_VARIABLE_DRIVER_CONFIG event (name:%q, GUID:%v, expected digest:%#x, measured digest:%#x)",
+						data.UnicodeName, data.VariableName, expectedDigest, ev.Digests[pcrAlg])
+				}
+
+				switch data.UnicodeName {
```
pedronis, line 198: this third level of switch is a bit much, I wonder if we can make an helper of this level of checks

efi/preinstall/check_pcr7.go
```diff
@@ -0,0 +1,529 @@
...
+				// This is an EV_EFI_BOOT_SERVICES_APPLICATION event during OS-present,
+				// and we haven't seen the event for the IBL yet. We stop once we see this
+				// because at this point, the rest of the measurements in this PCR are under
+				// the control of the OS.
+				data, ok := ev.Data.(*tcglog.EFIImageLoadEvent)
+				if !ok {
+					return nil, fmt.Errorf("invalid OS-present EV_EFI_BOOT_SERVICES_APPLICATION event data: %w", ev.Data.(error))
+				}
+
+				yes, err := isLaunchedFromLoadOption(ev, bootOpt)
```
pedronis, line 303: there is similar code in check_pcr4, maybe we can extract a helper to make this a bit more compact/readable?

pedronis (commented):
the nesting makes this a bit hard to follow/read, made some suggestions

efi/preinstall/check_pcr7.go
```diff
@@ -0,0 +1,537 @@
...
+	internal_efi "github.com/snapcore/secboot/internal/efi"
+	pe "github.com/snapcore/secboot/internal/pe1.14"
+)
+
+var (
+	internal_efiSecureBootSignaturesFromPEFile = internal_efi.SecureBootSignaturesFromPEFile
+	peNewFile                                  = pe.NewFile
+)
+
+// checSucureBootVariableData checks the variable data associated with a configuration
```
pedronis, line 44: typo checkSecure

efi/preinstall/check_pcr7.go
```diff
@@ -0,0 +1,537 @@
...
+					// sure it's consistent with the measured digest. We only do this for
+					// OS-present events because these are the ones we compute. Pre-OS
+					// events are just copied from the log.
+					seenOSPresentVerification = true
+					expectedDigest := tcglog.ComputeEFIVariableDataDigest(pcrAlg.GetHash(), data.UnicodeName, data.VariableName, data.VariableData)
+					if !bytes.Equal(ev.Digests[pcrAlg], expectedDigest) {
+						return nil, fmt.Errorf("event data inconsistent with %v event digest for EV_EFI_VARIABLE_AUTHORITY event (log digest:%#x, expected digest:%#x)", pcrAlg, ev.Digests[pcrAlg], expectedDigest)
+					}
+				}
+
```
pedronis, line 421: the check below is relevant both for tcglogPhasePreOSAfterMeasureSecureBootConfig and tcglogPhaseOSPresent: ?

efi/preinstall/check_pcr7.go
```diff
@@ -0,0 +1,537 @@
...
+						// we're in OS-present.
+						return nil, fmt.Errorf("unexpected EV_EFI_BOOT_SERVICES_APPLICATION event for %v after already seeing a verification event during the OS-present environment. "+
+							"This event should be for the initial boot loader", ev.Data.(*tcglog.EFIImageLoadEvent).DevicePath)
+					}
+					continue NextEvent
+				}
+
+				// This is the IBL for the OS. Obtain signatures from binary
+				seenIBLLoadEvent = true
+				sigs, err := func() ([]*efi.WinCertificateAuthenticode, error) {
```
pedronis, line 334: the code from here until the break should probably be an helper?

efi/preinstall/check_pcr7.go
```diff
@@ -0,0 +1,537 @@
...
+				// both the firmware and shim to both measure the same signature they used for verification from db
+				// because they both maintain their own de-duplication lists.
+				//
+				// If this test fails, the firmware is definitely broken. If this test doesn't fail, the opposite is
+				// not true - it's not a definitive guarantee that the firmware isn't broken, unfortunately.
+				for _, measured := range measuredSignatures {
+					if bytes.Equal(measured, ev.Digests[pcrAlg]) {
+						return nil, fmt.Errorf("EV_EFI_VARIABLE_AUTHORITY digest %#x has been measured by the firmware more than once", ev.Digests[pcrAlg])
+					}
+				}
```
pedronis, line 442: the rest of this stanza and maybe even the for or not should also be a helper

efi/preinstall/check_pcr7.go
```diff
@@ -0,0 +1,537 @@
...
+				//
+				// Just return an error here to prevent the use of WithSecureBootPolicyProfile(). The
+				// "UEFI Debug Mode" and "DMA Protection Disabled" cases are already picked up by the
+				// firmware protection checks, so we don't need any special handling here.
+				fallthrough
+			default:
+				// Anything that isn't EV_EFI_VARIABLE_DRIVER_CONFIG ends up here.
+				return nil, fmt.Errorf("unexpected %v event %q whilst measuring config", ev.EventType, ev.Data)
+			}
+		case tcglogPhasePreOSAfterMeasureSecureBootConfig, tcglogPhaseOSPresent:
```
pedronis, line 294: once we have more helpers maybe we can explore whether splitting in two cases with a little bit of repetition is easier to follow than the code below the rechecks the phase here and there

pedronis (commented):
sorry, a couple more readability suggestions

efi/preinstall/check_pcr7.go
chrisccoulson: It is.

chrisccoulson: I've split up the branches for tcglogPhasePreOSAfterMeasureSecureBootConfig and tcglogPhaseOSPresent entirely now.

efi/preinstall/check_pcr7.go
chrisccoulson: I've done this now. I hope it's a little bit cleaner.

efi/preinstall/check_pcr7.go
```diff
@@ -0,0 +1,712 @@
...
+				case err != nil:
+					return nil, fmt.Errorf("cannot determine if OS initial boot loader was verified by any X.509 certificate measured by any EV_EFI_VARIABLE_AUTHORITY event: %w", err)
+				}
+
+				ok, err := checkX509CertificatePublicKeyStrength(signer)
+				if err != nil {
+					return nil, fmt.Errorf("cannot determine public key strength of initial OS boot loader signer: %w", err)
+				}
+				if !ok {
+					result.Flags |= secureBootIncludesWeakAlg
```
pedronis, line 649: I suppose it's uncommon but this situation is not covered by tests

efi/preinstall/check_pcr7.go
```diff
@@ -0,0 +1,712 @@
...
+				ok, err := checkSignatureDataStrength(eslType, esdData)
+				if err != nil {
+					return nil, fmt.Errorf("cannot check strength of EFI_SIGNATURE_DATA associated with EV_EFI_VARIABLE_AUTHORITY event in OS-present phase: %w", err)
+				}
+				if !ok {
+					// XXX: unfortunately in the case where an image is signed, the verification event only includes the CA
+					// certificate - it's not possible from this to determine the actual signing certificate, it's signature
+					// algorithm, and the algorithm used for signing the binary. In this case, the check is on the CA's
+					// public key, which still has some value.
+					result.Flags |= secureBootIncludesWeakAlg
```
pedronis, line 685: same for this

pedronis (approved):
thanks, couple test comments

chrisccoulson:
This adds some checks for PCR7.

The caller supplies a context.Context to which an EFI variable backend
is attached, a internal_efi.HostEnvironment implementation, a TCG log, a
PCR digest algorith (the optimum for this is computed earlier by
another function that does a more general check of the TCG log) and the
image of the initial boot loader for the current boot.

As we're testing the firmware, this only checks the log up to the launch
of the initial boot loader. It ignores events after this, as these are
under the control of the OS and don't rely on any special firmware
features such as EFI_TCG2_PROTOCOL + PE_COFF_IMAGE vs EFI_TCG_PROTOCOL
in the same way that OS components measuring to PCR4 do.

This ensures that secure boot is enabled, else an error is returned, as
WithSecureBootPolicyProfile only generates profiles compatible with
secure boot being enabled.

If the version of UEFI is >= 2.5, it also makes sure that the secure
boot mode is "deployed mode". If the secure boot mode is "user mode",
then the "AuditMode" and "DeployedMode" values are measured to PCR7,
something that WithSecureBootPolicyProfile doesn't support today.
Support for "user mode" will be added in the future, although the public
RunChecks API will probably require a flag to opt in to supporting user
mode, as it is the less secure mode of the 2 (see the documentation for
SecureBootMode in github.com/canonical/go-efilib).

It also reads the "OsIndicationsSupported" variable to test for features
that are not supported by WithSecureBootPolicyProfile. These are
timestamp revocation (which requires an extra signature database -
"dbt") and OS recovery (which requires an extra signature database -
"dbr", used to control access to OsRecoveryOrder and OsRecover####
variables). Of the 2, it's likely that we might need to add support for
timestamp revocation at some point.

It reads the "BootCurrent" EFI variable and matches this to the
EFI_LOAD_OPTION associated with the current boot from the TCG log - it
uses the log as "BootXXXX" EFI variables can be updated at runtime and
might be out of data when this code runs. It uses this to detect the
launch of the initial boot loader, which might not necessarily be the
first EV_EFI_BOOT_SERVICES_APPLICATION event in the OS-present
environment in PCR4 (eg, if Absolute is active).

First of all, it iterates over the secure boot configuration in the log,
making sure that the configuration is measured in the correct order,
that the event data is valid, and that the measured digest is the tagged
hash of the event data. It makes sure that the value of "SecureBoot" in
the log is consistent with the "SecureBoot" variable (which is read-only
at runtime), and it verifies that all of the signature databases are
formatted correctly. It will return an error if any of these checks
fail.

If the pre-OS environment contains events other than
EV_EFI_VARIABLE_DRIVER_CONFIG, it will return an error. This can happen
if a firmware debugger is enabled, in which case PCR7 will begin with a
EV_EFI_ACTION "UEFI Debug Mode" event. This case is detected by earlier
firmware protection checks.

If not all of the expected secure boot configuration is measured, an
error is returned.

Once the secure boot configuration has been measured, it looks for
EV_EFI_VARIABLE_AUTHORITY events in PCR7, until it detects the launch of
the initial boot loader. It verifies that each of these come from db,
and if the log is in the OS-present environment, it ensures that the
measured digest is the tagged hash of the event data. It doesn't do this
for events in the pre-OS environment because WithSecureBootPolicyProfile
just copies these to the profile. It verifies that the firmware doesn't
measure a signature more than once. For each EV_EFI_VARIABLE_AUTHORITY
event, it also matches the measured signature to a EFI_SIGNATURE_LIST
structure in db. If the matched ESL is a X.509 certificate, it records
the use of this CA in the return value. If the CA is an RSA certificate
with a public modulus of < 256 bytes, it sets a flag in the return
value indicating a weak algorithm. If the matched ESL is a Authenticode
digest, it sets a flag in the return value indicating that pre-OS
components were verified using digests rather than signatures. This
makes PCR7 fragile wrt firmware updates, because it means db needs to
be updated to reflect the new components each time. If the digest being
matched is SHA-1, it sets the flag in the return value indicating a weak
algorithm. If any of these checks fail, an error is returned. If an
event type other than EV_EFI_VARIABLE_AUTHORITY is detected, an error is
returned.

Upon detecting the launch of the initial boot loader in PCR4, it
extracts the authenticode signatures from the supplied image, and
matches these to a previously measured CA. If no match is found, an
error is returned. If a match is found, it ensures that the signing
certificate has an RSA public key with a modulus at least as large as
256 bytes, else it sets a flag in the return value indicating a weak algorithm.
Once the event for the initial boot loader is complete, the function
returns.

If the end of the log is reached without encountering the launch of the
initial boot loader, an error is returned.

PR #338, opened 2024-10-04
preinstall: Add CheckResult and WithAutoTCGPCRProfile APIs

chrisccoulson:
`CheckResult` is returned from `RunChecks` on successful completion. It is
JSON serializable and intended to be supplied later on to
`WithAutoTCGPCRProfile` along with some user customization flags (defined
by `PCRProfileOptionsFlags`) in order to generate an option that can be
supplied to secboot_efi.AddPCRProfile.

Note that some user options don't work yet because of missing PCR
support, although these limitations will be addressed later. For now,
these will appear as failures in the `CheckResult` flags.


efi/preinstall/errors.go
```diff
@@ -20,14 +20,57 @@
...
 
+// indentLines is a helper for managing indenting in nested multi-line
+// errors.
+func indentLines(n int, str string) string {
+	r := bytes.NewReader([]byte(str))
+	w := new(bytes.Buffer)
+	br := bufio.NewReader(r)
+	for {
+		line, err := br.ReadString('\n')
+		fmt.Fprintf(w, "%*s%s", n, "", line)
```
pedronis, line 41: should we really do this print if line is "" and err is io.EOF ?

efi/preinstall/result.go
```diff
@@ -0,0 +1,550 @@
...
+	"no-drivers-and-apps-profile-support":        NoDriversAndAppsProfileSupport,
+	"no-drivers-and-apps-config-profile-support": NoDriversAndAppsConfigProfileSupport,
+	"no-boot-manager-code-profile-support":       NoBootManagerCodeProfileSupport,
+	"no-boot-manager-config-profile-support":     NoBootManagerConfigProfileSupport,
+	"no-secure-boot-policy-profile-support":      NoSecureBootPolicyProfileSupport,
+	"discrete-tpm-detected":                      DiscreteTPMDetected,
+	"startup-locality-not-protected":             StartupLocalityNotProtected,
+}
+
+type x509CertificateIdJSON struct {
```
pedronis, line 230: nitpick: I would split out the x509*Id* types to maybe their own x509id.go file

efi/preinstall/result.go
```diff
@@ -0,0 +1,550 @@
...
+			flags = append(flags, str)
+		}
+	}
+	fmt.Fprintf(w, "- Flags: %s\n", strings.Join(flags, ","))
+	if r.Warnings != nil && len(r.Warnings.Errs) > 0 {
+		fmt.Fprintf(w, "- Warnings:\n")
+		for i := 0; i < len(r.Warnings.Errs); i++ {
+			fmt.Fprintf(w, "%s\n", indentLines(2, "- "+r.Warnings.Errs[i].Error()))
+		}
+	}
```
pedronis, line 523: this bit of the code is not exercised

pedronis (commented):
did a first pass, probably a good idea for @valentindavid and maybe someone more from your side to look it, Mate?

efi/preinstall/errors.go
chrisccoulson: I've refactored this function a bit (it uses `bufio.Scanner` now and it's been renamed to make it clearer what its intended use is.

efi/preinstall/result.go
chrisccoulson: I've moved these types into their own file.

efi/preinstall/result.go
chrisccoulson: I've added a test for this now.

efi/preinstall/errors.go
```diff
@@ -20,14 +20,57 @@
...
+// to aborting early, in order for the caller to gather as much information as possible.
+type RunChecksErrors struct {
+	Errs []error // All of the errors collected during the execution of RunChecks.
+}
+
+func (e *RunChecksErrors) Error() string {
+	w := new(bytes.Buffer)
+	fmt.Fprintf(w, "one or more errors detected:\n")
+	for _, err := range e.Errs {
+		fmt.Fprintf(w, "%s\n", indentLines(2, "- "+err.Error()))
```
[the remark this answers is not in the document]
chrisccoulson: The expected formatting in the example would be:
```
- err1-line1
  err1-line2
- err2-line1
  err2-line2
```

In any case, I've refactored `indentLines` now and called it something else (`makeIndentedListItem`) to make its purpose clearer. I've also added some formatting tests for code that makes use of it too.

efi/preinstall/profile.go
```diff
@@ -0,0 +1,339 @@
...
+			// If PCR7 usage is ok, always include it
+			opts = append(opts, secboot_efi.WithSecureBootPolicyProfile())
+
+			if !knownCAs.trustedForBootManager(o.result.UsedSecureBootCAs) && o.opts&PCRProfileOptionTrustCAsForBootCode == 0 {
+				// We need to include PCR4 if any CAs used for verification are not generally trusted to sign boot applications
+				// (ie, they may have signed code in the past that can defeat our security model, such as versions of shim that
+				// don't extend anything to the TPM, breaking the root-of-trust. This is true of the Microsoft UEFI CA 2011,
+				// and for now, we assume to be true of the 2023 UEFI CA unless Microsoft are more transparent about what is
+				// signed under this CA). It's also assumed to be true for any unrecognized CAs.
+				// This can be overridden with PCRProfileOptionsTrustCAsForBootCode.
```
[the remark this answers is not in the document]
chrisccoulson: I've fixed this now.

efi/preinstall/profile.go
```diff
@@ -0,0 +1,339 @@
...
+				// Include PCR2 if the user explicitly distrusts non-host code running
+				// in attached embedded controllers.
+				return nil, fmt.Errorf("PCRProfileOptionDistrustVARSuppliedNonHostCode cannot be used: %w", newUnsupportedRequiredPCRsError(tpm2.HandleList{2}, o.result.Flags))
+			}
+			if !knownCAs.trustedForDrivers(o.result.UsedSecureBootCAs) && o.opts&PCRProfileOptionTrustCAsForVARSuppliedDrivers == 0 {
+				// We need to include PCR2 if any CAs used for verification are not generally trusted to sign UEFI drivers
+				// (ie, they may have signed code in the past that can defeat our security model. This is true of the Microsoft
+				// UEFI CA 2011, and for now, we assume to be true of the 2023 UEFI CA unless Microsoft are more transparent about
+				// what they sign under this CA). It's also assumed to be true for any unrecognized CAs.
+				// This can be overridden with PCRProfileOptionsTrustCAsForVARSuppliedDrivers.
```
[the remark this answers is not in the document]
chrisccoulson: And this one.

efi/preinstall/profile.go
```diff
@@ -0,0 +1,339 @@
...
+
+type authorityTrustData struct {
+	Authority *internal_efi.SecureBootAuthorityIdentity
+	Trust     authorityTrust
+}
+
+var (
+	knownCAs = authorityTrustDataSet{
+		{internal_efi.MSUefiCA2011, 0},
+		{internal_efi.MSUefiCA2023, 0}, // be conservative here for now, but will we be able to set the authorityTrustDrivers flag for the MS2023 CA?
```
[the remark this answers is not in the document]
chrisccoulson: You understood correctly - these are both explicitly distrusted, as is any certificate that doesn't appear in this list. Eventually we'll get to the point where we want security features that will require our own CA, which will be added here with both trust flags set (because we know we'll never use it to sign drivers). It also remains to be seen whether Microsoft will be any more open wrt to their signing requirements with the new UEFI CA, particularly wrt the signing of option ROMs. If this improves, we might be able to change one flag for this CA, but right now, I don't think it's been used to sign anything.

pedronis (approved):
thanks

valentindavid (approved):
Thank you.

PR #339, opened 2024-10-05
preinstall: Add public RunChecks API

chrisccoulson:
`RunChecks` is the first main entry point for running all of the platform
tests included in this package. It is considered to be quite a low
level and a future PR is going to add a higher level API that is easier
to use (`RunChecksContext`)

It can be customized with the `CheckFlags` argument, and the caller must
supply a list of images loaded during the current boot, in the correct
order.

On failure, it will return an error. In some cases, it fails early. In
other cases, it carries on and collects individual errors which are
then returned wrapped in `RunChecksError`. This can be obtained with the
`errors.As` API in order to obtain individual errors.

On success, it will return a `CheckResult`. Any errors that were
encountered but are considered non-fatal based on the supplied
`CheckFlags` will he added to the `Warnings` field of `CheckResult`.

The intention is that `CheckResult` is persisted and then used by an API
(to be added in the next PR) along with some user customization flags in
order to select the most appropriate set of TCG defined PCRs to seal
against. This can still fail if an appropriately secure set of PCRs
cannot be selected, even if `RunChecks` succeeded. The future
higher-level `RunChecksContext` API will make this case easier to handle
and will also help remove the burden of deciding which set of flags
to pass to `RunChecks`.


efi/preinstall/checks.go
```diff
@@ -0,0 +1,443 @@
...
+			case 4:
+				// There were H-CRTM events.  Mark PCR0 as reconstructible from anything that
+				// runs as part of the static OS for the reasons stated above if access to
+				// locality 4 isn't protected.
+				switch {
+				case protectedLocalities&tpm2.LocalityFour == 0 && flags&PermitNoDiscreteTPMResetMitigation > 0:
+					result.Flags |= StartupLocalityNotProtected
+				case protectedLocalities&tpm2.LocalityFour == 0:
+					mainErr.addErr(ErrTPMStartupLocalityNotProtected)
+				}
```
pedronis, line 311: this case is not tested

efi/preinstall/checks.go
```diff
@@ -0,0 +1,443 @@
...
+	if logResults.Lookup(internal_efi.DriversAndAppsPCR).Ok() {
+		// Only run PCR2 checks if we established earlier that the PCR value matches
+		// the reconstructed log value.
+		pcr2Results, err := checkDriversAndAppsMeasurements(log)
+		switch {
+		case err != nil && flags&DriversAndAppsProfileSupportRequired > 0:
+			mainErr.addErr(&DriversAndAppsPCRError{err})
+		case err != nil:
+			result.Flags |= NoDriversAndAppsProfileSupport
+			result.Warnings.addErr(&DriversAndAppsPCRError{err})
```
pedronis, line 337: same for these cases

pedronis (commented):
thanks, did a first pass

efi/preinstall/checks.go
chrisccoulson: We don't have a way to test these yet without adding some code for mocking whic might be a bit tricky, or until go-tpm2 has the ability to perform HCRTM sequences via the simulator API, which is something I'm working on separately (and is not trivial).

chrisccoulson: I should be able to do this via the simulator now, which has a public API for doing HCRTM sequences. The biggest challenge here is that the [simulator transport](https://pkg.go.dev/github.com/canonical/go-tpm2@v1.10.0/mssim#Transport.HashStart) implements a retry loop that talks to the actual `net.Conn` on a private goroutine, and introducing public APIs to send control commands along the same connection introduced some potential race conditions, but that's all fixed now.

efi/preinstall/checks.go
chrisccoulson: These should be easy to add test cases for.

chrisccoulson: Ah this one is actually relatively difficult to add a test for, because any injected failure will result in the earlier tests in `check_tcglog_test.go` failing first.

chrisccoulson: I've fixed this by removing the error from the return value and leaving a comment that errors are assumed to be caught by calling `checkFirmwareLogAndChoosePCRBank` prior to this.

pedronis (approved):
thank you, this is fairly large, is there a possible 2nd reviewer for this in Foundation?

PR #340, opened 2024-10-06
preinstall: move some extra TPM tests to only run during pre-install

chrisccoulson:
The `RunChecks` API will support being executed post-install even
though it's primarily a pre-install API. A use case for this is re-running
it as part of the post-recovery repair process. Not all tests makes sense
post-install, unless a motherboard (and TPM) change was the reason for
recovery, and so are skipped.

We were already skipping the lockout hierarchy auth value test during
post-install because this doesn't make sense, but also skip similar
tests for other hierarchies and also skip the lockout checkout and the
test for sufficient number of NV counters.


pedronis (approved):
thanks

PR #344, opened 2024-10-14
internal/luks2/cryptsetup.go: use --help output to check for --replace-token

valentindavid:
Because --replace-token is now required for installation, we need to backport this feature to 20.04. That is on cryptsetup 2.2.2.  However that version of cryptsetup does not support --test-args.  So instead we look at the output of --help.


valentindavid:
This is not a hight priority right now as we will deploy key datas in tokens only for core24+ for now.

valentindavid:
Closing for now. We will not use on 20 and 22 for now.

PR #345, opened 2024-10-15
preinstall: refinements to WithAutoPCRProfile

chrisccoulson:
(Draft until https://github.com/canonical/secboot/pull/340 has landed"

This makes some changes to `WithAutoPCRProfile`:

- It renames it to `WithAutoTCGPCRProfile` to make it clear that it only
  selects secure combinations of TCG defined PCRs.
- It introduces a new option -
  `PCRProfileOptionPermitNoSecureBootPolicyProfile`, so that
  configurations that don't use PCR7 aren't automatically selected.
  If PCR7 can't be used, this flag must be supplied to unlock a secure
  configuration that doesn't include PCR7.
- It introduces the `RequiredUnsupportedPCRsError` error type, returned
  from methods of `PCRProfileAutoEnablePCRsOption` to make error
  classification a bit easier later on.


chrisccoulson:
This is obsolete now.

PR #346, opened 2024-10-15
preinstall: Check the value of TPM_PT_NV_COUNTERS_AVAIL

chrisccoulson:
(Draft until https://github.com/canonical/secboot/pull/340 has landed).

When testing for sufficient NV counters, we check TPM_PT_NV_COUNTERS_MAX
(the maximum number of counters that are permitted to be allocated) and
TPM_PT_COUNTERS (the number of counters allocated).

There is another metric - TPM_PT_NV_COUNTERS_AVAIL which is an estimate
of the number of orderly counters that can be created based on the
amount of RAM remaining for orderly indices and the amount of shared NV
space remaining. This isn't perfect because we aren't creating orderly
counters (so the RAM constraints don't apply), and the TCG reference
library permits TPMs to always return 1 for this property, so we have to
handle this case as inconclusive.


chrisccoulson:
I'm going to close this one because I don't think it adds a huge amount of value.

PR #347, opened 2024-11-13
preinstall: Add explicit checks for empty PCR banks

chrisccoulson:
Whilst this has no consequence for FDE because we seal against a good
bank, it breaks measured boot as required by remote verifiers, as an
empty PCR bank provides a means for an adversary to spoof any host
platform of their choosing.

We reject systems with empty PCR banks by default, but with an opt-in to
permit it. Permitting it is fine for now, but when we get to the point of a
fully verified and attestable runtime in the future, this will not be permitted.
We will need to take some action here, such as capping PCRs 0-7 with an
`EV_SEPARATOR` type event that indicates an error occurred in those banks.

Whilst an empty PCR bank is considered a firmware bug and we shouldn't be
seeing it on newer devices, it's not that uncommon to see this on devices from
the era when TPM2 devices started introducing SHA384 before the firmware
was ready to use it. We may want to see if we can design a scheme where we
can provide proof that a PCR bank was populated by the firmware and not by
an adversary.

### DIFF
--- a/efi/preinstall/check_tcglog_test.go
+++ b/efi/preinstall/check_tcglog_test.go
@@ -222,6 +222,7 @@ type testCheckFirmwareLogAndChoosePCRBankParams struct {
 	startupLocality           uint8
 	disallowPreOSVerification bool
 	mandatoryPcrs             tpm2.HandleList
+	permitEmptyPCRBanks       bool
 
 	expectedAlg tpm2.HashAlgorithmId
 }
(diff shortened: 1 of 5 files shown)


efi/preinstall/check_tcglog_test.go
```diff
@@ -978,6 +984,95 @@ func (s *tcglogSuite) TestCheckFirmwareLogAndChoosePCRBankTruncatedLog(c *C) {
...
+		tpm2.HandleList{
+			internal_efi.PlatformConfigPCR,
+			internal_efi.DriversAndAppsPCR,
+			internal_efi.DriversAndAppsConfigPCR,
+			internal_efi.BootManagerCodePCR,
+			internal_efi.BootManagerConfigPCR,
+			internal_efi.PlatformManufacturerPCR,
+			internal_efi.SecureBootPolicyPCR,
+		},
+		true,
```
pedronis, line 1047: at least for the true is would be good to use a const:

const permitEmptyPCRBanks = true

before the call and use it here

pedronis (approved):
thx, one nitpick

PR #352, opened 2024-12-02
tpm2: Create TPM2 keys with noDA attribute if no user auth is required

chrisccoulson:
By default, all TPM2 keys are protected by the TPM's dictionary attack
protection logic. Given that the user auth mode is immutable, we should
take advantage of this to create keys that require no user auth with the
noDA attribute, so that they aren't protected by the TPM's dictionary
attack protection logic. This means that these keys will still be
recoverable, even if the TPM is in lockout mode.

### DIFF
--- a/tpm2/seal.go
+++ b/tpm2/seal.go
@@ -159,7 +159,9 @@ var makeSealedKeyData = func(tpm *tpm2.TPMContext, params *makeSealedKeyDataPara
 		return nil, nil, nil, xerrors.Errorf("cannot create symmetric key: %w", err)
 	}
 
-	// Seal the symmetric key and nonce.
+	// Seal the symmetric key and nonce. The final boolean argument is set to true in order
+	// to disable dictionary attack protection (ie, adding the noDA attribute). We want this
+	// when no user auth value is required.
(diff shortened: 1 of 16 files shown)


chrisccoulson:
I can't request a review from you @tonyespy, but I thought you might be interested in this one :)

tpm2/seal.go
```diff
@@ -160,7 +160,7 @@ func makeSealedKeyData(tpm *tpm2.TPMContext, params *makeSealedKeyDataParams, se
 	}
 
 	// Seal the symmetric key and nonce.
-	priv, pub, importSymSeed, err := sealer.CreateSealedObject(symKey[:], nameAlg, authPolicyDigest)
+	priv, pub, importSymSeed, err := sealer.CreateSealedObject(symKey[:], nameAlg, authPolicyDigest, !requireAuthValue)
```
pedronis, line 163: probably worth having a comment here explaining !RequireAuthValue policy/logic

pedronis (commented):
thanks, small comment

tpm2/seal.go
chrisccoulson: I've added a comment here now.

pedronis (approved):
thanks

PR #355, opened 2024-12-10
crypt.go: allow renaming keys without cryptsetup features

valentindavid:
Factory reset requires renaming of keyslots. Unfortunately `cryptsetup` on UC 20 and UC 22 do not support `--token-replace`. So we need a way rename without that feature.

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -890,9 +890,7 @@ func DeleteLUKS2ContainerKey(devicePath, keyslotName string) error {
 	return nil
 }
 
-// RenameLUKS2Container key renames the keyslot with the specified oldName on
-// the LUKS2 container at the specified path.
-func RenameLUKS2ContainerKey(devicePath, oldName, newName string) error {
+func renameLUKS2ContainerKey(devicePath, oldName, newName string, replace bool) error {
 	view, err := newLUKSView(devicePath, luks2.LockModeBlocking)
 	if err != nil {
 		return xerrors.Errorf("cannot obtain LUKS header view: %w", err)
@@ -928,11 +926,35 @@ func RenameLUKS2ContainerKey(devicePath, oldName, newName string) error {
 		return errors.New("cannot rename key with unexpected token type")
 	}
 
-	if err := luks2ImportToken(devicePath, newToken, &luks2.ImportTokenOptions{Id: id, Replace: true}); err != nil {
-		return xerrors.Errorf("cannot import new token: %w", err)
+	if replace {
+		if err := luks2ImportToken(devicePath, newToken, &luks2.ImportTokenOptions{Id: id, Replace: true}); err != nil {
+			return xerrors.Errorf("cannot import new token: %w", err)
+		}
+	} else {
+		if err := luks2ImportToken(devicePath, newToken, nil); err != nil {
+			return xerrors.Errorf("cannot import new token: %w", err)
+		}
+		if err := luks2RemoveToken(devicePath, id); err != nil {
+			return xerrors.Errorf("cannot remove existing token %d: %w", id, err)
+		}
 	}
 
 	return nil
+}
+
+// RenameLUKS2Container key renames the keyslot with the specified oldName on
+// the LUKS2 container at the specified path.
+func RenameLUKS2ContainerKey(devicePath, oldName, newName string) error {
+	const replace = true
+	return renameLUKS2ContainerKey(devicePath, oldName, newName, replace)
+}
+
+// CopyAndRemoveLUKS2ContainerKey key renames the keyslot with the
+// specified oldName on the LUKS2 container at the specified path. It
+// does not use new feature from cryptsetup. However it is not atomic.
+func CopyAndRemoveLUKS2ContainerKey(devicePath, oldName, newName string) error {
+	const replace = false
+	return renameLUKS2ContainerKey(devicePath, oldName, newName, replace)
 }
 
 // KeyslotAlreadyHasANameErr may be returned by
--- a/crypt.go
+++ b/crypt.go
@@ -890,7 +890,18 @@ func DeleteLUKS2ContainerKey(devicePath, keyslotName string) error {
 	return nil
 }
 
-func renameLUKS2ContainerKey(devicePath, oldName, newName string, replace bool) error {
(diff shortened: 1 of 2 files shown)


valentindavid:
This is used in https://github.com/canonical/snapd/pull/14471

crypt.go
```diff
@@ -928,9 +926,33 @@ func RenameLUKS2ContainerKey(devicePath, oldName, newName string) error {
...
-	if err := luks2ImportToken(devicePath, newToken, &luks2.ImportTokenOptions{Id: id, Replace: true}); err != nil {
-		return xerrors.Errorf("cannot import new token: %w", err)
+	if replace {
+		if err := luks2ImportToken(devicePath, newToken, &luks2.ImportTokenOptions{Id: id, Replace: true}); err != nil {
+			return xerrors.Errorf("cannot import new token: %w", err)
+		}
+	} else {
+		if err := luks2ImportToken(devicePath, newToken, nil); err != nil {
+			return xerrors.Errorf("cannot import new token: %w", err)
+		}
```
pedronis, line 936: @valentindavid what happens to the factory reset process if we reboot here?

pedronis (commented):
question

chrisccoulson:
This is a bit problematic. The reason that I added this feature to cryptsetup is that without an atomic replace, we need to add counters to tokens and provide a way of cleaning up in the event that the transaction is interrupted after the new token is added but before the old one is deleted. As it stands, if this PR were to land in its current form and the process was interrupted after adding the new token, then parsing the header fails on subsequent boots because we reject headers that have multiple tokens with the same name (in `internal/luksview`).

valentindavid:
> then parsing the header fails on subsequent boots because we reject headers that have multiple tokens with the same name

They have different names. They just point to the same keyslot. Would that be an issue?

crypt.go
valentindavid: Probably we should add some tests here. I think what we have here is multiple tokens pointing to the same key slot. So both newName and oldName are pointing to the keyslot.

valentindavid: In `internal/luksview/view_test.go`, `TestNewView` actually uses 2 tokens with the same keyslot. Tokens 0 and 1 point to keyslot 0. So I am guessing this is fine.

chrisccoulson: If we end up with a superfluous token because the process gets interrupted, and reprovision is implemented as described in the spec, then a future reprovision with the latest cryptsetup will fix this.

chrisccoulson (approved):
I'll let @pedronis comment on the actual API

crypt.go
```diff
@@ -928,9 +935,36 @@ func RenameLUKS2ContainerKey(devicePath, oldName, newName string) error {
...
+// RenameLUKS2Container key renames the keyslot with the specified oldName on
+// the LUKS2 container at the specified path.
+func RenameLUKS2ContainerKey(devicePath, oldName, newName string) error {
+	const replace = true
+	return renameLUKS2ContainerKey(nil, devicePath, oldName, newName, replace)
+}
+
+// CopyAndRemoveLUKS2ContainerKey key renames the keyslot with the
+// specified oldName on the LUKS2 container at the specified path. It
+// does not use new feature from cryptsetup. However it is not atomic.
```
pedronis, line 966: is the comment true? it seems it will use the atomic replace if available and it will fallback to not using it only if one passes the flag?

crypt.go
```diff
@@ -890,9 +890,16 @@ func DeleteLUKS2ContainerKey(devicePath, keyslotName string) error {
...
 
-// RenameLUKS2Container key renames the keyslot with the specified oldName on
-// the LUKS2 container at the specified path.
-func RenameLUKS2ContainerKey(devicePath, oldName, newName string) error {
+type nonProliferationFlag struct {
+}
+
+// INeedToUseAFeatureThatShouldNotProliferate gives a flag that can authorize
+// some functions that could be dangerous in some context to be used.
+func INeedToUseAFeatureThatShouldNotProliferate() *nonProliferationFlag {
```
pedronis, line 898: we probably need to think about the name of this and the type

pedronis (commented):
couple of questions

crypt.go
valentindavid: `RenameLUKS2ContainerKey` is atomic. `CopyAndRemoveLUKS2ContainerKey` is not. `RenameLUKS2ContainerKey` is still l the same as it was. `CopyAndRemoveLUKS2ContainerKey` was added, but need to construct that special object to be able to call it.

The fallback will happen in snapd. We will try the atomic call always. And in the factory reset, if secboot tell us that the feature needed is not available, then we will call `CopyAndRemoveLUKS2ContainerKey` instead.

crypt.go
```diff
@@ -890,9 +890,16 @@ func DeleteLUKS2ContainerKey(devicePath, keyslotName string) error {
 	return nil
 }
 
-// RenameLUKS2Container key renames the keyslot with the specified oldName on
-// the LUKS2 container at the specified path.
-func RenameLUKS2ContainerKey(devicePath, oldName, newName string) error {
+type nonProliferationFlag struct {
```
pedronis, line 893: let's rename this to: nonAtomicOperationAllowedFlag

crypt.go
pedronis: then let's rename this to: AllowNonAtomicOperation

crypt.go
pedronis, line 966: To add here:

This must be used with care and intentionally. AllowNontAtomicOperation result must be provided to allow for its use.

crypt.go
```diff
@@ -890,9 +890,16 @@ func DeleteLUKS2ContainerKey(devicePath, keyslotName string) error {
...
+}
+
+// INeedToUseAFeatureThatShouldNotProliferate gives a flag that can authorize
+// some functions that could be dangerous in some context to be used.
```
pedronis, line 897: // AllowNonAtomicOperation gives a flag to allow some operations that are not atomic.
// Those can be dangerous depending on usage context, they should be used with care and intentionally
// (that's why the extra hoops).

pedronis (commented):
some renaming and doc comments suggestions. Thank you

pedronis (approved):
thank you

Issue #350, opened 2024-11-23
Introduce github.com/canonical/go-tpm2/poliyutil


chrisccoulson:
We currently use [github.com/canonical/go-tpm2/util.TrialAuthPolicy](https://pkg.go.dev/github.com/canonical/go-tpm2@v1.7.7/util#TrialAuthPolicy) for computing policy digests, and then manually call the various `TPM2_Policy*` commands via [tpm2.TPMContext](https://pkg.go.dev/github.com/canonical/go-tpm2@v1.7.7#TPMContext) during policy execution. Handling of branches and the storage of metadata for branches is done in an ad-hoc fashion in various places, and this approach requires keeping policy computation in sync with policy execution. It also means that changes to policies require changing the key data format in some way, which is a pain. And it makes it hard to have slightly different policies depending on the key type (eg, keys without PIN or passphrase lack the `TPM2_PolicyAuthValue` assertion, but we add a boolean to the key data to indicate whether this assertion should be executed). In the future, we might want keys that are bound to the TPM reset count and time limited in order to avoid having the re-entering a PIN or passphrase after a reboot, after applying system updates (something I talked to Tim about before he left), but there's no realistic way of us doing this with our current approach to handling policies.

[https://pkg.go.dev/github.com/canonical/go-tpm2/policyutil#PolicyBuilder](https://pkg.go.dev/github.com/canonical/go-tpm2/policyutil#PolicyBuilder) is an API for building arbitrarily complicated policies and completely replaces the existing API. It can be used in the same way that the existing API is used (just to compute a policy digest), or you can make use of the [https://pkg.go.dev/github.com/canonical/go-tpm2/policyutil#Policy](https://pkg.go.dev/github.com/canonical/go-tpm2/policyutil#Policy) structure it builds in order to execute the policy. The policy can be serialized to persistent storage and read back later on. Where [https://pkg.go.dev/github.com/canonical/go-tpm2/policyutil#Policy.Execute](https://pkg.go.dev/github.com/canonical/go-tpm2/policyutil#Policy.Execute) encounters branches, it attempts to select the best branch (the one which is most likely to succeed), and there are things you can do when constructing policies to make this more likely to succeed. It does also provide a way to manually specify a path to execute in the policy in the case that a policy is constructed in such a way that a branch cannot automatically be selected.

Introducing this would mean we could make arbitrary changes to policies in order to introduce new features without having to introduce a new metadata version.

PR #358, opened 2025-01-16
tpm2: Stop using deprecated go-tpm2 APIs

chrisccoulson:
PR #357 migrated the tpm2 code to using the new tpm2.TPMDevice abstraction for opening TPM connections.

The go-tpm2 package contains some other deprecated APIs, and in some cases, entire sub-packages have been deprecated (crypto, templates, util). These have been replaced by alternative APIs, and the util package, which was a bit of a dumping ground for APIs that had nowhere else to go, has been split into more focused packages.

This ports secboot to using updated APIs. It's just a straight port for now - we may want to refactor some code to make better use of these APIs in future PRs - particularly those in the `policyutil` sub-package, something that will allow us to create keys with arbitrary policies without having to change the key data format for tpm2 keys (see issue #350).


pedronis (approved):
thank you

PR #359, opened 2025-01-16
internal/efi: Use internal/tpm2_device package

chrisccoulson:
Other code in secboot makes use of this recently introduced package,
so the code in internal/efi should use it too.


pedronis (approved):
thanks

Issue #353, opened 2024-12-05
Conflicting double TPM connections when creating a new passphrase protected key


ZeyadYasser:
`NewTPMPassphraseProtectedKey` takes a tpm connection as input, but internally the constructor ends up calling [`ChangeAuthKey` ](https://github.com/canonical/secboot/blob/1224447831c1a5f3194ea8e934b8624f5bb69791/tpm2/platform.go#L123-L124)on the tpm platform handler which tries to create a new TPM connection causing an error due trying to open two connection for the same TPM device.

My initial thoughts was to override `ConnectToTPM` to reuse the opened tpm connection and restore it when `NewTPMPassphraseProtectedKey` completes and it works https://github.com/ZeyadYasser/secboot/commit/56aea41e790038e235b322d8665e9bd3121f69e3 as expected, but unfortunately this approach is not concurrent-safe and given that there could be multiple ongoing snapd changes at the same time, it doesn't look like a viable solution.

chrisccoulson:
I'll be working on this now #357 has landed, which was a pre-requisite for this.

chrisccoulson:
This is fixed since #360 landed.

PR #360, opened 2025-01-16
tpm2: Avoid attempting to open duplicate TPM connections

chrisccoulson:
When using `NewTPMPassphraseProtectedKey`, it is passed an already open
connection. But the core secboot package calls back into the platform
handler to set the user auth value, which opens a second connection.
When using the direct device (/dev/tpm0), this doesn't work.

This updates the test suite to make sure that this scenario is caught
and results in an error. It also fixes the problem by providing a way to
pass the open TPM connection to the singleton tpm2 platform handler
via a new field on the `secboot.KeyWithPassphraseParams` struct which
can be set to any value and is passed to the `ChangeAuthKey` implementation
when setting the initial auth value. The tpm2 platform handler
implementation uses this connection if passed rather than trying to open a
new connection. If the `ChangeAuthKey` implementation is called via
`KeyData.ChangePassphrase`, then the new argument will be nil and the
tpm2 implementation will open a new TPM connection as required.

This also identified a bug that the core secboot package was not passing
the role value to the platform handler, which only seemed to be caught
by adding tests that created passphrase protected TPM keys with a role.

This addresses issue #353

### DIFF
--- a/tpm2/seal.go
+++ b/tpm2/seal.go
@@ -79,8 +79,13 @@ func makeKeyDataNoAuth(skd *SealedKeyData, role string, encryptedPayload []byte,
 	})
 }
 
-func makeKeyDataWithPassphraseConstructor(kdfOptions secboot.KDFOptions, passphrase string) keyDataConstructor {
+func makeKeyDataWithPassphraseConstructor(tpm *Connection, kdfOptions secboot.KDFOptions, passphrase string) keyDataConstructor {
 	return func(skd *SealedKeyData, role string, encryptedPayload []byte, kdfAlg crypto.Hash) (*secboot.KeyData, error) {
+		// Avoid trying to reopen another connection when setting the user auth value
+		orig := mainPlatformKeyDataHandler.tpm
+		mainPlatformKeyDataHandler.tpm = tpm
+		defer func() { mainPlatformKeyDataHandler.tpm = orig }()
+
 		return secbootNewKeyDataWithPassphrase(&secboot.KeyWithPassphraseParams{
 			KeyParams: secboot.KeyParams{
 				Handle:           skd,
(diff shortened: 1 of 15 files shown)


tpm2/seal.go
```diff
@@ -79,8 +79,13 @@ func makeKeyDataNoAuth(skd *SealedKeyData, role string, encryptedPayload []byte,
 	})
 }
 
-func makeKeyDataWithPassphraseConstructor(kdfOptions secboot.KDFOptions, passphrase string) keyDataConstructor {
+func makeKeyDataWithPassphraseConstructor(tpm *Connection, kdfOptions secboot.KDFOptions, passphrase string) keyDataConstructor {
 	return func(skd *SealedKeyData, role string, encryptedPayload []byte, kdfAlg crypto.Hash) (*secboot.KeyData, error) {
+		// Avoid trying to reopen another connection when setting the user auth value
+		orig := mainPlatformKeyDataHandler.tpm
+		mainPlatformKeyDataHandler.tpm = tpm
+		defer func() { mainPlatformKeyDataHandler.tpm = orig }()
```
pedronis, line 87: depending how things are used this could create a data race, do we need a global mutex around some of the handler usage?

pedronis (commented):
question

tpm2/seal.go
chrisccoulson: Yes, that might be a good idea - I'll add that. We might want to actually hold the lock throughout the whole code section that runs with the connection set, so that any other code that wants to set the connection will have to wait, else things might get a bit confusing if there are multiple connections around.

chrisccoulson: Whilst trying to figure out the best way to do this, and given that I'd like to reduce existing data races, I ended up taking a different approach by permitting the open connection to be passed to `NewKeyDataWithPassphrase` and subsequently back to the `tpm2` platform handler via a new `context` argument to `ChangeAuthKey`. It seemed like the safest way to do it in the end.

pedronis (commented):
thanks, this looks alright,  shouldn't now NewTPMProtectedKey doc comment say something abut when the TPM connection passed in can be closed?

pedronis (approved):
sorry my question didn't make sense, I hastily misread the changes to keydata.go. +1

PR #361, opened 2025-01-16
tpm2: Add more tests with the role field populated

chrisccoulson:
The vast majority of tests were performed with the role field empty, which isn't going to be the typical situation.


pedronis (approved):
thank you

PR #364, opened 2025-01-17
internal/testutil: Update import for KBKDF

chrisccoulson:
The previous import was github.com/canonical/go-sp800.108-kdf. I've since moved it to github.com/canonical/go-kbkdf and renamed the module accordingly, following the naming convention for this KDF on openssl.

The package is largely the same, just with the HMAC implementation of the PRF split into its own sub-package. This package has global variables for pre-defined PRFs for each supported SHA2 algorithm.


pedronis (approved):
+1

PR #365, opened 2025-01-21
preinstall: Rename EmptyPCRBankError to EmptyPCRBanksError

chrisccoulson:
The new error captures empty PCRs for all active banks in which they exist as opposed to terminating early.

Note that I needed a 3rd algorithm to properly test the multiple empty PCR bank case whilst still having one good PCR bank. Given that the simulator doesn't have SHA-512 enabled, this needed to be SHA1. This means that this PR also adds support for selecting SHA1 as a PCR digest, but it is hidden behind a flag (PermitWeakPCRBanks).


pedronis (approved):
thanks

Issue #330, opened 2024-09-09
Refactor testing for tpm2 package


chrisccoulson:
I want to refactor the testing to be more in line with one of the recently added packages, eg, plainkey:

1. For New* functions, test the returned private values against statically compiled values.
2. For serialization tests, test the serialization against a statically contained value, such as a JSON string or sequence of bytes
3. For deserialization tests, test the deserialization against statically contained values to detect for backwards compatibility issues
4. Add unit tests for other features as usual.

PR #366, opened 2025-01-22
Fix plumbing of the role via the PlatformKeyDataHandler interface

chrisccoulson:
PR #361 addressed the issue that most tpm2 tests were being performed with an empty role, which isn't realistic. It turns out that I missed a few places, particularly in some of the platform tests, that if I had fixed in the previous PR, would have noticed that the role parameter isn't being passed through the platform interface (this would get picked up when changing a password because key validation would fail).

This addresses this, and fixes tests in the top-level package to catch it as well.

Note that I'm using the same value in most tests - my normal approach to unit testing is to test each input with at least more than one value, but the tests in the top-level secboot package and the tpm2 sub-package are getting a bit complicated with a lot of inconsistencies in their approach to testing. I have issues #329 and #330 open to refactor the tests so they are structured in a more consistent way, like the tests for the recently added plainkey platform are.

### DIFF
--- a/tpm2/platform_test.go
+++ b/tpm2/platform_test.go
@@ -76,7 +76,7 @@ func (s *platformSuite) TestRecoverKeysIntegrated(c *C) {
 	params := &ProtectKeyParams{
 		PCRProfile:             tpm2test.NewPCRProfileFromCurrentValues(tpm2.HashAlgorithmSHA256, []int{7}),
 		PCRPolicyCounterHandle: s.NextAvailableHandle(c, 0x0181fff0),
-		Role:                   "",
+		Role:                   "foo",
 	}
 
 	k, primaryKey, unlockKey, err := NewTPMProtectedKey(s.TPM(), params)
@@ -92,7 +92,7 @@ func (s *platformSuite) TestRecoverKeysWithPassphraseIntegrated(c *C) {
 	params := &ProtectKeyParams{
 		PCRProfile:             tpm2test.NewPCRProfileFromCurrentValues(tpm2.HashAlgorithmSHA256, []int{7}),
 		PCRPolicyCounterHandle: s.NextAvailableHandle(c, 0x0181fff0),
-		Role:                   "",
+		Role:                   "bar",
 	}
 
 	passphraseParams := &PassphraseProtectKeyParams{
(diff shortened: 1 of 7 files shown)


chrisccoulson:
Requesting a review from Spyros as well because this is functionality that he worked on recently.

pedronis (commented):
thanks, there is still a few tests in platform_test.go and seal_test.go that have Role: "", I imagine those are intentional now?

tpm2/platform_test.go
```diff
@@ -239,19 +240,31 @@ func (s *platformSuite) TestRecoverKeysSimplePCRProfile(c *C) {
...
+		Role:                   "foo",
+	})
+}
+
+func (s *platformSuite) TestRecoverKeysDifferentRole(c *C) {
+	s.testRecoverKeys(c, &ProtectKeyParams{
+		PCRProfile:             tpm2test.NewPCRProfileFromCurrentValues(tpm2.HashAlgorithmSHA256, []int{7}),
+		PCRPolicyCounterHandle: s.NextAvailableHandle(c, 0x0181fff0),
+		Role:                   "bar",
+	})
```
sespiros, line 267: Not sure if that would be redundant, but an extra test case that tests that a calculated pcrpolicyref with role "foo" is different than a generated pcrpolicyref with role "bar", might also make sense. That would effectively test the private function `computeV3PcrPolicyRef` and that the role field is indeed affecting the calculated policy digest I think.

chrisccoulson:
> thanks, there is still a few tests in platform_test.go and seal_test.go that have Role: "", I imagine those are intentional now?

Yes, these should generally just be limited to tests that are testing error cases anyway.

tpm2/platform_test.go
chrisccoulson: We're doing that indirectly from the tests of `newKeyDataPolicy` since #361, which has an explicit test of the derived policyRef against a hardcoded value.

chrisccoulson: Actually, we may want to directly include the role value inside the AAD on the tpm2 side as well, else its only checked when methods that result in `keyData.ValidateData` being called or executed.

pedronis: > Actually, we may want to directly include the role value inside the AAD on the tpm2 side as well

that sounds like a key format change again, is that right?

PR #367, opened 2025-01-24
efi/preinstall: Use BootGuard Status (0x13a) MSR for TPM discreteness detection

kukrimate:
More detailed explanation in the commit message.


chrisccoulson:
@andersson1234 Do we have a selection of devices with Intel CPUs - some with discrete TPMs and some with Intel PTT (Intel's firmware TPM implementation) - that we can test these changes against to make sure that they will work reliably in production? Unfortunately, this stuff is very poorly documented in public.

andersson1234:
Yes. We have an hp pro sff 400 g9, which seems to have a discrete TPM, and we also have a dell precision 7865, which seems to have PTT. We also have a Lenovo Thinkstation P350, which has a discrete TPM. I've attached a spec document for each machine to this comment.

[dell-precision-7865.pdf](https://github.com/user-attachments/files/18600651/dell-precision-7865.pdf)
[hp-pro-sff-400-g9.pdf](https://github.com/user-attachments/files/18600652/hp-pro-sff-400-g9.pdf)

[ThinkStation_P350_Tower_Spec.pdf](https://github.com/user-attachments/files/18600664/ThinkStation_P350_Tower_Spec.pdf)

These machines are, however, currently "down" - we can bring them back up after next week :)

SimbiotVenom:
Yes, we have the HP EliteBook 850 G8 and HP EliteBook 860 G11, both with Intel CPU and discrete TPM.
Tested Ubuntu 24.04.1, 24.10, and 25.04, daily builds, still after TPM FDE installation, it asks for the recovery key for disk.

efi/preinstall/check_fw_protections_amd64.go
```diff
@@ -114,21 +114,25 @@ func checkPlatformFirmwareProtections(env internal_efi.HostEnvironment, log *tcg
 			// for implementing a D-CRTM with Intel TXT (which I think is SMX).
 			protectedStartupLocalities |= tpm2.LocalityThree | tpm2.LocalityFour
 		}
+		discreteTPM, err = checkIsTpmDiscreteIntel(amd64Env)
```
chrisccoulson, line 117: Rather than having this check as part of the platform firmware protection checks, I wonder whether we should move it to the existing `openAndCheckTPM2Device` function. It already has access to the `internal_efi.HostEnvironment` - the downside is that we have to call `determineCPUVendor` twice, but I think that's ok.

efi/preinstall/check_fw_protections_intel.go
```diff
@@ -232,7 +232,7 @@ func checkPlatformFirmwareProtectionsIntelMEI(env internal_efi.HostEnvironment)
 	// From here, these checks are based on the HSI checks performed in the pci-mei
 	// plugin in fwupd.
 	family := calculateIntelMEFamily(vers, hfsts1Reg)
-	if err != nil {
+	if family == meFamilyUnknown {
```
chrisccoulson, line 235: Good catch there

efi/preinstall/check_fw_protections_intel.go
```diff
@@ -298,3 +298,36 @@ func checkPlatformFirmwareProtectionsIntelMEI(env internal_efi.HostEnvironment)
 	// Everything is ok
 	return nil
 }
+
+const bootGuardStatusMsr = 0x13a
+
+const (
+	bootGuardStatusTpmMask uint64 = (3 << 1)
+	bootGuardStatusTpmNone uint64 = (0 << 1)
+	bootGuardStatusTpm12   uint64 = (1 << 1)
```
chrisccoulson, line 307: Given my earlier suggestion about moving the check to `openAndCheckTPM2Device` I wonder what the behaviour should be if we get conflicting values? (ie, we open a TPM2 device but the MSR indicates none or TPM1.2). Should this be an error or should we just assume a discrete device to be on the safe side?

efi/preinstall/check_fw_protections_intel.go
```diff
@@ -298,3 +298,36 @@ func checkPlatformFirmwareProtectionsIntelMEI(env internal_efi.HostEnvironment)
...
+
+func checkIsTpmDiscreteIntel(env internal_efi.HostEnvironmentAMD64) (discreteTPM bool, err error) {
+	bootGuardStatus, err := env.ReadMSRs(bootGuardStatusMsr)
+	if err != nil {
+		return false, fmt.Errorf("failed to read BootGuard status")
+	}
+	// NOTE: bootGuardStatus[0] is fine because BootGuard status MSR has the same value across all CPUs
+	switch bootGuardStatus[0] & bootGuardStatusTpmMask {
+	case bootGuardStatusTpmNone:
+		return false, fmt.Errorf("BootGuard status says the system has no TPM")
```
chrisccoulson, line 320: In both of these cases, we have `ErrNoTPM2Device` so that the caller can detect what the error is - the error can be wrapped with more context using `fmt.Errorf` and the `%w` verb.

chrisccoulson (changes requested):
Thanks for this - I've left a few comments inline. I think we should probably still keep this check within `openAndCheckTPM2Device` as it feels a bit strange to return this new value from the platform firmware protection tests.

efi/preinstall/check_fw_protections_intel.go
kukrimate: I think it is reasonable to just assume the TPM is discrete in that case. All this BootGuard stuff is badly/not documented, and I wouldn't throw hard errors based on that.

chrisccoulson (changes requested):
Thanks for updating this. I've given it another review - there's a fair few comments, although none of them are anything major. I'm just a little bit concerned about the blanket suppression of errors that happens with the `isTpmDiscrete` implementation.

I haven't ran it yet to check for test coverage, and I also haven't reviewed the test changes in detail, so I might have some follow-up comments. Did you check that every existing test inside `checks_test.go` and `check_tpm_test.go` get an implementation of `efitest.WithAMD64Environment` with the MSR defined where needed? (for tests that are testing for success and for tests that are testing for error cases, but where they still hit this new code).

Note, if we end up changing `isTpmDiscrete` so that it doesn't blanket suppress all errors, there might be some extra error case tests that you need to add afterwards to `check_tpm_test.go`

chrisccoulson (approved):
Thanks for the updates - I've made a passing glance over the changes, and I think things look ok, but do you want to take a quick look @pedronis ?

efi/preinstall/check_fw_protections_intel.go
```diff
@@ -298,3 +298,35 @@ func checkPlatformFirmwareProtectionsIntelMEI(env internal_efi.HostEnvironment)
...
 }
+
+const bootGuardStatusMsr = 0x13a
+
+const (
+	bootGuardStatusTpmMask uint64 = (3 << 1)
+	bootGuardStatusTpmNone uint64 = (0 << 1)
+	bootGuardStatusTpm12   uint64 = (1 << 1)
+	bootGuardStatusTpm2    uint64 = (2 << 1)
+	bootGuardStatusTpmPtt  uint64 = (3 << 1)
```
pedronis, line 309: do I read this right, this is == mask?

efi/preinstall/check_fw_protections_intel.go
```diff
@@ -298,3 +298,35 @@ func checkPlatformFirmwareProtectionsIntelMEI(env internal_efi.HostEnvironment)
...
+func checkIsTpmDiscreteIntel(env internal_efi.HostEnvironmentAMD64) (discreteTPM bool, err error) {
+	bootGuardStatus, err := env.ReadMSRs(bootGuardStatusMsr)
+	if err != nil {
+		return false, fmt.Errorf("failed to read BootGuard status: %w", err)
+	}
+	// NOTE: bootGuardStatus[0] is fine because BootGuard status MSR has the same value across all CPUs
+	switch bootGuardStatus[0] & bootGuardStatusTpmMask {
+	// System has no TPM or unsupported TPM 1.2 device
+	case bootGuardStatusTpmNone, bootGuardStatusTpm12:
+		return false, ErrNoTPM2Device
```
pedronis, line 321: this case doesn't seem reached by tests actually

pedronis (commented):
@kukrimate couple of comments on this

efi/preinstall/check_fw_protections_intel.go
kukrimate: Yeah max value of the field is the same as the mask

efi/preinstall/check_fw_protections_intel.go
kukrimate: Added new tests

pedronis (approved):
thanks

PR #369, opened 2025-01-27
tpm2: Directly authenticate the role field of the metadata.

chrisccoulson:
The role field currently isn't authenticated directly becuase it's used to generate the policyRef for the `TPM2_PolicyAuthorize` command. However, the polcyRef is stored in the metadata, and so modifications to the role field of a keyslot metadata will only be detected by calling a method of `SealedKeyData` that calls in to `keyData.ValidateData`. In this sense, it is authenticated indirectly.

However, we might make security decisions based on the role of a key used for unlocking, so the current approach isn't sufficient - it needs to be authenticated directly during unlocking.

### DIFF
--- a/tpm2/keydata.go
+++ b/tpm2/keydata.go
@@ -87,7 +87,7 @@ type keyData interface {
 
 	// Decrypt performs authenticated decryption of the encrypted payload and the associated data.
 	// This is relevant only for keydata versions 3 and later.
-	Decrypt(key, payload []byte, generation uint32, kdfAlg tpm2.HashAlgorithmId, authMode secboot.AuthMode) ([]byte, error)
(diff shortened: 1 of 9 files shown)


pedronis (approved):
thanks, this change now needs to be used by snapd before we go live with v3 right? 

cc @valentindavid

chrisccoulson:
> thanks, this change now needs to be used by snapd before we go live with v3 right?
> 
> cc @valentindavid

It will. I'm going to push one more PR as well to add another branch to the policy for the NV counter index to make it compatible with the logic in [github.com/canonical/go-tpm2/policyutil.Policy.Execute](https://pkg.go.dev/github.com/canonical/go-tpm2@v1.11.1/policyutil#Policy.Execute) in the future, in order to avoid future pain of having to change the policy.

sespiros:
@chrisccoulson I see this is merged already but can you also make the relevant changes to SE041 to keep it up-to-date? I see its last updated date is a year ago so I am not sure if there were other changes that were done and not reflected there since (that I missed).

PR #370, opened 2025-01-27
tpm2: Update the policy for the PCR policy counter index

chrisccoulson:
This adds a new branch which permits using the `TPM2_PolicyNV` command
against the index without any authorization, much like there is already
a branch that permits `TPM2_NV_Read`.

This will make it compatible with automatic branch selection when the
`tpm2` package starts to make use of `Policy.Execute` from the
`github.com/canonical/go-tpm2/policyutil` package in the future, without
having to retrospectively modify the policy of NV indexes on existing
devices.


chrisccoulson:
Note that most of the new code in this PR is a clone of the code in `tpm2/policy_v0`, with a few minor tweaks.

tpm2/policy_v3.go
```diff
@@ -148,10 +158,140 @@ type staticPolicyData_v3 struct {
...
+// It is a flattened tree which suitable for serializing - the tree is just
+// a slice of nodes, with each node specifying an offset to its parent node.
+type policyOrData_v3 = policyOrData_v0
+
 // pcrPolicyData_v3 represents version 3 of the PCR policy metadata for
 // executing a policy session, and can be updated. It has the same format
 // as version 2.
-type pcrPolicyData_v3 = pcrPolicyData_v2
+type pcrPolicyData_v3 struct {
+	Selection                 tpm2.PCRSelectionList
```
pedronis, line 170: couldn't this just embed pcrPolicyData_v2 and replace below only the methods that have changes? it's very hard to spot what are difference between the methods are atm, are they mostly in executeRevocationCheck ?

chrisccoulson: Yes, the changes are only in `executeRevocationCheck`. I'll try what you suggested.

chrisccoulson: Ok, I've done this now, although it did require a bit more work in the unit tests.

tpm2/policy_v3.go
```diff
@@ -148,10 +158,61 @@ type staticPolicyData_v3 struct {
...
+	authPolicies, err := computeV3PcrPolicyCounterAuthPolicies(counter.Name().Algorithm(), updateKey)
+	if err != nil {
+		return policyDataError{fmt.Errorf("cannot compute auth policies for PCR policy counter: %w", err)}
+	}
+	if err := tpm.PolicyCommandCode(revocationCheckSession, tpm2.CommandPolicyNV); err != nil {
+		return err
+	}
+	if err := tpm.PolicyOR(revocationCheckSession, authPolicies); err != nil {
+		return err
+	}
```
pedronis, line 200: it would be good to have a comment here that explain why we create a new session and what we do with it

pedronis (commented):
thanks for the change, one comment

tpm2/policy_v3.go
chrisccoulson: I've added a comment now.

PR #372, opened 2025-01-29
efi/preinstall: Add missing test for expected CheckResultFlags

chrisccoulson:
The unit tests were supplying expected flags but the tests were not
actually testing that the returned value matches.


pedronis (approved):
oops, thank you

PR #373, opened 2025-01-29
efi/preinstall: Fix broken SHA1 checks

chrisccoulson:
PR #365 introduced support for SHA1 PCR banks, hidden behind a flag (`PermitWeakPCRBanks`), and some additional unit tests for this. However, the unit tests are currently testing the wrong thing - they are enabling and testing for SHA256 instead.

This updates the tests to make sure they are testing the correct thing.


pedronis (approved):
thanks

PR #376, opened 2025-01-30
efi/preinstall: Include crypto/sha1 package

chrisccoulson:
As SHA1 is a PCR algorithm that this package can choose (with the `PermitWeakPCRBanks` flag enabled), this package should include the `crypto/sha1` package so that the behaviour is consistent with other algorithms where we already include those packages for supported algorithms.


pedronis (approved):
+1

PR #378, opened 2025-01-31
internal/testenv: Rename IsTestBinary

chrisccoulson:
It is now MustBeTestBinary, and will panic if called from something that isn't a test binary.


pedronis (approved):
+1

PR #380, opened 2025-02-02
efi/preinstall: Fail by default if PCR0 is unusable, we have a dTPM and the startup locality is 0

chrisccoulson:
If we have a discrete TPM (dTPM) and the `TPM2_Startup` command was executed from locality 3, or there was a H-CRTM event, we return an error by default if the corresponding localities 3 or 4 are not locked down by the silicon vendor's firmware before handing over to the platform firmware (we assume Intel platforms with TXT do lock these down because TXT provides the primitives to create a DRTM), returning a `ErrTPMStartupLocalityNotProtected` error to indicate that we can't enable partial mitigation of PCR replay from software in the event of the dTPM being reset independently of the host platform (either because a platform integration bug which makes this trivially possible with a pair of tweezers or because an adversary is determined enough to perform the necessary hardware tampering to trigger a reset of the dTPM independently of the host platform), after which they are able execute commands (eg, `TPM2_Startup`) at their desired locality from ring 0 in order to replay arbitrary PCR values from software and spoof any host platform.

We provide a flag (`PermitNoDiscreteTPMResetMitigation`) to make this check non-fatal if desired. In this case, the returned result will have the `StartupLocalityNotProtected` flag to tell `WithAutoTCGPCRProfile` not to include PCR0.
The next paragraph describes the relevance of PCR0.

Our partial mitigation is achieved by including PCR0 in the policy if `TPM2_Startup` was not executed from locality 0, because it changes the initial value of the PCR. If `TPM2_Startup` occurred from locality 3 or there was a H-CRTM event, and access to these localities is subsequently restricted by the silicon vendor's firmware before handing over to the platform manufacturer's firmware, it's impossible to replay the PCR0 values from software in the event of a dTPM being reset independently of the host platform, even from platform firmware code or other ring 0 code. Of course, a more determined adversary can still spoof a host platform by executing commands at any locality if they are able to actively communicate directly with the dTPM's bus, which is why it is only a partial mitigation, and which is why there is an
option to opt out of it.

Previously, the checks were always succeeding but adding the `StartupLocalityNotProtected` flag to the returned results if `TPM2_Startup` was executed from locality 0. This is clearly a case where the partial mitigation cannot be enabled, as the initial value of PCR0 in this case is 0, making it trivial to replay from software in the event of the dTPM being reset independently of the host platform. This behaviour is wrong though. If a dTPM is detected and `TPM2_Startup` occurred from locality 0, the default behaviour should be to return an error (`ErrTPMStartupLocalityNotProtected`), with the option to use the existing flag (`PermitNoDiscreteTPMResetMitigation`) to make this non-fatal.


chrisccoulson:
This is a fairly short PR, but with a very long description.

pedronis (approved):
+1

PR #384, opened 2025-02-25
internal/luks2: Fix parsing of requirements

chrisccoulson:
The LUKS2 metadata format spec says that this field is an array of strings, but cryptsetup actually defines it as an object. To avoid failing to parse headers, copy what cryptsetup does.


pedronis (approved):
thank you. Do we need to worry about supporting the previous format at all?

internal/luks2/metadata.go
```diff
@@ -360,20 +360,26 @@ func (n JsonNumber) Uint64() (uint64, error) {
 	return strconv.ParseUint(string(n), 10, 64)
 }
 
+// Requirements corresponds to the requirements object in the JSON metadata of
+// a LUKS2 volume.
```
bboozzoo, line 364: given that there's discrepancy between the official pdf and the code, this could use a reference to upstream repo: https://gitlab.com/cryptsetup/cryptsetup/-/blob/93bc8997eaf6168bd196c82cb3adbd28b8d2ac56/lib/luks2/luks2_json_metadata.c?page=2#L1065 unless there's a better place where metadata is defined

bboozzoo (approved):

chrisccoulson:
> thank you. Do we need to worry about supporting the previous format at all?

No - I checked as far back as cryptsetup v2.0.0, the first release with LUKS2 support, and it's always been implemented like this, slightly contrary to the spec.

internal/luks2/metadata.go
chrisccoulson: I've added a comment now.

PR #386, opened 2025-02-26
efi/preinstall: Add some more TPM tests

chrisccoulson:
This adds 2 more TPM tests:
- That no hierarchy has an authorization policy set.
- That the TPM is not in failure mode.

It also moves the lockout check until after we've checked that all of
the hierarchies are unowned. This is because it simplifies potential
options for rectifying this failure later.

Also, the hierarchy ownership checks are rolled into a new error
(TPM2OwnedHierarchiesError) which can capture multiple hierarchies in a
single error, which is a bit more useful.

### DIFF
--- a/efi/preinstall/check_tpm.go
+++ b/efi/preinstall/check_tpm.go
@@ -83,7 +83,8 @@ func openAndCheckTPM2Device(env internal_efi.HostEnvironment, flags checkTPM2Dev
 	// testing - the important detail is that this statement only applies to full self-tests, which
 	// we aren't requesting here (and don't want to - we only want to test functionality that hasn't
 	// been tested yet). For now, we'll assume that the same statement applies also to
-	// TPM2_SelfTest(NO) and only support the case where it runs the tests synchronously - we can
+	// TPM2_SelfTest(NO) (there's no good reason why an implementation would implement both cases
+	// differently) and only support the case where it runs the tests synchronously - we can
 	// uncomment the other path later on if we come across implementations that need it, perhaps
 	// with some sort of timeout.
 	err = tpm.SelfTest(false)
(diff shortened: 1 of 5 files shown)


efi/preinstall/check_tpm.go
```diff
@@ -67,6 +68,54 @@ func openAndCheckTPM2Device(env internal_efi.HostEnvironment, flags checkTPM2Dev
...
+			switch rc {
+			case tpm2.ResponseFailure:
+				// One or more tests failed and the TPM is now in failure mode.
+				return nil, false, ErrTPMFailure
+			case tpm2.ResponseTesting:
+				// The tests are still running. We need to wait for a bit and
+				// then try again.
+			default:
+				return nil, false, fmt.Errorf("unexpected self test result: %#x", rc)
+			}
```
pedronis, line 103: do we need to timeout at some point and consider that a failure?

pedronis (commented):
question

efi/preinstall/check_tpm.go
chrisccoulson: We could do - I'm not sure what to set it to though.

pedronis: maybe it needs to be passed in from outside? do we have a sense of how long this kind of self-testing takes?

chrisccoulson: I don't - there are many factors here - eg, firmware based (faster) vs discrete (likely slower) and what tests are left to run (did the firmware execute `TPM2_SelfTest` before, or has functionality been self-tested incrementally as it's used?).

For now, I've commented out this particular code path with a comment explaining why it's likely to be ok to not have it. I've left the code there though for now, and we can uncomment it and re-open the timeout question another day if we discover that we do actually need it - I suspect the best thing will be to just use the `context.Context` that's already passed into the public API.

pedronis (approved):
+1

PR #387, opened 2025-02-26
efi/preinstall Add initial RunChecksContext API

chrisccoulson:
This API is easier to use than the other exported API (RunChecks) and
converts errors into error kinds that are suitable for representing in
JSON.

This API will eventually map certain error kinds to actions that can be
performed by this package in order to resolve some issues. The intention
here is that the installer UI eventually gives the user some information
about why FDE cannot be enabled, and it gives them the option of choosing
to fix some issues where it is possible and appropriate to do so (eg, say
the TPM device is detected but disabled, there should be an action to
enable it via the PPI).

The Action type and some pseudo actions are already defined - these are
actions that cannot be performed by this package, but are hints for
actions that can be performed by the caller, eg, shutdown, reboot and
reboot to firmware settings are some examples.

RunChecksContext will keep track of which actions are expected (based on
the previous response) and which actions are enabled - some actions are
enabled by default but may become disabled later on because of an error,
and some other actions will only be made available after performing
tests that indicate the action can be performed.

### DIFF
--- a/efi/preinstall/actions.go
+++ b/efi/preinstall/actions.go
@@ -1,0 +1,83 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2024 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package preinstall
+
+// Action describes an Action to resolve a detected error. Some [ErrorKind]s may
+// be associated with one or more Actions that can be taken in order to resolve
+// the error. The code that calls [RunChecksContext.Run] can respond with one of
+// these actions.
+//
+// An installer UI may offer some of these actions in response to detected errors,
+// although note that it doesn't have to offer all actions associated with an error,
+// and the documentation for some actions provide hints as to whether they are
+// inappropriate for an installer UI. In some cases, it may be appropriate for snapd
+// to choose an action as opposed to exposing it to the installer UI.
+//
+// TODO: Add some meaningful actions here later on.
+type Action string
+
+const (
+	// ActionNone corresponds to no action.
+	ActionNone Action = ""
+
+	// ActionReboot corresponds to rebooting the device. Note that this is a
+	// pseudo-action. It cannot be performed by this package - the caller
+	// should trigger the reboot.
+	ActionReboot Action = "reboot"
+
+	// ActionShutdown corresponds to shutting down the device. Note that this
+	// is a pseudo-action. It cannot be performed by this package - the caller
+	// should trigger the shutdown.
+	ActionShutdown Action = "shutdown"
+
+	// ActionRebootToFWSettings corresponds to rebooting the device to the firmware
+	// settings in order to resolve a problem manually. Note that this is a
+	// pseudo-action. It cannot be performed by this package - the caller should
+	// trigger the reboot to FW settings.
+	//
+	// TODO: How do we improve this by offering a hint of what needs to change
+	// from entering the firmware settings UI?
+	ActionRebootToFWSettings Action = "reboot-to-fw-settings"
+
+	// ActionContactOEM is a hint that the user should contact the OEM for the
+	// device because of a bug in the platform. It is a pseudo-action and cannnot
+	// be performed by this package.
+	ActionContactOEM Action = "contact-oem"
+
+	// ActionContactOSVendor is a hint that the user should contact the OS vendor
+	// because of a bug in the OS. It is a pseudo-action and cannnot be performed
+	// by this package.
+	ActionContactOSVendor Action = "contact-os-vendor"
+)
+
+// IsPseudoAction will return true if the action cannot actually be executed by
+// [RunChecksContext.Run], but the action is expected to be performed by the caller
+// (eg, snapd or the installer) instead.
+//
+// TODO: Add extra actions that can be performed by this package by passing the
+// action to [RunChecksContext.Run].
+func (a Action) IsPseudoAction() bool {
+	switch a {
+	case ActionReboot, ActionShutdown, ActionRebootToFWSettings, ActionContactOEM, ActionContactOSVendor:
+		return true
+	default:
+		return false
+	}
+}
--- a/efi/preinstall/actions.go
+++ b/efi/preinstall/actions.go
@@ -67,13 +67,13 @@ const (
 	ActionContactOSVendor Action = "contact-os-vendor"
 )
 
-// IsPseudoAction will return true if the action cannot actually be executed by
+// IsExternalAction will return true if the action cannot actually be executed by
 // [RunChecksContext.Run], but the action is expected to be performed by the caller
 // (eg, snapd or the installer) instead.
 //
 // TODO: Add extra actions that can be performed by this package by passing the
 // action to [RunChecksContext.Run].
-func (a Action) IsPseudoAction() bool {
+func (a Action) IsExternalAction() bool {
 	switch a {
 	case ActionReboot, ActionShutdown, ActionRebootToFWSettings, ActionContactOEM, ActionContactOSVendor:
 		return true
--- a/efi/preinstall/checks.go
+++ b/efi/preinstall/checks.go
@@ -270,7 +270,9 @@ func RunChecks(ctx context.Context, flags CheckFlags, loadedImages []secboot_efi
 
 	logResults, err := checkFirmwareLogAndChoosePCRBank(tpm, log, mandatoryPcrs, checkLogFlags)
 	switch {
-	case tpm2.IsTPMError(err, tpm2.AnyErrorCode, tpm2.AnyCommandCode):
+	case tpm2.IsTPMError(err, tpm2.AnyErrorCode, tpm2.AnyCommandCode) ||
+		tpm2.IsTPMWarning(err, tpm2.AnyWarningCode, tpm2.AnyCommandCode) ||
+		isInvalidTPMResponse(err) || isTPMCommunicationError(err):
 		return nil, &TPM2DeviceError{err}
 	case isEmptyPCRBanksError(err):
 		// Save this error and return it unwrapped when the checks complete
@@ -461,6 +463,7 @@ func RunChecks(ctx context.Context, flags CheckFlags, loadedImages []secboot_efi
 			}
 			result.UsedSecureBootCAs = pcr7Result.UsedAuthorities
 
+			// Only return these errors if PCR7 is required.
 			if result.Flags&WeakSecureBootAlgorithmsDetected > 0 && flags&PermitWeakSecureBootAlgorithms == 0 {
 				// We don't support weak secure boot verification algorithms
 				deferredErrs = append(deferredErrs, ErrWeakSecureBootAlgorithmDetected)
--- a/efi/preinstall/checks.go
+++ b/efi/preinstall/checks.go
@@ -29,10 +29,14 @@ import (
 	internal_efi "github.com/snapcore/secboot/internal/efi"
 )
 
-// CheckFlags can be used to customize the behaviour or [RunChecks].
+// CheckFlags can be used to customize the behaviour or [RunChecks] and [NewRunChecksContext].
 type CheckFlags int
 
 const (
+	// CheckFlagsDefault is the default flags for RunChecks and
+	// NewRunChecksContext if no other flags are supplied.
+	CheckFlagsDefault CheckFlags = 0
+
 	// PlatformFirmwareProfileSupportRequired indicates that support for
 	// [secboot_efi.WithPlatformFirmwareProfile] to generate profiles for
 	// PCR 0 is not optional.
--- a/efi/preinstall/checks_context.go
+++ b/efi/preinstall/checks_context.go
@@ -31,6 +31,9 @@ import (
 	internal_efi "github.com/snapcore/secboot/internal/efi"
 )
 
+// errorKindToActions maps an error kinds to one or more possible actions. The
+// slice of actions is an OR in the sense that any one of these actions can
+// be eecuted to attempt to resolve the associated error kind.
 var errorKindToActions map[ErrorKind][]Action
 
 func init() {
--- a/efi/preinstall/checks_context.go
+++ b/efi/preinstall/checks_context.go
@@ -175,7 +175,6 @@ func NewRunChecksContext(initialFlags CheckFlags, loadedImages []secboot_efi.Ima
 			ActionContactOEM:         true,
 			ActionContactOSVendor:    true,
 		},
-		expectedActions: []Action{ActionNone},
 	}
 }
 
@@ -217,6 +216,10 @@ func (c *RunChecksContext) filterUnavailableActions(actions []Action) (out []Act
 // isActionExpected determines if the supplied action is an expected
 // response to the last call to [RunChecksContext.Run].
 func (c *RunChecksContext) isActionExpected(action Action) bool {
+	if action == ActionNone {
+		// ActionNone is always expected.
+		return true
+	}
 	for _, expected := range c.expectedActions {
 		if expected == action {
 			return true
@@ -483,7 +486,7 @@ func (c *RunChecksContext) Run(ctx context.Context, action Action, args ...any) 
 		return nil, err
 	}
 
-	c.expectedActions = []Action{ActionNone}
+	c.expectedActions = nil
 	var errs []error
 	for {
 		result, err := RunChecks(ctx, c.flags, c.loadedImages)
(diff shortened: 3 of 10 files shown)


efi/preinstall/checks_context.go
```diff
@@ -0,0 +1,585 @@
...
+	env          internal_efi.HostEnvironment
+	flags        CheckFlags
+	loadedImages []secboot_efi.Image
+	profileOpts  PCRProfileOptionsFlags
+
+	errs    []error
+	lastErr error
+	result  *CheckResult
+
+	availableActions map[Action]bool
```
pedronis, line 142: this probably merits a comment as it seems it can store both true and false entries which is not always the case with such maps, sometimes they are really meant to be sets (so either no entry or true)

efi/preinstall/checks_context.go
```diff
@@ -0,0 +1,585 @@
...
+
+// isActionExpected determines if the supplied action is an expected
+// response to the last call to [RunChecksContext.Run].
+func (c *RunChecksContext) isActionExpected(action Action) bool {
+	for _, expected := range c.expectedActions {
+		if expected == action {
+			return true
+		}
+	}
+	return false
```
pedronis, line 222: tests don't seem to hit the false case

efi/preinstall/checks_context.go
```diff
@@ -0,0 +1,585 @@
...
+		// unless we discover that they aren't later on.
+		availableActions: map[Action]bool{
+			ActionNone:               true,
+			ActionReboot:             true,
+			ActionShutdown:           true,
+			ActionRebootToFWSettings: true,
+			ActionContactOEM:         true,
+			ActionContactOSVendor:    true,
+		},
+		expectedActions: []Action{ActionNone},
```
pedronis, line 175: do we need to set this as is also set in Run?

pedronis (commented):
did a first pass, didn't look at tests yet

efi/preinstall/checks_context.go
```diff
@@ -0,0 +1,585 @@
...
+
+// Run will run the specified action, and if that completes successfully will run another
+// iteration of [RunChecks] and test the result against the preferred [WithAutoTCGPCRProfile]
+// configuration. On success, this will return the CheckResult. On failure, this will return
+// an error which will either be a single ErrorKindAndActions, or multiple ErrorKindAndActions
+// wrapped by an error type that implements the [CompoundError] interface. If there are any
+// actions associated with an error, the install environment may try one or more of them in
+// order to try to resolve the issue that caused the error. In some cases, it may be appropriate
+// to ask permission from the user to perform an action.
+func (c *RunChecksContext) Run(ctx context.Context, action Action, args ...any) (*CheckResult, error) {
```
ernestl, line 478: Should we consider returning a compound error always instead of mixing it up, to not require inspection to determine what we are dealing with?

efi/preinstall/errors.go
```diff
@@ -814,3 +823,25 @@ func (e *UnsupportedRequiredPCRsError) Error() string {
 		return fmt.Sprintf("PCRs %v are required, but are unsupported", e.PCRs)
 	}
 }
+
+// ErrorKindAndActions is an error type that can be serialized to JSON, represented by
+// an error kind, associated arguments and a set of potential remedial actions.
+type ErrorKindAndActions struct {
```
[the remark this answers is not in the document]
chrisccoulson: That sounds ok with me.

efi/preinstall/errors.go
```diff
@@ -814,3 +823,25 @@ func (e *UnsupportedRequiredPCRsError) Error() string {
...
+	ErrorKind ErrorKind       `json:"kind"`    // The error kind
```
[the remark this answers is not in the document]
chrisccoulson: I think it's ok for me to just rename these.

efi/preinstall/actions.go
```diff
@@ -0,0 +1,83 @@
...
+	ActionContactOSVendor Action = "contact-os-vendor"
+)
+
+// IsPseudoAction will return true if the action cannot actually be executed by
+// [RunChecksContext.Run], but the action is expected to be performed by the caller
+// (eg, snapd or the installer) instead.
+//
+// TODO: Add extra actions that can be performed by this package by passing the
+// action to [RunChecksContext.Run].
+func (a Action) IsPseudoAction() bool {
```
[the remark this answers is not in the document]
chrisccoulson: Yes, that's ok - I wasn't sure of what name to use here.

efi/preinstall/checks.go
```diff
@@ -270,7 +270,9 @@ func RunChecks(ctx context.Context, flags CheckFlags, loadedImages []secboot_efi
 
 	logResults, err := checkFirmwareLogAndChoosePCRBank(tpm, log, mandatoryPcrs, checkLogFlags)
 	switch {
-	case tpm2.IsTPMError(err, tpm2.AnyErrorCode, tpm2.AnyCommandCode):
+	case tpm2.IsTPMError(err, tpm2.AnyErrorCode, tpm2.AnyCommandCode) ||
+		tpm2.IsTPMWarning(err, tpm2.AnyWarningCode, tpm2.AnyCommandCode) ||
+		isInvalidTPMResponse(err) || isTPMCommunicationError(err):
```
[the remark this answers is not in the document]
chrisccoulson: There aren't - I can probably add some more tests for this.

There are a few known cases of missing test coverage, but I have left some notes throughout the test files and in checks_context.go where I know test coverage is missing - it's mostly just for the harder-to-test edge cases and I'm going to fill in the gaps (maybe as a follow-up PR)

efi/preinstall/checks_context.go
```diff
@@ -0,0 +1,585 @@
...
+	if errors.Is(err, ErrPreOSVerificationUsingDigests) {
+		return ErrorKindPreOSDigestVerificationDetected, nil, nil
+	}
+
+	var sbPcrErr *SecureBootPolicyPCRError
+	if errors.As(err, &sbPcrErr) {
+		return ErrorKindPCRUnusable, []any{internal_efi.SecureBootPolicyPCR}, nil
+	}
+
+	return ErrorKindInternal, nil, nil
```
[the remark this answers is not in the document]
chrisccoulson: Yes, I need to figure out the best way of generating an error that gets here. I'll take a look.

efi/preinstall/checks_context.go
chrisccoulson: It's reset in `Run` after running the supplied action - it's set initially here to bootstrap it otherwise `Run` would fail with an unexpected action. It might be better to just clear `expectedActions` and have `ActionNone` as implicitly accepted instead.

efi/preinstall/error_kinds.go
```diff
@@ -0,0 +1,216 @@
...
+ */
+
+package preinstall
+
+// ErrorKind describes an error detected during preinstall checks when
+// using the [RunChecksContext] API.
+type ErrorKind string
+
+const (
+	// ErrorKindNone indicates that no error occurred.
```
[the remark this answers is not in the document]
chrisccoulson: Ah, I thought I'd done this - I'll have another scan through to see where I've missed them.

efi/preinstall/checks_context.go
```diff
@@ -0,0 +1,585 @@
...
+		for _, prop := range []tpm2.Property{tpm2.PropertyLockoutCounter, tpm2.PropertyLockoutInterval} {
+			val, err := tpm.GetCapabilityTPMProperty(prop)
+			if err != nil {
+				return ErrorKindNone, nil, fmt.Errorf("cannot read property %d: %w", prop, err)
+			}
+			vals = append(vals, val)
+		}
+		lockoutCounter = vals[0]
+		lockoutInterval = vals[1]
+		return ErrorKindTPMDeviceLockout, []any{time.Duration(lockoutInterval) * time.Second, time.Duration(lockoutInterval) * time.Second * time.Duration(lockoutCounter)}, nil
```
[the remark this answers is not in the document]
chrisccoulson: That's fine with me.

efi/preinstall/checks_context.go
```diff
@@ -0,0 +1,585 @@
...
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/canonical/go-tpm2"
+	secboot_efi "github.com/snapcore/secboot/efi"
+	internal_efi "github.com/snapcore/secboot/internal/efi"
+)
+
+var errorKindToActions map[ErrorKind][]Action
```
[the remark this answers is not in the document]
chrisccoulson: I can add some comments, but it's an OR rather than a sequence. There aren't any cases where a sequence of actions is required. There might be more than one error - each with an OR of associated actions, in which case a single action from each error can be executed, but they can be executed in any order.

efi/preinstall/checks_context.go
```diff
@@ -0,0 +1,585 @@
...
+// should match the flags later passed to this function when creating a PCR profile. The PCRs that are
+// determined to be required to build the profile using [WithAutoTCGPCRProfile] will be made mandatory
+// automatically by this function by passing the relevant flags to [RunChecks].
+//
+// As [PCRProfileOptionsFlags] is intended to provide some limited amount of user customisation, this
+// API [RunChecksContext] should be executed again with the new set of flags, should the user wish to
+// change them. In this case, the caller should pass the PostInstallChecks flag as an initial flag.
+//
+// There is no need for the caller to supply any of these *SupportRequired flags as the initial flags,
+// and this may have the effect of limiting the number of devices which pass the checks.
```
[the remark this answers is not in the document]
chrisccoulson: That is ok to me.

efi/preinstall/checks_context.go
chrisccoulson: This is intended to have 3 states - empty implies that we need to test that the action is available, false is "tested, but unavailable" and true is "available".

efi/preinstall/checks_context.go
chrisccoulson: I can add a test for this.

efi/preinstall/checks_context.go
```diff
@@ -0,0 +1,585 @@
...
+		var requiredPCRsErr *UnsupportedRequiredPCRsError
+		if !errors.As(profileErr, &requiredPCRsErr) {
+			return nil, &ErrorKindAndActions{
+				ErrorKind: ErrorKindInternal,
+				err:       fmt.Errorf("cannot test whether a PCR combination can be generated: %w", err),
+			}
+		}
+
+		// Make any PCRs we're lacking support for mandatory so that they end
+		// up being returned in the RunChecks error on the next iteration,
```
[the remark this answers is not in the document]
chrisccoulson: Yes, I think that's the case.

efi/preinstall/checks_context.go
```diff
@@ -0,0 +1,585 @@
...
+				c.flags |= SecureBootPolicyProfileSupportRequired
+			}
+		}
+	}
+
+	if c.result != nil {
+		return c.result, nil
+	}
+
+	return nil, joinErrors(errs...)
```
ernestl, line 584: This is more a comment on joinErrors.Error(), and also taking into consideration the suggestion to always return a compound error:

From the snapd side, at the point where we receive the error we are firstly interested to know that a preinstall-check error was detected, without needing all the detail about possible multiple errors, with args and actions. 

Could we perhaps provide the compound error with a concise high level error message without all the details of multiple errors, args and actions?

More specifically, on the snapd side, we have the concept of a unavailable reason (short error message) and a separate field for slice of error information that is only exposed at API level. It would be convenient if the compound error [can be used for the unavailable reason as is](https://github.com/canonical/snapd/pull/15373/files#diff-8133fed769942c6fe6d9075fbd0253dffcba86bc3fc94b96e1a546c7f707a416R387).

ernestl:
Hi @chrisccoulson, 

Regarding the topic of ensuring that the [existing preinstall check](https://github.com/canonical/snapd/blob/master/secboot/secboot_tpm.go#L88) is sufficiently covered by the new preinstall check.

(1) is efi secureboot enabled [read efi var](https://github.com/canonical/snapd/blob/master/secboot/secboot_tpm.go#L121)
(2) check if tpm available (connect)
(3) check if tpm enabled
(4) if mode == TPMProvisionFull, ensure [tpm is not locked out](https://github.com/canonical/snapd/blob/master/secboot/secboot_tpm.go#L109)

- 1 is not covered, but secboot has capability to read efi vars, so something to add.
- In previous discussion we said 2  and 3 likely sufficiently covered, to confirm.

However, we do not have equivalency in terms of considering `tpm mode`. 
I want to propose that we add `tpm mode`  as parameter for run-checks to allow for more granular checks.

efi/preinstall/checks.go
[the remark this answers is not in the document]
chrisccoulson: If it's ok, I'm going to add some test coverage here in a follow-up. Properly fixing this requires an update to go-tpm2 and I already have a PR with some changes in to accommodate this - it would be a bit easier to land these extra tests after that.

efi/preinstall/checks.go
ernestl, line 1: Consider CheckFlags constant for default (0), same as what is available for PCRProfileOptionsFlags.

efi/preinstall/errors.go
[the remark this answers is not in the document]
chrisccoulson: I've renamed this now.

efi/preinstall/errors.go
```diff
@@ -814,3 +823,25 @@ func (e *UnsupportedRequiredPCRsError) Error() string {
...
+
+	err error `json:"-"` // The original error. This is not serialized to JSON.
+}
+
+func (e *ErrorKindAndActions) Error() string {
+	data, err := json.Marshal(e)
+	if err != nil {
+		data = []byte(fmt.Sprintf("cannot serialize error: %v", err))
+	}
+	return fmt.Sprintf("%v %s", e.err, string(data))
```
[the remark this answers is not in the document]
chrisccoulson: I've removed the JSON from the String implementation.

efi/preinstall/checks_context.go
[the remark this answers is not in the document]
chrisccoulson: I've added a doc comment to this now.

efi/preinstall/checks_context.go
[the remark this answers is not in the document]
chrisccoulson: I've added a test case that tests this specific sequence now.

efi/preinstall/checks_context.go
chrisccoulson: I've added a doc comment for this now.

efi/preinstall/checks_context.go
chrisccoulson: I've added a test case with an invalid action name now.

efi/preinstall/checks_context.go
chrisccoulson: I've changed the semantics of `expectedActions` now so that `ActionNone` is always expected, even when set to nil.

efi/preinstall/errors.go
ernestl, line 1: Multiple cases of ` wrapped wrapped`

efi/preinstall/error_kinds.go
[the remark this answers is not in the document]
chrisccoulson: I think each one of these with arguments is documented now and with a description of what the arguments look like.

chrisccoulson:
> Hi @chrisccoulson,
> 
> Regarding the topic of ensuring that the [existing preinstall check](https://github.com/canonical/snapd/blob/master/secboot/secboot_tpm.go#L88) is sufficiently covered by the new preinstall check.
> 
> (1) is efi secureboot enabled [read efi var](https://github.com/canonical/snapd/blob/master/secboot/secboot_tpm.go#L121) (2) check if tpm available (connect) (3) check if tpm enabled (4) if mode == TPMProvisionFull, ensure [tpm is not locked out](https://github.com/canonical/snapd/blob/master/secboot/secboot_tpm.go#L109)
> 
> In a previous discussion we agreed:
> 
> * 1 is not covered, but secboot has capability to read efi vars, so something to add.
> * In previous discussion we said 2  and 3 likely sufficiently covered, to confirm.
> 
> However, we do not have equivalency in terms of considering `tpm mode`. I want to propose that we add `tpm mode` as parameter for run-checks to allow for more granular checks.

There are explicit tests already for 2-4. The only one there isn't an explicit test for is 1, although some tests do read EFI variables and go-efilib generates a fairly consistent error if efivarfs isn't available ([efi.ErrVarsUnavailable])

efi/preinstall/checks_context.go
chrisccoulson: I need to have a bit more of a think about this one as it's a bit more complicated than I initially realized because [RunChecksContext] is keeping a log of returned errors and I was going to do this by adding a deferred function to do the wrapping.

efi/preinstall/errors.go
chrisccoulson: These are fixed now.

efi/preinstall/checks_context.go
chrisccoulson: I'm not sure I've fully understood the suggestion here, but I'm not sure I'm keen on the idea of updating the `Error()` implementation to hide information about errors - a lot of information in these errors already isn't going to be serialized via `WithKindAndActionsError` and it would be useful for debugging in some cases to see the full error message in a log somewhere.

ernestl:
Discussed with Chris the idea of also passing warnings to installer via snapd /v2/systems/{label} api.
The implication if we did it:
 - warnings would need to classified to kind
 - end up in same error type as errors with actions, but without args and actions

Chris advised that this is beyond what was intended for warnings, which is only to allow snapd to log it for debugging purposes, "as a tool for extracting test failures that don't otherwise prevent the install from continuing, so in that sense, they only really exist for debugging purposes".

@pedronis

efi/preinstall/checks_context.go
ernestl: This was a convenience request, it's ok as is.

efi/preinstall/checks.go
ernestl: Thanks

efi/preinstall/errors.go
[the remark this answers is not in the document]
ernestl: Thanks

ernestl: Thanks

PR #390, opened 2025-03-20
auth_requestor_systemd.go: do not use text/template

valentindavid:
text/template brings 4MB into snap-bootstrap. That makes some customers not able to update their kernels.

### DIFF
--- a/auth_requestor_systemd.go
+++ b/auth_requestor_systemd.go
@@ -22,25 +22,18 @@ package secboot
 import (
 	"bytes"
 	"errors"
+	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
-	"text/template"
 
 	"golang.org/x/xerrors"
 )
 
-type askPasswordMsgParams struct {
-	VolumeName       string
-	SourceDevicePath string
-	// PartLabel string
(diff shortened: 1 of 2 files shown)


alfonsosanchezbeato:
For reference: https://github.com/canonical/snapd/pull/15226

alfonsosanchezbeato (approved):
LGTM

alfonsosanchezbeato:
FTR I have double checked and this PR plus the snapd one does produce the 4MB reduction.

pedronis (approved):
+1

valentindavid:
Pleas someone ping me when this is merged. So I can open the PR on snapd that updates the templates and updates secboot.

chrisccoulson (approved):
This looks ok for now, although I'll probably have a different fix for this when I refactor the API, as this isn't really compatible with the way I'd planned for this to work.

I plan to merge the 2 methods of the `AuthRequestor` interface as the new activation API won't request a passphrase or recovery key separately - it will call a single method with a set of flags that indicate which auth types are being requested based on the key slots that are in the header, allowing the user to enter either a passphrase, PIN or recovery key using a single prompt. This will probably require a bit of logic wrt how the prompt is constructed.

PR #391, opened 2025-03-26
preinstall: Fix declaration of tcglogPhase enum constants

chrisccoulson:
This doesn't have any effect on the declared enum constants, hence there
are no unit test changes. It just fixes a mistake in the way that the
constants are declared, and makes them consistent with others in this
repository.


pedronis (approved):
+1

PR #392, opened 2025-04-01
preinstall: Simplify NoSuitablePCRAlgorithmError

chrisccoulson:
NoSuitablePCRAlgorithm error had 2 fields from which to obtain errors -
a field for errors that affect an entire PCR bank, and a field for
errors that affect a single PCR in a bank.

Simplify this by having a single field for all errors, which are stored
in a map that is keyed by the PCR bank. PCR-specific errors are added to
this field wrapped in the relevant PCR-specific error type instead, so
that the PCR-specific errors are returned for any errors affecting a PCR,
regardless of whether that error occurs during or after the bank
selection.

### DIFF
--- a/efi/preinstall/errors.go
+++ b/efi/preinstall/errors.go
@@ -299,14 +299,13 @@ func (e *TPM2OwnedHierarchiesError) isEmpty() bool {
 
 var (
 	// ErrPCRBankMissingFromLog may be returned wrapped by NoSuitablePCRAlgorithmError
-	// in the event where a PCR bank does not exist in the TCG log. It may be obtained from
-	// the BankErrs field.
+	// in the event where a PCR bank does not exist in the TCG log.
 	ErrPCRBankMissingFromLog = errors.New("the PCR bank is missing from the TCG log")
 )
 
-// PCRValueMismatchError may be returned wrapped by NoSuitablePCRAlgorithmError for a specific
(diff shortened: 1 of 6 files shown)


efi/preinstall/errors.go
```diff
@@ -362,57 +361,47 @@ func (e *EmptyPCRBanksError) Error() string {
...
+//     or SHA-512).
+//   - The TCG log is inconsistent with the TPM values when the log is reconstructed for one
+//     or more mandatory PCRs, for all algorithms in the log.
+//   - PCR 0 is mandatory and there is an issue with the way that the startup locality event
+//     (if present) is recorded.
+//   - There is a problem with the sequence of measurements that isn't specific to a PCR or
+//     a PCR bank, and which is incompatible with predicting PCR policies.
+//
+// As multiple errors can occur during testing, this error wraps each individual error that
+// occurred and provides access to them.
```
pedronis, line 376: I'm not sure whether it should be added  here or somewhere else but it doesn't seem to me that the doc comments explain the wrapping in *PCRError wrapper errors?

pedronis (commented):
thanks, doc comments question

efi/preinstall/errors.go
chrisccoulson: I've added an extra bit to the documentation now, which I think is what you're looking for.

PR #394, opened 2025-04-02
preinstall: Check TCG log structure before selecting a PCR bank

chrisccoulson:
checkFirmwareLogAndChoosePCRBank currently attempts to select a PCR bank
before checking the structure of the TCG log. This is a minor change,
but it makes more sense to perform the check of the TCG log first.


pedronis (approved):
+1

PR #395, opened 2025-04-02
preinstall: Rename PlatformFirmwareProtectionError to HostSecurityError

chrisccoulson:
PlatformFirmwareProtectionError is returned not just for errors that are
related to firmware security, but in the case where the currently running
kernel doesn't have a IOMMU. Renaming it to HostSecurityError makes
a bit more sense, and it's shorter.


chrisccoulson:
This fixed  https://github.com/canonical/secboot/pull/395 - I forgot that I'd opened a Github issue for this.

PR #396, opened 2025-04-07
preinstall: return TPM command errors immediately

chrisccoulson:


pedronis (approved):
thanks for the simplification as well

PR #397, opened 2025-04-07
preinstall: only check for CPU debugging on Intel

chrisccoulson:
The check for CPU silicon-level debugging is testing the contents of a
MSR that is Intel specific. When we have checks for AMD CPUs, this test
won't apply to that (or, we may need an AMD-specific test).


pedronis (approved):
+1

PR #398, opened 2025-04-08
preinstall: Make some changes to error wrapping

chrisccoulson:
This fixes a few consistency errors. The changes are:
- TCGLogError is renamed to MeasuredBootError because not all of the
errors it wraps are errors with the log - in some cases it wraps errors
where the log is fine but the error is associated with some condition
that is indicated in the log (eg, an EV_SEPARATOR event that indicates
an error.
- ErrTPMStartupLocalityNotProtected is wrapped in HostSecurityError
because it's a class of error that is not really any different to any
error associated with the firmware configuration that affects
security.
- ErrNotAllBootManagerCodeDigestsVerified is wrapped in
BootManagerCodePCRError because the error indicates that the PCR4 tests
couldn't fully check that this PCR can be used. The other related errors
(ErrSysPrepApplicationsPresent and ErrAbsoluteComputraceActive) are left
untouched because these are problems with system configuration rather
than problems with testing that PCR4 is ok.

### DIFF
--- a/efi/preinstall/checks.go
+++ b/efi/preinstall/checks.go
@@ -169,7 +169,7 @@ var (
 // is best to supply all images that executed before ExitBootServices, in the correct order.
 //
 // This can return many types of errors. Some errors are returned immediately, such as [ErrVirtualMachineDetected],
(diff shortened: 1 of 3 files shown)


pedronis (approved):
+1

PR #400, opened 2025-04-09
preinstall: use a single type for wrapping multiple errors

chrisccoulson:
There are currently 2 error types that wrap multiple errors:
RunChecksErrors and joinError which is used internally. The additional
API that will come in the next PR also wants a way to wrap multiple
errors, so this removes RunChecksErrors in favour of a single type which
isn't public, and a public interface.


pedronis (approved):
+1

PR #401, opened 2025-04-15
preinstall: Move TPM discreteness check

chrisccoulson:
The TPM discreteness check relies on the BootGuard status MSR on Intel.
Move this check until after the point where we know that BootGuard is
configured properly.


chrisccoulson:
I was hoping to request Mate have a look at this, but he's out until next week and I don't think we want to block this sequence of PRs that long.

PR #403, opened 2025-04-16
preinstall: Reverse the logic of the *SupportRequired flags

chrisccoulson:
Every other flag works in reverse and has to be specified in order to
permit an error condition as ok. Reverse the logic in these flags so
that they also have to be specified in order to permit support for
specific PCRs to be missing.


pedronis (approved):
+1

PR #405, opened 2025-04-22
preinstall: return some PCR errors in more circumstances

chrisccoulson:
In the event that a PCR has an unexpected value, don't then skip the
later PCR-specific tests that can detect whether system-preparation
applications or value-added-retailer drivers are being executed.


pedronis (approved):
thanks

PR #406, opened 2025-04-22
preinstall: remove non-persistent CheckResult flags

chrisccoulson:
Instead, add a warning to the Warnings field for each of the conditions
that previously set these flags.

This also removes the ErrNotAllBootManagerCodeDigestsVerified error -
this error doesn't have a corresponding ErrorKind because it's really an
error with the way the API is used rather than an error with the
configuration of the device, and so it doesn't need a distinct error
value or an override flag.


pedronis (approved):
+1

PR #407, opened 2025-04-24
preinstall: Fix an issue with the TPM connection being closed prematurely

chrisccoulson:
This also updates the go-tpm2 version to pull in a fix that would have
meant the existing secboot tests would have caught this bug.


pedronis (approved):
+1

PR #416, opened 2025-05-30
efi/preinstall: Add explicit check that the current host system is EFI

chrisccoulson:
If the host system is not EFI, an `ErrSystemNotEFI` error will be
returned, with the error kind `ErrorKindSystemNotEFI`.

This also detects and wraps EFI variable access errors with
`EFIVariableAccessError`, which is considered fatal and returned
immediately, along with the error kind `ErrorKindEFIVariableAccess`. This
error kind is associated with an argument - `EFIVariableAccessErrorArg`,
which contains a description of the type of access error, and maps to
errors exported by `github.com/canonical/go-efilib`.


pedronis (commented):
question

pedronis (approved):
+1

PR #417, opened 2025-05-31
efi/preinstall: Provide a way to indicate that a required kernel module is missing

chrisccoulson:
Some tests rely on kernel APIs that are currently shipped in modules as
opposed to being shipped in the main kernel image. The required modules
may not be loaded in all circumstances.

Whilst I don't think that secboot, as a dependency of snapd, should be directly
loading kernel modules, there needs to be a mechanism to tell snapd which
kernel modules are required in order for the tests to run successfully. The
required kernel modules are an implementation detail of secboot and may be
architecture / platform dependent as well, so snapd shouldn't have a list of
kernel modules to load. Instead, if a feature that requires a kernel module
cannot complete because it appears that the kernel module isn't loaded,
secboot will return a `MissingKernelModuleError` which describes the module
that snapd needs to subsequently load in order for the tests to complete.

As the need to load modules is an implementation detail of the preinstall
checks in general and shouldn't be exposed to the installer, there's no specific
error kind for this - it just gets converted to `ErrorKindInternal` and snapd is
expected to just resolve this without exposing it publicly.

This fixes https://github.com/canonical/secboot/issues/409

### DIFF
--- a/efi/preinstall/errors.go
+++ b/efi/preinstall/errors.go
@@ -139,6 +139,22 @@ func (e *joinError) Unwrap() []error {
 	return e.errs
 }
 
+// MissingKernelModuleError is returned unwrapped from [RunChecks] to indicate that
+// the specified kernel module is not built as part of the currently executing kernel,
+// but is required to be loaded in order for tests to continue. The caller is expected
+// to load the required kernel module.
+type MissingKernelModuleError string
+
+func (e MissingKernelModuleError) Error() string {
+	return fmt.Sprintf("the kernel module %q must be loaded", string(e))
+}
+
+// String implements [fmt.Stringer] and returns the name of the kernel module that
+// should be loaded before calling [RunChecks].
+func (e MissingKernelModuleError) String() string {
+	return string(e)
+}
+
 var (
 	// ErrVirtualMachineDetected is returned unwrapped from RunChecks when the current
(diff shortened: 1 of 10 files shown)


efi/preinstall/errors.go
```diff
@@ -139,6 +139,22 @@ func (e *joinError) Unwrap() []error {
...
+// to load the required kernel module.
+type MissingKernelModuleError string
+
+func (e MissingKernelModuleError) Error() string {
+	return fmt.Sprintf("the kernel module %q must be loaded", string(e))
+}
+
+// String implements [fmt.Stringer] and returns the name of the kernel module that
+// should be loaded before calling [RunChecks].
+func (e MissingKernelModuleError) String() string {
```
pedronis, line 154: I wonder if this should be Module instead?

pedronis (approved):
thanks, question

efi/preinstall/errors.go
chrisccoulson: I did wonder about doing that. I think I'll keep the String method for formatting purposes but with no guarantees about the actual string value returned, and I'll add a Module method which returns just the module name.

chrisccoulson: oh, actually, I'll just replace the String method with Module - I forgot that this was an error.

PR #421, opened 2025-06-11
efi: convert loadParams struct to a map instead

chrisccoulson:
This avoids the need to add a new member to the loadParams struct for every new parameter. I would also like implementations of PCRProfileOption to be able to add a list of root options to the pcrProfileGenerator which can be used to generate a sequence of loadParams in pcrProfileGenerator.addPCRProfileBranchForVars from which multiple branches can be created.

I'm hoping that this will help the work for https://github.com/canonical/secboot/pull/412

### DIFF
--- a/efi/image.go
+++ b/efi/image.go
@@ -122,6 +122,21 @@ func (p loadParams) clone() loadParams {
 	return out
 }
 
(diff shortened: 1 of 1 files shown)


pedronis (commented):
not against, but also not sure what is the win here? This seems a win if we needed code that would go generically over the entries of loadParams for example, but I didn't spot code like that here?

PR #422, opened 2025-06-11
Add new public API `TestLUKS2ContainerKey` for testing LUKS2 container keys

kukrimate:
Also remove internal API luks2test.CheckLUKS2Passphrase and adjust existing internal users of that API to use the new one.

### DIFF
--- a/crypt.go
+++ b/crypt.go
@@ -1030,3 +1030,8 @@ func NameLegacyLUKS2ContainerKey(devicePath string, keyslot int, newName string)
 
 	return nil
 }
+
+// Check if key is valid key for LUKS2 container at devicePath.
+func TestLUKS2ContainerKey(devicePath string, key []byte) bool {
+	return luks2.TestContainerKey(devicePath, key)
+}
--- a/crypt_test.go
+++ b/crypt_test.go
@@ -3114,7 +3114,7 @@ func (s *cryptSuiteUnmockedBase) testInitializeLUKS2Container(c *C, options *Ini
 	c.Check(keyslot.KDF.Iterations, Equals, 1000)
 	c.Check(keyslot.KDF.Hash, Equals, luks2.HashSHA256)
 
-	luks2test.CheckLUKS2Passphrase(c, path, key)
+	c.Check(TestLUKS2ContainerKey(path, key), testutil.IsTrue)
 }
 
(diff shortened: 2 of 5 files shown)


chrisccoulson:
One thing that would be good would be to have a quick check to make sure that this is sufficient for the proposed snapd API in https://docs.google.com/document/d/1Mbj7Ut0tgW0PCz8VWdhvBN52kHTxYqkZZKQODB1NlLk/edit?tab=t.0#heading=h.gk1q8f5in2qh (I think it is - for some reason, I thought that the API needed a way to respond with the names of keyslots that the recovery key works for, but that doesn't seem to be the case - it seems like the API is just meant to respond that the key worked for at least 1 keyslot on a specific container)

crypt.go
```diff
@@ -1030,3 +1031,9 @@ func NameLegacyLUKS2ContainerKey(devicePath string, keyslot int, newName string)
 
 	return nil
 }
+
+func TestLUKS2ContainerKey(devicePath string, key []byte) bool {
+	cmd := exec.Command("cryptsetup", "open", "--test-passphrase", "--key-file", "-", devicePath)
```
chrisccoulson, line 1036: Rather than use `exec.Command` here, it might be worthwhile exposing an API for this from the `internal/luks2` package which this would then just wrap around. That's generally what we're doing with other APIs that require a call to cryptsetup.

crypt_test.go
```diff
@@ -3114,7 +3114,7 @@ func (s *cryptSuiteUnmockedBase) testInitializeLUKS2Container(c *C, options *Ini
 	c.Check(keyslot.KDF.Iterations, Equals, 1000)
 	c.Check(keyslot.KDF.Hash, Equals, luks2.HashSHA256)
 
-	luks2test.CheckLUKS2Passphrase(c, path, key)
+	c.Check(TestLUKS2ContainerKey(path, key), Equals, true)
```
chrisccoulson, line 3117: Secboot's `internal/testutil` package has a `IsTrue` checker, so you can do something like:
`c.Check(TestLUKS2ContainerKey(path, key), testutil.IsTrue)`.

There is also an equivalent `IsFalse` checker.

kukrimate: Okay I will update this.

pedronis (approved):
lgtm

chrisccoulson (approved):
Thanks for the updates. This is looking fine with me.

PR #425, opened 2025-06-14
luks2: Add initial LUKS2 StorageContainerBackend

chrisccoulson:
This provides an initial LUKS2 implementation of
`StorageContainerBackend`, with read-only, activate and deactivate support
for now, and the initial implementation largely just wrapping around the
existing secboot APIs.

Plans for the future for this package:
- Adding an implementation of `StorageContainerReadWriter` once it's
  defined in secboot.
- Introduction of locking to serialize access to implementations of
  `StorageContainerReader` and `StorageContainerReadWriter`.
- Moving LUKS2 functionality from the code secboot package and
  `internal/luksview` into this package, and deleting the existing code.

### DIFF
--- a/luks2/backend.go
+++ b/luks2/backend.go
@@ -113,7 +113,10 @@ func (b *storageContainerBackend) Probe(ctx context.Context, path string) (secbo
 // ProbeActivated implements [secboot.StorageContainerBackend.ProbeActivated].
 //
 // The path can be the path to a DM device that is backed by a LUKS2 container, or
-// a symbolic link to one.
+// a symbolic link to one. Using the supplied path, it will walk up the stack of
+// DM devices until it finds a block device containing a LUKS2 container. It
+// supports nesting of DM devices, although it currently only has support for the
+// crypt and linear targets.
 //
 // It identifies a [StorageContainer] by its device number, and it will always return
 // the same instance for any paths that reference the same container, regardless of
--- a/luks2/backend.go
+++ b/luks2/backend.go
@@ -32,11 +32,7 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-type probeDepthKeyType struct{}
-
 var (
-	probeDepthKey = probeDepthKeyType{}
-
 	devRoot              = "/dev"
 	filepathEvalSymlinks = filepath.EvalSymlinks
 	osStat               = os.Stat
@@ -62,53 +58,25 @@ func newStorageContainerBackend() *storageContainerBackend {
 
 // Probe implements [secboot.StorageContainerBackend.Probe].
 //
-// The path can be the path to the LUKS2 source device, or a symbolic link to it, or it can
-// be a path to an open DM volume that is backed by a LUKS2 storage device - this function
-// will walk the DM tables in this case to find the underlying source device. It identifies
-// a [StorageContainer] by its device number, and so it will always return the same instance
-// for any paths that reference the same container, regardless of what type of path is supplied.
+// The path can be the path to the LUKS2 source device, or a symbolic link to one.
+//
+// It identifies a [StorageContainer] by its device number, and it will always return
+// the same instance for any paths that reference the same container, regardless of
+// what path is supplied.
 func (b *storageContainerBackend) Probe(ctx context.Context, path string) (secboot.StorageContainer, error) {
 	// Use the luksview package to test if the supplied path is
 	// a device or file containing a LUKS2 container.
 	_, err := newLuksView(ctx, path)
-	if err != nil {
-		if errors.Is(err, internal_luks2.ErrInvalidMagic) {
-			// We expect this error if the supplied path genuinely doesn't reference
-			// a LUKS2 container. It's possible we were supplied a path to a DM
-			// device, so walk the tree of tables to obtain a source device.
-			sourcePath, err := sourceDeviceFromDMDevice(ctx, path)
-			switch {
-			case errors.Is(err, errNotDMBlockDevice):
-				// Not a DM block device, so ignore this.
-				return nil, nil
-			case errors.Is(err, errUnsupportedTargetType):
-				// Ignore DM block devices with unrecognized target types.
-				return nil, nil
-			case err != nil:
-				// Any other type of error is unexpected, so return this.
-				return nil, fmt.Errorf("cannot obtain source device for dm device %s: %w", path, err)
-			default:
-				// The supplied path is a DM device and we have a source path - try
-				// again to see if we can find a LUKS2 header on the new source path.
-				// Apply a depth limit here to avoid too much recursion. The permitted
-				// depth here should be sufficient for lvm (linear) inside crypt or
-				// the stack of DM devices associated with an in-progress reencrypt.
-				var depth uint
-				if d := ctx.Value(probeDepthKey); d != nil {
-					depth = d.(uint)
-					if depth > 10 {
-						return nil, errors.New("path to dm device that is too deeply nested")
-					}
-				}
-				ctx := context.WithValue(ctx, probeDepthKey, depth+1)
-				return b.Probe(ctx, sourcePath)
-			}
-		}
-
-		// The supplied device begins with what looks like a LUKS2 header,
-		// but the header failed to parse for some reason. Treat the supplied
-		// path as a path to a LUKS2 container and return an error in this
-		// case.
+	switch {
+	case errors.Is(err, internal_luks2.ErrInvalidMagic):
+		// We expect this error if the supplied path genuinely doesn't reference
+		// a LUKS2 container. Return nothing in this case.
+		return nil, nil
+	case err != nil:
+		// This error may indicate that the supplied device begins with what
+		// looks like a LUKS2 header, but the header failed to parse for some
+		// reason. Or it could be any other unexpected error. We'll return the
+		// error in this case.
 		return nil, err
 	}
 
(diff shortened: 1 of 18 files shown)


luks2/dm_helper.go
```diff
@@ -0,0 +1,122 @@
...
+		if os.IsNotExist(err) {
+			// This is not a DM device. Ignore it because another
+			// backend might handle it.
+			return "", errNotDMBlockDevice
+		}
+		return "", err
+	}
+
+	// The supplied path is a device mapper device. Obtain the table using dmsetup.
+	cmd := exec.CommandContext(ctx, "dmsetup", "table", path)
```
pedronis, line 75: I asked @valentindavid to check that is there in initramfs

chrisccoulson: It's not the end of the world if it isn't - I don't really like parsing the output of these tools anyway, and I had a longer term plan to implement a simple go library that implements the `ioctl` based interface for `/dev/mapper/control`. This is just a temporary workaround for the spike.

valentindavid: dmsetup is in. Udev needs to call it.

luks2/backend.go
```diff
@@ -0,0 +1,147 @@
...
+// for any paths that reference the same container, regardless of what type of path is supplied.
+func (b *storageContainerBackend) Probe(ctx context.Context, path string) (secboot.StorageContainer, error) {
+	// Use the luksview package to test if the supplied path is
+	// a device or file containing a LUKS2 container.
+	_, err := newLuksView(ctx, path)
+	if err != nil {
+		if errors.Is(err, internal_luks2.ErrInvalidMagic) {
+			// We expect this error if the supplied path genuinely doesn't reference
+			// a LUKS2 container. It's possible we were supplied a path to a DM
+			// device, so walk the tree of tables to obtain a source device.
```
valentindavid, line 78: Not really fan of letter the caller being confused on what path it gives and still trying to make sense of it.

What if you have a luks2 in a luks2? That does the path mean then?

luks2/container.go
```diff
@@ -0,0 +1,225 @@
...
+	return c.path
+}
+
+// BackendName implements [secboot.StorageContainer.BackendName]
+func (c *storageContainerImpl) BackendName() string {
+	return storageContainerBackendName
+}
+
+// Activate implements [secboot.StorageContainer.Activate]
+func (c *storageContainerImpl) Activate(ctx context.Context, ki secboot.KeyslotInfo, key []byte, opts ...secboot.ActivateOption) error {
```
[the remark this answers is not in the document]
pedronis: this one takes a concrete key, see the impl. That concerns is for the level above of the API and we sketched a solution I think by being able to pass keydata in?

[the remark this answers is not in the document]
valentindavid: Is the key parameter the unsealed key?

luks2/backend.go
pedronis: it wouldn't be too unusual too have a New* method with more strict prereqs on the arguments, and a different function that is more DWIM

chrisccoulson: Actually, I'm wondering if `NewStorageContainer` is an appropriate name anyway, given that it's doc note contradicts that of the corresponding `StorageContainerBackend.Probe`, which says it should return the same `StorageContainer` implementation regardless of what path it's called with - it's only "new" on the first call to it. I'm going to split `Probe` into `Probe` (which assumes the path is to a source storage container) and `ProbeActivated` (which assumes the path is to a storage container that is backed by one that can be returned by `Probe` with the corresponding source path), but I think `NewStorageContainer` should probably be renamed to something else anyway.

chrisccoulson: Ok, I've split this into `Probe` and `ProbeActivated`, and the existing `NewStorageContainer` is `FindStorageContainer` and `FindActivatedStorageContainer`.

luks2/dm_helper.go
```diff
@@ -0,0 +1,122 @@
...
+	// XXX: Is the device always identified by device number? Can it be referenced
+	// by its device node path (in which case, this code needs some readjustment)?
+	// Resolve the sysfs path using the device number.
+	resolvedDev, err := filepathEvalSymlinks(filepath.Join(sysfsRoot, "dev/block", dev))
+	if err != nil {
+		return "", fmt.Errorf("cannot resolve symlink for block device %s: %w", dev, err)
+	}
+
+	// Return the device node path, which uses the basename of the sysfs dir.
+	return filepath.Join("/dev", filepath.Base(resolvedDev)), nil
```
valentindavid, line 121: I can see at least 2 block devices in the kernel source that do not have the device node right under `/dev`: ATA over Ethernet and Packet writing on CD/DVD media.

Could we read the `uevent` attribute instead and extract `DEVNAME`?

chrisccoulson: Yes, I can do that.

luks2/reader.go
```diff
@@ -0,0 +1,192 @@
...
+		// once all of this functionality is implemented natively in this package.
+		view, err := newLuksView(ctx, s.container.Path())
+		if err != nil {
+			return nil, fmt.Errorf("cannot obtain new luksview.View: %w", err)
+		}
+		token, _, inUse := view.TokenByName(name)
+		if !inUse {
+			return nil, fmt.Errorf("no metadata for keyslot %q", name)
+		}
+		info.keyslotId = token.Keyslots()[0] // luksview guarantees there is always 1 keyslot here.
```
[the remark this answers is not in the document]
chrisccoulson: I've removed the contradictory comment now.

luks2/reader.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func (s *storageContainerReadWriterImpl) ReadKeyslot(ctx context.Context, name string) (secboot.KeyslotInfo, error) {
+	if err := s.ensureKeyslotNames(); err != nil {
+		return nil, err
+	}
+
+	info, exists := s.keyslots[name]
+	if !exists {
+		return nil, secboot.ErrKeyslotNotFound
```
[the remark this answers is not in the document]
chrisccoulson: I've added a test for this now.

luks2/backend.go
```diff
@@ -0,0 +1,160 @@
...
+		b.containers[st.Rdev] = container
+	}
+
+	return container, nil
+}
+
+// ProbeActivated implements [secboot.StorageContainerBackend.ProbeActivated].
+//
+// The path can be the path to a DM device that is backed by a LUKS2 container, or
+// a symbolic link to one.
```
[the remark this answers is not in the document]
chrisccoulson: I've updated this now.

luks2/reader.go
```diff
@@ -64,13 +77,63 @@ func (s *storageContainerReadWriterImpl) ensureKeyslotNames() error {
...
+		}
+		info.keyslotId = token.Keyslots()[0] // luksview guarantees there is always 1 keyslot here.
+	case secboot.KeyslotTypePlatform:
+		r, err := luks2Ops.NewKeyDataReader(s.container.Path(), name)
+		if err != nil {
+			return fmt.Errorf("cannot obtain reader for %q: %w", name, err)
+		}
+		info.keyslotId = r.KeyslotID()
+		info.keyslotPriority = r.Priority()
+		info.keyslotData = r
```
pedronis, line 129: how does this will notice updates if it is cached if we don't have yet a write side?

pedronis (commented):
question

luks2/container.go
```diff
@@ -0,0 +1,233 @@
...
+	// and in order to obtain the volume name, we need to iterate over
+	// the existing device mapper volumes to find any that are backed by
+	// this storage container.
+	dmDevices, err := filepath.Glob(filepath.Join(devRoot, "dm-*"))
+	if err != nil {
+		return "", fmt.Errorf("cannot build list of active dm devices: %w", err)
+	}
+
+	// Iterate over the list of active DM volumes.
+	for _, path := range dmDevices {
```
valentindavid, line 113: Maybe we could make it look at only the holders of the device?

luks2/dm_helper.go
```diff
@@ -0,0 +1,152 @@
...
+	switch targetType {
+	case "crypt":
+		// For crypt device, we expect the following extra args:
+		// <cipher> <key> <iv_offset> <device> <offset> <features>...
+		if len(tbl) < 8 {
+			return "", fmt.Errorf("invalid dm table: unexpected number of arguments (%d) for device with crypt target: expected at least 8", len(tbl))
+		}
+
+		dev = tbl[6]
+	case "linear":
```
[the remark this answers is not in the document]
chrisccoulson: We aren't planning to do anything with this - I only added it because the existing passphrase based FDE uses lvm inside luks2 which this code would support. It's there largely because it was the easiest target to add, and it makes it possible to write unit tests that demonstrate the nesting. What we will want here eventually is a branch for `integrity`.

luks2/keyslot_info.go
```diff
@@ -0,0 +1,60 @@
...
+func (i *keyslotInfoImpl) Data() secboot.KeyDataReader {
+	return i.keyslotData
+}
+
+func (i *keyslotInfoImpl) KeyslotID() int {
+	return i.keyslotId
+}
+
+// KeyslotInfo provides information about a LUKS2 keyslot.
+type KeyslotInfo interface {
```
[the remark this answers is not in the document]
chrisccoulson: Maybe that would be ok

luks2/container.go
chrisccoulson: I wonder if we could do that and skip the call to `sourceDeviceFromDMDevice`. For that to work, we would need to ensure that there is only ever a single DM device holding the top-level block device. I checked on my classic install which uses the stacked LVM on LUKS2 setup and the nested DM device for the LVM volume isn't a holder of the top-level block device, eg:
```
$ ls -l /sys/dev/block/259\:3/holders/
total 0
lrwxrwxrwx 1 root root 0 Sep  4 16:33 dm-0 -> ../../../../../../../../../../virtual/block/dm-0
$ ls -l /sys/dev/block/259\:3/holders/dm-0/holders/
total 0
lrwxrwxrwx 1 root root 0 Sep  4 16:34 dm-1 -> ../../dm-1
$ ls -l /sys/dev/block/259\:3/holders/dm-0/holders/dm-1/holders/
total 0
```

I wonder if we can always rely on this to be true?

pedronis (approved):
thanks

luks2/container.go
valentindavid: We cannot totally rely on holders. Some device mapper types have multiple devices (verity, integrity, etc.), so we need to figure out which is which. But I think we can rely on holders to be there to trim down what we look at, instead of scanning for all devices.
Could you add a TODO or something here to mention that there might be better ways? And we can maybe optimize things in a later PR.

chrisccoulson: Sure, that sounds good to me.

PR #426, opened 2025-06-17
Add public API CheckPassphraseEntropy() that calculates entropy

kukrimate:
Returns the estimated symbol pool size, number of symbol (taking into account runs of identical, and nearby chars), and the actual base2 entropy.

### DIFF
--- a/entropy.go
+++ b/entropy.go
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2025 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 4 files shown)


entropy.go
```diff
@@ -0,0 +1,37 @@
...
+
+type PassphraseEntropyStats struct {
+	SymbolPoolSize  uint32
+	NumberOfSymbols uint32
+	EntropyBits     uint32
+}
+
+// CheckPassphraseEntropy calculates entropy for PINs and passphrases (PINs will be supplied as a numeric passphrase).
+func CheckPassphraseEntropy(passphrase string) (*PassphraseEntropyStats, error) {
+	stats := &PassphraseEntropyStats{}
```
chrisccoulson, line 32: It doesn't matter here and tbh I do this inconsistently as well so this is more of a note than a request for any change, but you can just do `new(PassphraseEntropyStats)` here. You can even just declare it as a stack variable - `var stats PassphraseEntropyStats` - and the toolchain recognizes that it escapes the current context and stack allocates it automatically.

chrisccoulson (approved):
Thanks, this LGTM!

PR #428, opened 2025-06-18
storage: Rename ActivateOption and ActivateOptionVisitor

chrisccoulson:
The new activation API will also have a way of specifying options, and I
want a way to distiguish between options that are supplied to the
activate API and options that are supplied to StorageContainer.Activate.
In preparation, this renames ActivateOption and ActivateOptionVisitor to
ContainerActivateOption and ContainerActivateConfig.


chrisccoulson:
I'm going to reject this one - it will be part of the main PR that introduces the initial version of the new API.

PR #430, opened 2025-06-23
AuthRequestor: Merge the 2 methods into 1

chrisccoulson:
The current interface has 2 methods - one to request a passphrase and
one to request a recovery key. In the future, we will also have PINs and
we want the ability for the user to supply any requested user credential
rather than having to wait to exhaust all of their passphrase/PIN
requests before they are able to enter a recovery key.

The new method can be asked to request any combination of passphrase,
PIN or recovery key using a flags argument.

One thing I'm not sure about yet is how best to make it possible to
localize messages for every combination of credential that can be
requested. My original idea was to use the text/template package which
gives the application a lot more flexiblity, but this does not work
because it brings in too many dependencies.

Note that the argument names are different - sourceDevicePath is now
just "path" because the encrypted container may not always be a device.
The volumeName argument is just "name" and there will be an option for
the new activation API to customize this to be any human readable name.
The idea of "volumeName" only makes sense for backends that make use of
the device mapper to create new volumes, which of course, LUKS2 does.


auth_requestor_systemd_test.go
```diff
@@ -55,219 +57,180 @@ func (s *authRequestorSystemdSuite) setPassphrase(c *C, passphrase string) {
...
 
-	requestor := NewSystemdAuthRequestor(data.tmpl, "")
+	requestor := NewSystemdAuthRequestor(map[UserAuthType]string{
+		UserAuthTypePassphrase:                                             "Enter passphrase for %[1]s (%[2]s):",
+		UserAuthTypePIN:                                                    "Enter PIN for %[1]s (%[2]s):",
+		UserAuthTypeRecoveryKey:                                            "Enter recovery key for %[1]s (%[2]s):",
+		UserAuthTypePassphrase | UserAuthTypePIN:                           "Enter passphrase or PIN for %[1]s (%[2]s):",
+		UserAuthTypePassphrase | UserAuthTypeRecoveryKey:                   "Enter passphrase or recovery key for %[1]s (%[2]s):",
+		UserAuthTypePIN | UserAuthTypeRecoveryKey:                          "Enter PIN or recovery key for %[1]s (%[2]s):",
+		UserAuthTypePassphrase | UserAuthTypePIN | UserAuthTypeRecoveryKey: "Enter passphrase, PIN or recovery key for %[1]s (%[2]s):",
```
valentindavid, line 81: So we have to provide 2^N-1 strings. If we two different kind of keys, that will become 31 strings? I do not know if there are any clever way to handle it, but I feel it would be nice if we could capture at compile time if we are missing some cases.

valentindavid (approved):

pedronis (commented):
question about the map of strings

PR #431, opened 2025-06-26
[WIP] Add initial stateful activation API

chrisccoulson:
This initial implementation doesn't yet retain state, perform the
primary key cross check or support passphrases or PINs.


activate.go
```diff
@@ -0,0 +1,665 @@
...
+	// inherits the global configuration).
+	cfg activateConfig
+
+	stderr io.Writer // For writing error messages to.
+
+	next activateOneContainerStateMachineTask // The next state to run.
+
+	err            error                            // The first fatal error for this statemachine
+	status         ActivationStatus                 // Whether and how this container is activated.
+	keyslotRecords map[string]*keyslotAttemptRecord // Keyslot specific status.
```
pedronis: the comment probably should explain what the key is? the keyslot name?

activate.go
```diff
@@ -0,0 +1,665 @@
...
+			slot: slot,
+		}
+
+		switch slot.Type() {
+		case KeyslotTypePlatform:
+			kd, err := ReadKeyData(slot.Data())
+			if err != nil {
+				rec.err = &InvalidKeyDataError{fmt.Errorf("cannot decode keyslot metadata: %w", err)}
+				fmt.Fprintf(m.stderr, "Error with keyslot %q: %v\n", name, rec.err)
+			}
```
pedronis: is kd nil here?

activate.go
```diff
@@ -0,0 +1,665 @@
...
+		if !record.usable() {
+			// Skipping this unusable keyslot.
+			continue
+		}
+		if record.slot.Type() != KeyslotTypePlatform {
+			continue
+		}
+		if record.data.AuthMode() != AuthModeNone {
+			// Skip this one
+			continue
```
pedronis: this will be tested later when we have passphrases?

activate.go
```diff
@@ -0,0 +1,665 @@
...
+		return nil
+	}
+
+	// We have failed to unlock
+	m.next = activateOneContainerStateMachineTask{}
+
+	return ErrCannotActivate
+}
+
+func (m *activateOneContainerStateMachine) tryRecoveryKeyslots(ctx context.Context, slotRecords keyslotAttemptRecordSlice, recoveryKey RecoveryKey) Keyslot {
```
pedronis: this is a helper and not one of the states, not sure if we should mark it somehow but it got me slightly confused

activate_options.go
```diff
@@ -0,0 +1,261 @@
...
+
+// ActivateOption represents an option that can be supplied to
+// [ActivateContext.ActivateContainer].
+type ActivateOption interface {
+	// ApplyOptionToConfig is called to make changes to the supplied ActivateConfig.
+	ApplyOptionToConfig(ActivateConfig)
+}
+
+// ActivateContextOption represents an option that can be supplied to
+// [NewActivateContext] or [ActivateContext.ActivateContainer].
```
pedronis: will we have global only options?

activate.go
chrisccoulson: I can add a comment for this.

activate.go
```diff
@@ -0,0 +1,665 @@
...
+	"golang.org/x/sys/unix"
+)
+
+// ErrCannotActivate is returned from ActivateContext.ActivateContainer
+// if a storage container cannot be activated because there are no valid
+// keyslots and / or there are no more passphrase, PIN or recovery key
+// attempts remaining.
+var ErrCannotActivate = errors.New("cannot activate: no valid keyslots and / or no more passphrase, PIN or recovery key tries remaining")
+
+type errorKeyslot struct {
```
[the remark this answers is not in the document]
chrisccoulson: I'll add one to this.

activate_external_keys.go
```diff
@@ -0,0 +1,83 @@
...
+// [ReadKeyData] and will eventually be deleted along with [KeyDataReader]
+// when the legacy activation API is deleted.
+func (d *ExternalKeyData) ReadableName() string {
+	return d.name
+}
+
+// Read implements [io.Reader].
+func (d *ExternalKeyData) Read(data []byte) (int, error) {
+	return d.data.Read(data)
+}
```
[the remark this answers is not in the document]
chrisccoulson: Ah, I can see why this doesn't get tested now.

activate.go
chrisccoulson: It is.

activate.go
```diff
@@ -0,0 +1,665 @@
...
+			return fmt.Errorf("cannot request user credential: %w", err)
+		}
+
+		// We have a user credential.
+		// 1) TODO: Try it against every keyslot with a passphrase.
+		// 2) TODO: See if it decodes as a PIN and try it against every keyslot with a passphrase.
+		// 3) See if it decodes as a recovery key, and try it against every recovery keyslot.
+
+		recoveryKey, err := ParseRecoveryKey(cred)
+		if err == nil {
```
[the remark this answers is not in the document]
chrisccoulson: We don't - the supplied credential might be a passphrase or a PIN, so we want to handle the case where it's not a valid recovery key. This will probably make more sense when passphrase and PIN support is added.

activate_external_keys.go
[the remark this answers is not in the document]
chrisccoulson: There are tests for both of these now.

pedronis (commented):
thanks for the changes, one question still

PR #432, opened 2025-06-27
Perform primary key cross-check with the new activation API

chrisccoulson:
All unlocked storage containers must be part of the same install to
prevent an attack where an adversary replaces a container with one that
contains their own credentials, in order to gain access to confidential
data on another storage container. This PR implements checks to enforce
this, using 1 of 2 mechanisms:

1. As the unlock key for each keyslot is derived from a primary key that
   is common for all keyslots, and a unique key, there is a
   cryptographic binding between the unlock key and the primary key. If
   a keyslot is used successfully, then verifying that the primary key
   matches the primary key from keyslots used to unlock other containers
   is sufficient to determine that the storage containers are related to
   the same install. Where a previous storage container has been
   unlocked using a platform keyslot, any keyslots that do not have a
   matching primary key will be dismissed as being unsuitable for
   unlocking its associated storage container.
2. If the first container is unlocked using a recovery key, then there
   is no primary key to which keyslots from subsequent containers can be
   compared. In this case, subsequent containers must either be unlocked
   using a recovery key, or must be unlocked using a platform keyslot
   protected by a platform that is registered with the new
   `PlatformProtectedByStorageContainer` flag. The existing "plainkey"
   platform now registers itself with this flag. These keyslots are
   protected using a key that must be stored inside the first unlocked
   storage container. If unlocking succeeds using a keyslot protected by
   a platform registered with this flag, then this is sufficient to
   determine that the storage container is part of the same install as
   the first container. The recovered primary key can be used to check
   the binding of subsequent storage containers using the first method.

There are some additional restrictions:
- If more than one container is unlocked using a recovery key, then all
  containers must be unlocked using a recovery key. This is because it
  is only possible to prove the binding of storage containers to a
  single container using a platform that is registered with the
  `PlatformProtectedByStorageContainer` flag.
- Once any storage containers have been unlocked using a platform
  keyslot, all subsequent containers must be unlocked using platform
  keyslots. This is because it is not possible to properly verify that
  subsequent containers are bound to existing containers if they are
  unlocked with a recovery key. In this case, existing containers must
  be deactivated and then activated only with recovery keys. A future PR
  will provide a way to handle this case properly.

### DIFF
--- a/activate.go
+++ b/activate.go
@@ -107,9 +107,20 @@ func (s keyslotAttemptRecordSlice) Swap(i, j int) {
 type activateOneContainerStateMachineFlags int
 
 const (
-	activatePermitPlatformKey activateOneContainerStateMachineFlags = 1 << iota
-	activatePermitRecoveryKey
+	// activatePermitRecoveryKey allows recovery keyslots to be used.
+	// Note that platform keyslots are always permitted to be used.
+	activatePermitRecoveryKey activateOneContainerStateMachineFlags = 1 << iota
+
+	// activateRequrePlatformProtectedByStorageContainer is used to
+	// require that platform keyslots are protected by platforms registered
+	// with the PlatformProtectedByStorageContainer flag in order to
+	// be used.
 	activateRequirePlatformKeyProtectedByStorageContainer
+
+	// activateCrossCheckPrimaryKey is used to require that the
+	// primary key recovered from a platform keyslot is cross-checked
+	// against a previously used primary key before it can be used
+	// for unlocking.
 	activateCrossCheckPrimaryKey
 )
 
@@ -228,16 +239,9 @@ func (m *activateOneContainerStateMachine) initExternalKeyAttempts(ctx context.C
 }
 
 func (m *activateOneContainerStateMachine) initKeyslotAttempts(ctx context.Context) error {
-	if m.flags&activatePermitPlatformKey > 0 {
-		m.next = activateOneContainerStateMachineTask{
-			name: "try-no-user-auth-keyslots",
-			fn:   m.tryNoUserAuthKeyslots,
-		}
-	} else {
-		m.next = activateOneContainerStateMachineTask{
-			name: "try-with-user-auth-keyslots",
-			fn:   m.tryWithUserAuthKeyslots,
-		}
+	m.next = activateOneContainerStateMachineTask{
+		name: "try-no-user-auth-keyslots",
+		fn:   m.tryNoUserAuthKeyslots,
 	}
 
 	r, err := m.container.OpenRead(ctx)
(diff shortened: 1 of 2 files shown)


activate.go
```diff
@@ -615,65 +703,236 @@ func (m *activateOneContainerStateMachine) runNextState(ctx context.Context) err
...
+		// protected with a platform registered with the
+		// PlatformProtectedByStorageContainer flag.
+		flags = activatePermitPlatformKey | activatePermitRecoveryKey | activateRequirePlatformKeyProtectedByStorageContainer
+	case c.state.NumActivatedContainersWithRecoveryKey() > 1:
+		// If more than one storage container has been unlocked with a
+		// recovery key, then only permit recovery keyslots. We can only
+		// associate keyslots protected by a platform registered with the
+		// PlatformProtectedByStorageContainer flag with a single container,
+		// meaning that 1 or more storage containers might be supplied by
+		// an adversary.
```
pedronis: is this true in general? can't the first container contain multiple plainkeys?

pedronis (commented):
small comment and a question, this needs a review also from @valentindavid I think

activate.go
```diff
@@ -615,65 +703,236 @@ func (m *activateOneContainerStateMachine) runNextState(ctx context.Context) err
...
+		flags = activatePermitPlatformKey | activatePermitRecoveryKey
+	case c.state.TotalActivatedContainers() == 1 && c.state.NumActivatedContainersWithRecoveryKey() == 1:
+		// If this is the second container to be activated and the first
+		// was activated with a recovery keyslot, permit either a platform
+		// or recovery keyslot to be used. The platform keyslot must be
+		// protected with a platform registered with the
+		// PlatformProtectedByStorageContainer flag.
+		flags = activatePermitPlatformKey | activatePermitRecoveryKey | activateRequirePlatformKeyProtectedByStorageContainer
+	case c.state.NumActivatedContainersWithRecoveryKey() > 1:
+		// If more than one storage container has been unlocked with a
```
valentindavid: What if we had 3 containers. 2 unlocked with recovery keys, but they both have matching platform keys. Which means we unlock the 3rd one with plainkey. In this case the primary key of the 3rd one should match the 2 first containers, right?

Should it be that we remember which containers where we found platform keys. And once we unlock a disk with a plainkey, we consider any recovery unlocked containers containing the platform key to have a matching primary key as the plainkey unlocked one.

I know it is not a scenario we have for Ubuntu Core. But it makes more sense. Or at least if this is correct, I understood the intention here.

activate.go
chrisccoulson: It can, but that's not the issue here. We can make use of the plainkey platform to associate 2 storage containers in the case where the source of the plainkey is unlocked using a recovery key, but the plainkey keyslots can only be associated with 1 source container.

Thinking about this again, if the user sat in front of the device knows the recovery key for 2 storage containers, then it's probably ok for me to remove this restriction.

PR #435, opened 2025-06-29
[WIP] Save activation status using the new activation API

chrisccoulson:


activate_state.go
```diff
@@ -35,22 +41,78 @@ const (
...
+}
+
 // ContainerActivateState contains the activation state for a single
 // [StorageContainer].
 type ContainerActivateState struct {
-	Status ActivationStatus `json:"status"`
+	Status           ActivationStatus            `json:"status"`                      // The overall activation status for this storage container.
+	Keyslot          string                      `json:"keyslot,omitempty"`           // If the container was activated, the name of the keyslot used.
+	DeactivateReason DeactivationReason          `json:"deactivate-reason,omitempty"` // An argument supplied to ActivateContext.DeactivateContainer.
+	KeyslotErrors    map[string]KeyslotErrorType `json:"keyslot-errors"`              // A map of errors for tried keyslots, keyed by name.
```
valentindavid: Note sure, but maybe that would be interesting to have it ordered. Because if we ever have to report details to users for example, in some advanced reporting mode, then it might be confusing if we cannot show the errors in order of them happening.

activate.go
```diff
@@ -675,6 +698,38 @@ func (m *activateOneContainerStateMachine) primaryKeyInfo() (PrimaryKey, keyring
 	return m.primaryKey, m.primaryKeyID, nil
 }
 
+func (m *activateOneContainerStateMachine) activationState() (*ContainerActivateState, error) {
+	if m.hasMoreWork() {
+		return nil, errors.New("state machine has not finished")
```
pedronis: this little bit is not exercised

activate_state.go
pedronis: maybe we should record a list of order we tried the keyslots then?

keydata.go
```diff
@@ -506,17 +543,22 @@ func (d *KeyData) updatePassphrase(payload, oldAuthKey []byte, passphrase string
...
 	if d.data.PassphraseParams.Encryption != passphraseEncryption {
 		// Only AES-CFB is supported
-		return fmt.Errorf("unexpected encryption algorithm \"%s\"", d.data.PassphraseParams.Encryption)
+		return keyDataError{fmt.Errorf("unexpected encryption algorithm \"%s\"", d.data.PassphraseParams.Encryption)}
 	}
 
 	handle, err := handlerInfo.handler.ChangeAuthKey(d.platformKeyData(), oldAuthKey, authKey, platformContext)
 	if err != nil {
-		return err
+		return processPlatformHandlerError(err)
```
pedronis: can you explain this change?

activate_state.go
chrisccoulson: I've added a list of keyslot names ordered by when they failed.

activate.go
```diff
@@ -77,7 +80,27 @@ func (r *keyslotAttemptRecord) usable() bool {
...
+
+	// In general, a keyslot that has encountered an error becomes unusable,
+	// with one exception being that if the error is a result of the user
+	// supplying an incorrect credential (passphrase, PIN, or recovery key),
+	// it should remain usable.
+	var expectedUserAuthErr error
+	switch {
+	case r.slot.Type() == KeyslotTypeRecovery:
+		// Recovery keyslot
+		expectedUserAuthErr = errInvalidRecoveryKey
```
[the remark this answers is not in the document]
chrisccoulson: I've commented this branch out for now with a note, as it's currently unreachable. The return is unreachable too, but gets tested with the PR that adds passphrase support.

activate.go
chrisccoulson: There's a similar path in another function that isn't tested as well. I've added a couple of tests for both of them now.

pedronis (approved):
thank you

PR #439, opened 2025-07-09
preinstall: Fix PR412 on go1.18

chrisccoulson:
I landed PR412 without noticing that the tests are failing on go 1.18,
but passing with a newer go. It looks like it's because the test of each
error value returned from `checkHostSecurity` is being performed on the
wrong error value - it works with newer go versions because it has
native support for `[]error` with `errors.Is`.


pedronis (approved):
thanks

PR #453, opened 2025-07-23
preinstall: Add PPI actions to enable and clear the TPM

chrisccoulson:
This adds 3 PPI actions for resolving various issues with the TPM:
- `ActionEnableTPMViaFirmware`, to enable the TPM.
- `ActionEnableAndClearTPMViaFirmware`, to enable (or ensure enabled) and
  clear the TPM.
- `ActionClearTPMViaFirmware`, to clear the TPM.

Once a PPI action has been submitted, the system needs to transition back
to the firmware for it to be handled. The `RunChecksContext.Run` API will
return an error kind that is either `ErrorKindReboot` or
`ErrorKindShutdown` to indicate how to transition back to firmware (known
as the state transition action). I've only ever seen "reboot" for this,
but the PPI can specify it to be "reboot", "shutdown" and some other
values that we ignore.

Once a PPI action has been submitted, `RunChecksContext.Run` will
return the same error kind on subsequent calls, to tell the caller to
reboot or shutdown. This is because the PPI operation may resolve more
than one error. It doesn't work like this right now because go-tpm2's
`ppi.PPI` interface does not provide a way to do it (see
https://github.com/canonical/secboot/issues/451.

The PPI indicates which operations are supported, and whether those
operations require physical presence or not. If an operation is not
supported, then `RunChecksContext.Run` will not return the corresponding
Action. It is common for firmware implementations to only support one
of `ActionEnableAndClearTPMViaFirmware` and `ActionClearTPMViaFirmware`.


efi/preinstall/actions_ppi.go
```diff
@@ -0,0 +1,183 @@
...
+// the operation has a corresponding Action, then that action is returned along with
+// the state transition action, which tells the caller how to transition back to the
+// platform firmware in order to process the submitted operation.
+// If there is no pending PPI operation, or it doesn't have a corresponding action,
+// then ActionNone is returned.
+//
+// XXX(chrisccoulson): This currently only returns ActionNone because ppi.PPI does
+// not provide a way to determine if there is a pending PPI operation. It will
+// require an update to go-tpm2.
+func pendingPPIAction(env internal_efi.HostEnvironment) (Action, ppi.StateTransitionAction, error) {
```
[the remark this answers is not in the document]
chrisccoulson: I've removed this for now. As I wrote this, I can't remember what my original intention was with this function, but once I remember then I'll open an issue so fully implementing it doesn't fall off my radar.

efi/preinstall/actions_ppi.go
```diff
@@ -0,0 +1,133 @@
...
+	case ppi.StateTransitionRebootRequired:
+		// ok
+	default:
+		return 0, fmt.Errorf("unsupported state transition action %q", sta)
+	}
+	return sta, nil
+}
+
+// isPPIActionAvailable checks whether the PPI operation associated with the supplied
+// action is avaiable. An action is considered to be available regardless of whether
```
ernestl, line 89: avaiable -> available

efi/preinstall/actions_ppi.go
```diff
@@ -0,0 +1,133 @@
...
+// it ins't implemented, it is a firmware only operation, or it is currently blocked
+// because of the firmware settings.
+func isPPIActionAvailable(env internal_efi.HostEnvironment, action Action) (bool, error) {
+	dev, err := env.TPMDevice()
+	if err != nil {
+		return false, err
+	}
+	p, err := obtainTPMDevicePPI(dev)
+	if err != nil {
+		return false, fmt.Errorf("cannot obtain physical presence interface: %w", err)
```
ernestl, line 100: nitpick: when do we refer to it as "physical presence interface" and when as PPI?

efi/preinstall/checks_context_test.go
```diff
@@ -95,6 +111,23 @@ func (s *runChecksContextSuite) testRun(c *C, params *testRunChecksContextRunPar
...
 			s.resetTPMAndReplayLog(c, log, log.Algorithms...)
 		}
+
+		p := params.ppi
+		if p == nil {
+			// Create a mockPPI if one wasn't supplied
+			p = &mockPPI{}
+		}
+		if p.ops == nil {
+			// Initialize an uninitialzed PPI. Supply the
```
ernestl, line 121: uninitialzed -> uninitialized

efi/preinstall/actions.go
ernestl, line 1: When does this come into play "Add extra actions that can be performed by this package by passing the action to [RunChecksContext.Run]". The default will still work fine, just checking. IsExternalAction not covered with tests.

efi/preinstall/checks_context.go
ernestl, line 1: general comment not specifically applicable to this PR: shows significant amount to missing test coverage

ernestl (approved):
Looks good, few nitpicks

efi/preinstall/actions_ppi.go
chrisccoulson: Thanks, I've fixed this.

efi/preinstall/checks_context_test.go
chrisccoulson: This is fixed now.

PR #454, opened 2025-07-23
preinstall: Add ActionProceed action

chrisccoulson:
There are some error kinds that don't necessarily have to be resolved
and don't have to prevent FDE from being enabled, but can be used to
provide information to a user so that they can evaluate them and
determine whether they would like to proceed at their own risk. Eg, if
the system is executing third party code before the OS and the user
cannot or does not want to disable it, then there is an increased risk
of PCR fragility as a result of those third party components being
updated by something other than the OS. If an error indicates that
pre-boot DMA remapping is disabled and the user cannot enable it, then
there is an increased risk of a locally present adversary obtaining
secrets or altering code execution via external ports.

This introduces a new action - `ActionProceed`, which can be executed only
if all of the previously returned error kinds permit this action. When
used, a CheckFlags flag will be added for each unresolved error kind,
which has the effect of ignoring those errors and making them warnings
instead.

### DIFF
--- a/efi/preinstall/checks_context.go
+++ b/efi/preinstall/checks_context.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/canonical/go-tpm2"
+	"github.com/canonical/go-tpm2/ppi"
 	secboot_efi "github.com/snapcore/secboot/efi"
 	internal_efi "github.com/snapcore/secboot/internal/efi"
 )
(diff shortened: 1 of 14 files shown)


efi/preinstall/checks_context.go
```diff
@@ -163,7 +168,15 @@ type RunChecksContext struct {
...
-	expectedActions  []Action
+
+	// expectedActions contains a slice of actions that are expected on a subsequent call
+	// to Run. Trying to execute an action that is not in here will result in an error
+	// being returned.
+	expectedActions []Action
+
+	// proceedFlags indicates the CheckFlags that will be enabled if Run is called
+	// with ActionProceed.
+	proceedFlags CheckFlags
```
pedronis, line 179: is plan to remember to always pass ActionProceed or will there be a way to extract/record these? in theory asking to proceed at a later point might be different than remember the flags for the errors the user reviewed

pedronis (commented):
question

efi/preinstall/checks_context.go
chrisccoulson: I intend to add this information to a new field in `CheckResult` in a follow-up PR, so that it will be possible to remember which errors the user reviewed and accepted in the past.

efi/preinstall/checks_context.go
```diff
@@ -619,11 +649,33 @@ func (c *RunChecksContext) Run(ctx context.Context, action Action, args ...any)
 						fmt.Errorf("cannot filter unavailable actions: %w", err),
 					)
 				}
+				if _, canProceed := errorKindToProceedFlag[kind]; !canProceed {
```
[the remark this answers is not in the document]
ernestl: Ok thanks I think this will do the job for now.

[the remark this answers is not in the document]
chrisccoulson: I've implemented this now.

efi/preinstall/checks_context.go
```diff
@@ -548,7 +549,58 @@ func (c *RunChecksContext) runAction(action Action, args ...any) error {
 
 		return NewWithKindAndActionsError(kind, nil, errorKindToActions[kind], err)
```
pedronis, line 550: @ernestl do we have a problem that in you design args for actions are a map (like for errors) but the code actually takes []any ?

pedronis (commented):
comment

PR #458, opened 2025-09-17
efi/preinstall,tpm2: Clean up handling of DA lockouts 

chrisccoulson:
This makes various changes to the checks of the TPM's dictionary attack
protection and the availability of the TPM's lockout hierarchy:
- Both checks now run in all modes, rather than only during pre-install.
- The ordering has changed - the check for the availability of the
  lockout hierarchy executes first now. This check has the side-effect
  of clearing a DA lockout if it can execute successfully (ie, the
  lockout hierarchy is available and has an empty authorization value),
  suppressing any `ErrTPMLockout` error.
- There is now a `ErrTPMLockoutAvailabilityNotChecked` error that is
  returned as a warning in the `Warnings` field of `CheckResult` if the
  check for the availability of the lockout hierarchy cannot run because
  the lockout hierarchy has a non-empty authorization value.
- `tpm2.Connection.EnsureProvisioned` now resets the TPM's dictionary
  attack protection using `TPM2_DictionaryAttackLockReset` as long as it
  is able to use the lockout hierarchy.
- As a DA lockout should be cleared during installation (eg, by calling
  `tpm2.Connection.EnsureProvisioned`) and as part of a normal boot
  sequence, `ErrTPMLockout` is now only ever returned as a warning as part
  of the `Warnings` field of `CheckResult` as it shouldn't block an install
  from proceeding. As a consequence, the associated `ErrorKind`
  (`ErrorKindTPMDeviceLockout`) and argument struct (`TPMDeviceLockoutArgs`)
  are no longer needed. If it is not possible to clear a DA lockout
  during install or after a normal boot because the lockout hierarchy is
  not available, a `ErrTPMLockoutLockedOut` error will be returned from
  the checks if the test can run.

This kind of fixes https://github.com/canonical/secboot/issues/433 but not as the original description is written -
see the second comment.

In the future, I will adjust the public API to support snapd supplying
the authorization for the lockout hierarchy, so that the lockout
hierarchy availability check can execute even before a factory reset.

### DIFF
--- a/efi/preinstall/check_tpm.go
+++ b/efi/preinstall/check_tpm.go
@@ -182,14 +182,61 @@ func openAndCheckTPM2Device(env internal_efi.HostEnvironment, flags checkTPM2Dev
 		}
 	}
 
-	if flags&checkTPM2DevicePostInstall == 0 {
-		var errs []error
+	// Some errors from this point may be collected and returned together.
+	var errs []error
 
-		// Perform some checks only during pre-install.
-		perm, err := tpm.GetCapabilityTPMProperty(tpm2.PropertyPermanent)
-		if err != nil {
-			return nil, fmt.Errorf("cannot obtain value for TPM_PT_PERMANENT: %w", err)
+	// Obtain the permanent attributes from the TPM.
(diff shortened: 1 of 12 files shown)


pedronis (approved):
thanks, one question

PR #461, opened 2025-09-23
efi/preinstall: Fix a panic that can happen when secure boot is disabled

chrisccoulson:
If secure boot is disabled and the `SbatLevel` UEFI variable didn't exist
when shim executed, there will be no `EV_EFI_VARIABLE_AUTHORITY` events in
PCR7. At least one of these is required by
`checkSecureBootPolicyPCRForDegradedSettings` in order for the function to
return rather than the loop terminating and then hitting a panic at the
end of the function.

### DIFF
--- a/efi/preinstall/check_host_security.go
+++ b/efi/preinstall/check_host_security.go
@@ -106,6 +106,8 @@ Loop:
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


ZeyadYasser (approved):
Thank you!

valentindavid (approved):

PR #465, opened 2025-11-07
efi/fw_load_handler.go: copy variables for disable deployed mode

valentindavid:
In case of Ubuntu Core with deployed mode disabled using UEFI 2.5, then AuditMode and DeployedMode will be measured. Preinstall check is not in use. We have allowed this prior to snapd 2.71, and we should allow users to update passed it.


pedronis (approved):
do we need to make this optional using some With* at some future point?

otherwise +1

Issue #18, opened 2020-03-16
Investigate switching to go.mozilla.org/pkcs7


chrisccoulson:
We currently depend on a fork of github.com/fullsailor/pkcs7 which contains a bug fix to work with Authenticode signatures. It seems that Mozilla have a better maintained fork of this repo, and we should investigate switching to it instead.

PR #466, opened 2025-11-07
efi/preinstall: Make Intel BootGuard tests work with CSME #18

chrisccoulson:


efi/preinstall/check_host_security_intel_csme18.go
```diff
@@ -0,0 +1,95 @@
...
+	// code in bits 7:2 are valid. This bit is set of my XPS16 with CSME #18
+	// and unset on my older XPS15.
+	if regs.Hfsts5&hfsts5Csme18BtgProfileValid == 0 {
+		return &NoHardwareRootOfTrustError{errors.New("invalid BootGuard profile")}
+	}
+
+	// Check the BootGuard profile.
+	switch regs.Hfsts5.btgProfile() {
+	case btgProfileFVE, btgProfileFVME:
+		// ok
```
pedronis: same

chrisccoulson:
@kukrimate would you mind taking a look at this please?

kukrimate:
This looks good to me!
The logic looks like what we want to do, the list of BootGuard profiles supported is a good subset of the strictly allowable ones. Allowing the measurement only profile, if the platform correctly implements localities would also make sense, but since the measurement only profile isn't often used, this isn't high priority.
I don't have any Intel documentation to double check the bit locations, but I am sure Chris has tested the ones we care about to be right.

PR #468, opened 2025-11-12
Implement new options for external key data

chrisccoulson:
This replaces the existing `WithExternalKeyData` option which previously
accepted a variable number of `ExternalKeyData` structures with a new
`WithExternalKeyDataFromReader` option that accepts a name and a
`KeyDataReader` (which will eventually be replaced with `io.Reader`). The
new API can be supplied an arbitrary number of times.

Alongside this is a replacement for `WithExternalKeyData` which behaves
the same as `WithExternalKeyDataFromReader` but accepts an already
constructed `KeyData` instead.


pedronis (approved):
looks reasonable, probably @valentindavid should also check that it fits snapd needs

PR #470, opened 2025-11-21
efi/preinstall: reordered some preinstall check actions 

ernestl:
Reordered some error-kinds actions according in accordance with prioritization guideline and use new helper to insert ActionProceed in accordance to the guideline.

This proposal is a result of discussions around action order that happened during review of [Using the Snapd Error Kinds and Actions APIs](https://docs.google.com/document/d/1KXQhoH0vKbrxYKi0iPul4tkceWPtKPyYQbbkXU2TZHw/edit?usp=sharing).

It was also suggested that we do not need manual actions at all, given the current manual actions all have automatic PPI based counterparts, however I though to keep them in place for the moment while its still early days, so that we have an additional path to achieve actions.

Also corrected a few spelling mistakes.

### DIFF
--- a/efi/preinstall/checks_context.go
+++ b/efi/preinstall/checks_context.go
@@ -336,6 +336,24 @@ func (c *RunChecksContext) isActionExpected(action Action) bool {
 	return false
 }
 
+// insertActionProceed inserts [ActionProceed] into the actions slice.
+// It inserts before [ActionContactOEM] or [ActionContactOSVendor] if present,
+// otherwise appends to the end. This ensures [ActionProceed] (which ignores
+// the error) appears before "give up" actions per Rule 1.
+func insertActionProceed(actions []Action) []Action {
+	for i, action := range actions {
+		if action == ActionContactOEM || action == ActionContactOSVendor {
+			// Create new slice with capacity for the additional action.
+			result := make([]Action, len(actions)+1)
+			copy(result, actions[:i])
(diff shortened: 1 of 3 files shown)


chrisccoulson:
I think this looks ok to me, but I am a bit confused about the user experience. Whilst I understand how ordering the returned errors will be presented (that's something I'm going to work on before Christmas), I'm still a bit unsure how this will be presented - ie, is the intention to only present the first action, and how are the others presented if there is more than one?

ernestl:
> I think this looks ok to me, but I am a bit confused about the user experience. Whilst I understand how ordering the returned errors will be presented (that's something I'm going to work on before Christmas), I'm still a bit unsure how this will be presented - ie, is the intention to only present the first action, and how are the others presented if there is more than one?

Hi @chrisccoulson, thanks for having a look. The basic UX was captured here: [Using the Snapd Error Kinds and Actions APIs](https://docs.google.com/document/d/1KXQhoH0vKbrxYKi0iPul4tkceWPtKPyYQbbkXU2TZHw/edit?usp=sharing). (Requires your review).

- The installer handles a single error at a time, from the top down.
- For each error, it finds the associated actions (explicitly provided by secboot via snapd API or implicit as required for some special cases). Here the order of the actions as provided by secboot matters, and the installer will display the list of actions as options in that order. The user will pick the action they prefer from the list and press a button to kick it off.

PR #471, opened 2025-11-23
efi/preinstall: only indicate the mei_me module is missing if the ME PCI device exists

chrisccoulson:
A `MissingKernelModuleError` error with the value "mei_me" should only be
returned if the ME device cannot be found but the PCI device exists. If
the PCI device does not exist, then `UnsupportedPlatformError` is a more
appropriate error to return.

This also ensures that in the case where more than one device exists in
the mei subsystem, we pick the one associated with the ME (a PCI device
that is bound to the mei_me module). On some platforms, there may be
more than one device in the mei subsystem, eg, Intel's Vision Sensing
Controller, which is a platform device that is bound to the intel_vsc
module.

Note that the code in `internal/efi` has become a bit of a dumping ground
for things that aren't related to EFI, and I'm going to split this up in
a future PR.

### DIFF
--- a/activate.go
+++ b/activate.go
@@ -72,11 +72,12 @@ func (i *errorKeyslot) Data() KeyDataReader {
 
 // keyslotAttemptRecord binds together information about a keyslot.
 type keyslotAttemptRecord struct {
-	slot      Keyslot                     // The backend supplied Keyslot.
-	data      *KeyData                    // A cache of the decoded KeyData for platform keys.
-	flags     PlatformKeyDataHandlerFlags // The flags that the handling platform is registered with.
-	err       error                       // The first error that occurred with this keyslot.
-	errNumber int                         // The number of the error, used for ordering.
+	slot              Keyslot                     // The backend supplied Keyslot.
+	data              *KeyData                    // A cache of the decoded KeyData for platform keys.
+	flags             PlatformKeyDataHandlerFlags // The flags that the handling platform is registered with.
+	externalUnlockKey *externalUnlockKey          // An externally recovered unlock key.
+	err               error                       // The first error that occurred with this keyslot.
+	errNumber         int                         // The number of the error, used for ordering.
 }
 
 func (r *keyslotAttemptRecord) usable(flags activateOneContainerStateMachineFlags) bool {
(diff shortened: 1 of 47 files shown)


internal/efi/default_env.go
```diff
@@ -141,41 +199,30 @@ func (d *defaultEnvSysfsDevice) AttributeReader(attr string) (rc io.ReadCloser,
...
+		case dev, more := <-queue:
+			if !more {
+				return devices, nil
+			}
+			devices = append(devices, &defaultEnvSysfsDevice{
+				path:      dev.KObj,
+				props:     dev.Env,
+				subsystem: dev.Env["SUBSYSTEM"],
+			})
+			delete(dev.Env, "SUBSYSTEM")
```
pedronis: can you comment on the why of this delete?

internal/efi/default_env.go
```diff
@@ -141,41 +199,30 @@ func (d *defaultEnvSysfsDevice) AttributeReader(attr string) (rc io.ReadCloser,
...
+		case err := <-errs:
+			return nil, err
```
pedronis: this path doesn't seem to be tested

efi/preinstall/check_host_security_intel.go
```diff
@@ -239,14 +240,60 @@ func calculateIntelMEFamily(vers meVersion, hfsts1Reg hfsts1) meFamily {
...
+		if err != nil {
+			return fmt.Errorf("cannot determine parent device of %s: %w", candidate.Path(), err)
+		}
+
+		if parent.Subsystem() != "pci" {
+			// The ME should be a PCI device.
+			continue
+		}
+		if parent.Properties()["DRIVER"] != "mei_me" {
+			// The ME device should be bound to the mei_me driver.
```
pedronis: this bit of logic doesn't seem tested

pedronis (commented):
thanks, some small comments

internal/efi/default_env.go
chrisccoulson: I've added a comment for this now.

chrisccoulson:
I've just had to resolve some conflicts - I'll wait for the tests to complete again before merging.

PR #473, opened 2025-11-25
efi/preinstall: Remove ErrorKindTPMStartupLocalityNotProtected

chrisccoulson:
This error indicates that a partial mitigation againt discrete TPM reset
attacks cannot be enabled. This partial mitigation takes advantage of
the fact that PCR0 cannot be reconstructed from the OS if `TPM2_Startup`
is normally executed from a locality that isn't available (eg,
localities 3 and 4 on Intel systems, which are only available to Intel
ACMs).

There is nothing the user can do for this error other than continue,
ignoring it at their own risk. As it's going to be difficult to convey
useful information about this error to most users, it's better to just
drop it entirely, converting it instead to a warning that snapd can log
during install if it wants to.

Fixes FR-12028


pedronis (approved):
+1

PR #477, opened 2025-12-02
efi/preinstall: Refactor how the dTPM reset attack mitigation is enabled

chrisccoulson:
This adds a dedicated function for detecting whether the partial
mitigation against discrete TPM reset attacks should be enabled. It also
only enables it on Intel platforms, something that PR: #472 achieves by
making `isTPMDiscrete` lie instead.

The associated `CheckResultFlags` are also cleaned up, keeping backwards
compatibility with the original flags when unmarshalling.


chrisccoulson:
I've had to fix some merge conflicts in this now so I'll wait for the tests to complete again before landing.

PR #480, opened 2025-12-09
efi/preinstall: Support TPMs where TPM_PT_PS_FAMILY_INDICATOR is not TPM_PS_PC

chrisccoulson:
We check that the TPM is a PC-Client device because we can make
assumptions about its feature set based on what is in the TCG PC-Client
Platform TPM Profile spec. Some TPMs don't return the correct value
though - eg, some return TPM_PS_PDA. There is already a workaround for
the swtpm case which also returns an unexpected value.

This replaces the existing workaround and adds explicit feature
detection for these TPMs.

Fixes: FR-11806
Fixes: https://github.com/canonical/secboot/issues/448
Fixes: https://github.com/canonical/secboot/issues/460


efi/preinstall/check_tpm.go
```diff
@@ -30,17 +31,190 @@ const (
...
+	return true, nil
+}
+
 // checkTPM2DeviceFlags are passed to openAndCheckTPM2Device
 type checkTPM2DeviceFlags int
 
 const (
-	// checkTPM2DeviceInVM indicates that the current environment is a
-	// virtual machine.
-	checkTPM2DeviceInVM checkTPM2DeviceFlags = 1 << iota
```
pedronis: this goes away because we don't have to special case this anymore?

chrisccoulson: That's right

PR #482, opened 2025-12-12
efi/preinstall: Record accepted errors in CheckResult

chrisccoulson:
This records errors accepted via `ActionProceed` to a new `AcceptedErrors`
field of `CheckResult`.

There are 2 uses for this:
- To use accepted errors to turn on options when generating PCR profiles.
  The `WithAllowInsufficientDmaProtection()` opt-in makes use of this new
  field with this PR. Future PRs will add additional opt-ins which make
  use of this new field.
- To compare future check results against, in order to determine whether
  anything has changed.

At the moment, only the accepted error kinds are retained. However, the
new field is a map of `ErrorKind` to arbitrary JSON value so that a future
update can use the arbitrary JSON value to record arguments that can
limit the scope of an accepted error.

Fixes: FR-11919

### DIFF
--- a/efi/preinstall/checks_context_test.go
+++ b/efi/preinstall/checks_context_test.go
@@ -137,6 +137,7 @@ func (s *runChecksContextSuite) testRun(c *C, params *testRunChecksContextRunPar
 	c.Assert(params.actions, HasLen, iterations)
 
 	var result *CheckResult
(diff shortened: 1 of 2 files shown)


pedronis (commented):
question about test for fix

pedronis (approved):
thank you

PR #483, opened 2026-01-06
efi/preinstall: Ignore drivers loaded from firmware volumes

chrisccoulson:
The point of returning an error when addon drivers are detected is to
notify the user of the presence of code that is authenticated with a
secure boot authority that we have marked as not trusted to do so (which
is all of them today because we only detect the Microsoft UEFI CAs).

With this in mind, we want to ignore any drivers that are not
authenticated using the secure boot authorized signature database. This
means that we can ignore firmware blobs (as these would have to be
verified using some trust anchor embedded directly in the platform
firmware). For PE images, the platform firmware will contain some policy
that determines whether secure boot image verification is required, based
on the source of the image (eg, flash volume, internal storage, removable
storage, option ROM). Whilst we can't determine this policy directly, it
will generally be configured for no image verification from flash volumes
(as flash volumes are verified earlier on during the PEI phase), and
configured to require image verification from internal storage, removable
storage and option ROMs. This means that we can ignore drivers that are
loaded from flash volumes. In any case, these drivers are not really addon
drivers because they are loaded from the SPI flash, but some firmware
implementations are measuring the launch of drivers for things like onboard
network interfaces that are loaded from a flash volume.

This should reduce a bit of noise.

Fixes: FR-12185

### DIFF
--- a/efi/preinstall/check_pcr2.go
+++ b/efi/preinstall/check_pcr2.go
@@ -30,6 +30,29 @@ import (
 	internal_efi "github.com/snapcore/secboot/internal/efi"
 )
 
+// isLaunchedFromFirmwareVolume indicates that the supplied event is associated
+// with an image launch from a firmware volume.
+func isLaunchedFromFirmwareVolume(ev *tcglog.Event) (yes bool, err error) {
+	// The caller should check this.
+	switch ev.EventType {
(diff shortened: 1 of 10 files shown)


pedronis (commented):
looks good but a clarification question

PR #485, opened 2026-01-13
efi/preinstall: Remove ErrorKindUEFIDebuggingEnabled

chrisccoulson:
This is just a more specific ErrorKindHostSecurity error which is the
error that is returned if the platform is in manufacturing mode or if
BootGuard (for Intel) or PSB (for AMD) is not properly enabled. There's
no need to special case the presence of a debugger in the firmware, and
they have the same actions as well.

Fixes: FR-12151


pedronis:
@chrisccoulson there are conflicts now

PR #486, opened 2026-01-13
efi/preinstall: Remove ErrorKindPCRUnsupported

chrisccoulson:
This error kind is a temporary one whilst we lack support for certain
PCRs. Rather than having a completely separate error, instead:
- Re-use the same `ErrorKindPCRUnusable` error that is returned for PCRs
  that fail tests.
- Move the github issue URLs that were supplied with the
  `ErrorKindPCRUnsupported` errors to the error strings.
- Remove the `ActionContactOEM` action for errors associated with
  unsupported PCRs, as this is not appropriate.

Fixes: FR-12152


pedronis (commented):
looks good but code org question

pedronis (approved):
+1 but suggestion

PR #487, opened 2026-01-13
cmd/test_efi_fde_compat: Update error messaging

chrisccoulson:
The command prints "not suitable for FDE" on any error, but this is not
necessarily the case if the returned errors can be rectified in some
way.


pedronis (commented):
question

PR #488, opened 2026-01-14
efi/preinstall: Rename ErrorKindPreOSVerificationUsingDigests

chrisccoulson:
This changes an error name that I've never been particularly happy
about (what verification? and all authentication ultimately uses a
digest whether it is signed or not) - it's now
`ErrorKindPreOSSecureBootAuthByEnrolledDigests`.

There will be a corresponding error kind for OS components as an
additional check in the future as well.


ernestl (approved):
Thanks

PR #496, opened 2026-01-23
Make AuthRequestor indicate the returned credential type

chrisccoulson:
This extends `AuthRequestor.RequestUserCredential` to indicate the type of
the returned credential. Whilst the existing systemd and plymouth
implementations of `AuthRequestor` will always return the requested
credential types (as we don't know what the user intended to enter), a
future boot UI may provide a way for the user to decide whether to enter
a passphrase, PIN or recovery key and for this to be communicated back
to us.

Whilst this isn't particularly useful right now, I wanted to add the
logic to `ActivateContext` because it will affects the logic of a follow-up
PR for sending error messages to Plymouth.

Fixes: FR-12439

### DIFF
--- a/activate.go
+++ b/activate.go
@@ -661,22 +661,24 @@ func (m *activateOneContainerStateMachine) tryWithUserAuthKeyslots(ctx context.C
 			break
 		}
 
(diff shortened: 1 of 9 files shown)


pedronis (approved):
+1

PR #497, opened 2026-01-27
efi/fw_load_handler.go: only add DeployedMode/AuditMode if it is in log

valentindavid:


chrisccoulson (approved):
Thanks - I do have another idea for this, but this change is ok as a temporary fix for the immediate customer issue. I'm going to add the ability to proceed during install if the system is in user mode rather than deployed mode, and that will involve an additional option to enable support for user mode here (`WithSecureBootUserMode` or something). With that option, we'll generate 2 branches (one for deployed mode without the additional measurements and one for user mode with the additional measurements) which will enable a user who proceeded with user mode to enable deployed mode without needing a recovery key. Without the option, only deployed mode would be supported as we don't want to automatically create profiles for user mode if the user didn't previously choose this.

That would indirectly solve the issue here because you could just use `WithSecureBootUserMode` outside of the options created by `CheckResult` and you would get a profile with both branches.

PR #499, opened 2026-02-02
AuthRequestor: Add NotifyUserAuthResult API

chrisccoulson:
This adds a new API to `AuthRequestor` so that `ActivcateContext` can notify
the user about auth results. It notifies about success, failure and when
the user supplied credential is badly formatted for the indicated type.
It also provides a way to indicate when there are no more tries left for
a type.

The Plymouth implementation of this sends messages to Plymouth to be
displayed. The systemd implementation of this logs errors to stderr.

Fixes: FR-12404

### DIFF
--- a/auth_requestor_systemd.go
+++ b/auth_requestor_systemd.go
@@ -38,7 +38,7 @@ import (
 type SystemdAuthRequestorStringFn func(name, path string, authTypes UserAuthType) (string, error)
 
 type systemdAuthRequestor struct {
-	stderr   io.Writer
+	console  io.Writer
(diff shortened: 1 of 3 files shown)


valentindavid (approved):

auth_requestor.go
```diff
@@ -38,6 +42,45 @@ const (
 	UserAuthTypeRecoveryKey
 )
 
+func formatUserAuthTypeString(authTypes UserAuthType) string {
```
pedronis: question, any reason for this not to be a String method instead?

PR #510, opened 2026-02-26
tpm2: Make EnsureProvisioned accept a variable length set of options

chrisccoulson:
Future PRs are going to introduce additional parameters to this function
and change its behaviour. Having it accept a variable length set of
options is going to scale better than the current API.

This PR doesn't change any of its behaviour yet.


pedronis (approved):
+1

PR #532, opened 2026-04-01
efi: add CheckImageSignatureIsValidForHost function

zyga:
The new function allows a update manager to look if the given image is signed with keys that are allowed by the host secure boot stack. This way an update manager can detect that an asset signature is unlikely to pass verification by the boot firmware before writing or updating the asset in persistent storage.

### DIFF
--- a/efi/export_test.go
+++ b/efi/export_test.go
@@ -186,6 +186,14 @@ func MockOpenPeImage(fn func(Image) (peImageHandle, error)) (restore func()) {
 	}
 }
 
+func MockCheckImageSignatureIsValidForHost(fn func(context.Context, Image) error) (restore func()) {
+	orig := mockedCheckImageSignatureIsValidForHost
+	mockedCheckImageSignatureIsValidForHost = fn
+	return func() {
+		mockedCheckImageSignatureIsValidForHost = orig
+	}
+}
+
 func MockSnapdenvTesting(testing bool) (restore func()) {
 	orig := snapdenvTesting
 	snapdenvTesting = func() bool { return testing }
--- a/efi/image_trust.go
+++ b/efi/image_trust.go
@@ -1,0 +1,296 @@
+// -*- Mode: Go; indent-tabs-mode: t -*-
+
+/*
+ * Copyright (C) 2026 Canonical Ltd
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License version 3 as
+ * published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+package efi
+
+import (
+	"bytes"
+	"context"
+	"crypto"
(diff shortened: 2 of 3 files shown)


efi/image_trust.go
```diff
@@ -0,0 +1,296 @@
...
+			return xerrors.Errorf("cannot read authorized signature database: %w", err)
+		}
+		// If DB doesn't exist, then there are no authorized signatures, so the image cannot be
+		// authorized.
+		return errNoTrustedSignature
+	}
+
+	digests := newImageDigestCache(pei)
+
+	for _, sigList := range db {
```
valentindavid: Iterating through db then through signatures seems wrong. I think the spec specifies the way the loop is and it is the opposite. We should check. While the result will be the same, the certificate that will used to validate might be different, and if this function evolves for example to return which certificate would be measured... that will be wrong.

efi/image_trust.go
```diff
@@ -0,0 +1,296 @@
...
+				for _, imageSig := range imageSigs {
+					if !imageSig.CertWithIDLikelyTrustAnchor(efi.NewX509CertIDFromCertificate(cert)) {
+						continue
+					}
+
+					// If the signature chains to a trusted certificate, then the image is authorized.
+					return nil
+				}
+			}
+		} else {
```
valentindavid: Something I would like to check the specs. Is something signed never checked for its digest?

efi/export_test.go
```diff
@@ -186,6 +186,14 @@ func MockOpenPeImage(fn func(Image) (peImageHandle, error)) (restore func()) {
 	}
 }
 
+func MockCheckImageSignatureIsValidForHost(fn func(context.Context, Image) error) (restore func()) {
```
chrisccoulson: This isn't used anywhere

chrisccoulson (commented):
Thanks for working on this. I've done an initial pass and left some comments.

efi/image_trust.go
```diff
@@ -0,0 +1,296 @@
...
+	// Check for forbidden signatures in DBX first, to give a more specific
+	// error if the image is actually signed by a trusted CA but is revoked
+	// by DBX.
+	if err := checkDbxRevocation(ctx, pei, sigs); err != nil {
+		return err
+	}
+
+	// Check for a trusted signature in DB.
+	return checkDbAuthorization(ctx, pei, sigs)
+}
```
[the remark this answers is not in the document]
zyga: Probably yes

zyga:
@chrisccoulson thank you for the review. I need to spend a moment to iterate on the changes.

efi/image_trust.go
```diff
@@ -0,0 +1,296 @@
...
+
+	efi "github.com/canonical/go-efilib"
+	"golang.org/x/xerrors"
+)
+
+var mockedCheckImageSignatureIsValidForHost func(ctx context.Context, image Image) error
+
+// CheckImageSignatureIsValidForHost checks whether the supplied image has at
+// least one Authenticode signature that is authorized by the host's authorized
+// signature database (the UEFI "db" variable).
```
ndyer, line 37: Frederic suggested that we could test dbx as well.

efi/image_trust.go
```diff
@@ -0,0 +1,296 @@
...
+			}
+
+			if sig.CertWithIDLikelyTrustAnchor(efi.NewX509CertIDFromCertificate(revokedCert)) {
+				return true
+			}
+		}
+	}
+
+	return false
+}
```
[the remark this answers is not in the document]
valentindavid: That is a good point. We can do revocation checks in a separate part not to delay.

frederic-hoerni:
Finally it was not needed.

PR #536, opened 2026-04-09
efi/preinstall: Stop relying on BootCurrent

### LINKED ISSUE #517: PCR_UNUSABLE error with secure boot policy (PCR7)
- Image: 25th Feb, https://cdimages.ubuntu.com/ubuntu/daily-live/pending/resolute-desktop-amd64.iso
 - Machine: HP - Elite SFF 800 G9 Desktop PC -https://certification.canonical.com/hardware/202404-33976/ (Intel(R) Core(TM) i7-14700)

```
PCR_UNUSABLE
error with secure boot policy (PCR7) measurements: unexpected
EV_EFI_BOOT_SERVICES_APPLICATION event for
\PciRoot(0x0)\Pci(0x14,0x0)\USB(0x14,0x0)\HD
(2,GPT,20e7e352-9438-47f0-9de3-cd34ef1e18ba)\EFI\BOOT\BOOTX64.EFI
after already seeing a verification event during the OS-present
environment. This event should be for the initial boot loader
```

The test tool was reporting:
```
"Testing this platform for compatibility with EFI based TPM protected FDE

This platform may be suitable for FDE if the following problem is fixed:
error with system security: no kernel IOMMU support was detected
"
```

Image

### LINKED ISSUE #519: cannot access EFI_VARIABLE on hardware (AMD Ryzen AI 5)
Image: 1st Mar, https://cdimages.ubuntu.com/ubuntu/daily-live/pending/resolute-desktop-amd64.iso
Machine: https://certification.canonical.com/hardware/202509-37886/

```
Hardware-backed encryption could not be enabled
There is an issue with this computer's firmware.
Contact IT support, or choose a different encryption method.
Technical details
EFI_VARIABLE_ACCESS
cannot access EFI variable: cannot read BootCurrent variable: variable does not exist
```

Image

chrisccoulson:
The PCR4 and PCR7 checks were relying on the `BootCurrent` EFI variable to
identify the `EV_EFI_BOOT_SERVICES_APPLICATION` event in the TCG log that
corresponds to the launch of the initial OS loader. This is proving
problematic on some systems:
- Whilst some firmware populates a Boot variable containing a load
  option with a path matching the OS loader and sets `BootCurrent` to
  point to this when booting from removable media, some firmware
  implementations dosn't set a path that matches the OS loader, and this
  path doesn't match any event in the TCG log.
- Some systems don't seem to have `BootCurrent` set at all when booting
  from removable media.

Instead, assume that the first OS-present
`EV_EFI_BOOT_SERVICES_APPLICATION` event that isn't Absolute is the
initial OS loader. If it isn't, the PCR4 checks will fail as expected
because os the mismsatched digest. We have to have a special case for
Absolute here because some systems measure this launch as part of
OS-present.

This also makes the checks a bit more permissive where firmware bugs
don't cause problems with generating profiles:
- The PCR4 checks no longer enforce that there can be only one
  `EV_OMIT_BOOT_DEVICE_EVENT` event.
- The PCR4 checks no longer enforce that there cannot be a `EV_EFI_ACTION`
  "Calling EFI Application from Boot Option" if there is a
  `EV_OMIT_BOOT_DEVICE_EVENT` event. We already don't enforce the
  existence of the `EV_EFI_ACTION` event if there is no
  `EV_OMIT_BOOT_DEVICE_EVENT` event even though it should be there,
  because we know that some firmware misbehaves here.
- The PCR4 checks no longer gate the reading of the `SysPrepOrder` and
  SysPrep variables on the `BootOptionSupport` flags.

Fixes: https://github.com/canonical/secboot/issues/517
Fixes: https://github.com/canonical/secboot/issues/519


valentindavid (approved):

cmmrandau:
when will this reach the installer? can't install ubuntu with fde on my x1 carbon

frugaltech:
Just adding a problem report, echoing the above:
* Installing Ubuntu 26.04 in Qemu is problematic for the same reason when using emulated TPM alongside OVMF/OVMF_CODE_4M.secboot.fd.
* Same is true for an Emerson IPC6010 industrial PC.
* Enabling the hardware backed encryption works fine post-install as outlined here: https://askubuntu.com/questions/1565961/pcr-unusable-when-installing-ubuntu-26-04-with-hardware-backed-encryption

pedrohqb:
Did this reach 26.04.1?

frederic-hoerni:
> Did this reach 26.04.1?

Yes. `ubuntu-26.04.1-desktop-amd64.iso` has snapd-2.76.3, which is linked to a revision of secboot with this fix:
https://github.com/canonical/snapd/blob/2.76.3/go.mod#L25

PR #539, opened 2026-04-22
tpm2: Support safely rotating the lockout hierarchy authorization value

chrisccoulson:
This updates the `Connection.EnsureProvisioned` API to safely support
rotating the authorization value for the lockout hierarchy, which can be
used during reprovisioning. This is particularly required during a
factory reset, when secrets should be rotated.

The lockout hierarchy is used on successful boots to reset the DA
counter. A single authorization failure makes the lockout hierarchy
unavailable for the preprogrammed recovery time (currently 24 hours).
This makes it challenging to update the value because it's not possible
to atomically update both the value in the TPM and the value stored in
persistent storage.

This PR works around this by using a transient authorization policy that
permits the use of the `TPM2_HierarchyChangeAuth` command with a signed
assertion (`TPM2_PolicySigned`), using a key that's derived from the
original authorization value. An authorization policy is used by default
for other uses of the lockout hierarchy as this provides a way to detect
a transient state resulting from a previously interrupted update before
using the incorrect authorization value.

The `WithProvisionNewLockoutAuthValue` option no longer allows an
authorization value to be specified directly. One is derived
automatically from the supplied random source, with the length
determined by the value of the `TPM_PT_CONTEXT_HASH` property. A callback
is supplied so that the authorization data can be stored to persistent
storage. This happens several times during an update.

### DIFF
--- a/tpm2/lockoutauth.go
+++ b/tpm2/lockoutauth.go
@@ -96,6 +96,12 @@ type lockoutAuthParams struct {
 	AuthPolicy    *policyutil.Policy
 	NewAuthValue  tpm2.Auth
 	NewAuthPolicy *policyutil.Policy
+
+	// noAuthValue is a special value only set by WithUnconfiguredLockoutAuth so
+	// that we can return an appropriate error rather than triggering a lockout
+	// if the option is supplied when the lockout hierarchy already has an
+	// authorization value.
+	noAuthValue bool
 }
 
 func (p *lockoutAuthParams) MarshalJSON() ([]byte, error) {
@@ -302,6 +308,10 @@ func newLockoutAuthValueUpdateStateMachine(rand io.Reader, tpm *Connection, auth
 	return m, nil
 }
 
+func (m *lockoutAuthValueUpdateStateMachine) authorizeLockout(allowFallbackToHMACSession bool, command tpm2.CommandCode) (session tpm2.SessionContext, done func(), err error) {
+	return m.tpm.authorizeLockout(m.authParams, allowFallbackToHMACSession, command, m.signedAuthorizer)
+}
+
 func (m *lockoutAuthValueUpdateStateMachine) signedAuthorizer(sessionAlg tpm2.HashAlgorithmId, sessionNonce tpm2.Nonce, authKey tpm2.Name, policyRef tpm2.Nonce) (*policyutil.PolicySignedAuthorization, error) {
 	params := &policyutil.PolicySignedParams{
 		HashAlg:  sessionAlg,
@@ -353,7 +363,7 @@ func (m *lockoutAuthValueUpdateStateMachine) prepare() (lockoutAuthValueUpdateSt
 // temporary policy to update it's authorization value to a specific value, and to update it's policy
 // using it's authorization value. Note that it is not safe to use the authorization value yet though.
 func (m *lockoutAuthValueUpdateStateMachine) setNewAuthValuePolicy(policyAlg tpm2.HashAlgorithmId, policyDigest tpm2.Digest) (lockoutAuthValueUpdateStateMachineState, error) {
-	session, done, err := m.tpm.authorizeLockout(m.authParams, true, tpm2.CommandSetPrimaryPolicy, m.signedAuthorizer)
+	session, done, err := m.authorizeLockout(true, tpm2.CommandSetPrimaryPolicy)
 	switch {
 	case errors.Is(err, errLockoutAuthPolicyNotSupported):
 		return m.setAuthValueWithoutPolicy, nil
@@ -378,7 +388,7 @@ func (m *lockoutAuthValueUpdateStateMachine) setAuthValueWithoutPolicy() (lockou
 	m.authParams.AuthPolicy = nil
 	m.authParams.NewAuthPolicy = nil
 
-	session, done, err := m.tpm.authorizeLockout(m.authParams, false, tpm2.CommandHierarchyChangeAuth, m.signedAuthorizer)
+	session, done, err := m.authorizeLockout(false, tpm2.CommandHierarchyChangeAuth)
 	if err != nil {
 		return nil, err
 	}
@@ -395,6 +405,7 @@ func (m *lockoutAuthValueUpdateStateMachine) setAuthValueWithoutPolicy() (lockou
 
 	m.authParams.AuthValue = m.authParams.NewAuthValue
 	m.authParams.NewAuthValue = nil
+	m.authParams.noAuthValue = false
 
 	return nil, nil
 }
@@ -404,7 +415,7 @@ func (m *lockoutAuthValueUpdateStateMachine) setAuthValueWithoutPolicy() (lockou
 // On completion, the updated state can only be used to authorize the lockout hierarchy using the
 // temporary policy to set a new policy using an authorization value.
 func (m *lockoutAuthValueUpdateStateMachine) setNewAuthValue(policyAlg tpm2.HashAlgorithmId) (lockoutAuthValueUpdateStateMachineState, error) {
-	session, done, err := m.tpm.authorizeLockout(m.authParams, false, tpm2.CommandHierarchyChangeAuth, m.signedAuthorizer)
+	session, done, err := m.authorizeLockout(false, tpm2.CommandHierarchyChangeAuth)
 	if err != nil {
 		return nil, err
 	}
@@ -426,6 +437,7 @@ func (m *lockoutAuthValueUpdateStateMachine) setNewAuthValue(policyAlg tpm2.Hash
 
 	m.authParams.AuthValue = m.authParams.NewAuthValue
 	m.authParams.NewAuthValue = nil
+	m.authParams.noAuthValue = false
 
 	var newPolicyDigest tpm2.Digest
 	newPolicyDigest, m.authParams.NewAuthPolicy, err = newDefaultLockoutAuthPolicy(policyAlg)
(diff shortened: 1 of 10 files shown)


tpm2/provisioning.go
```diff
@@ -337,8 +347,6 @@ func (t *Connection) EnsureProvisioned(options ...EnsureProvisionedOption) error
 			return AuthFailError{tpm2.HandleLockout}
 		case tpm2.IsTPMWarning(err, tpm2.WarningLockout, command):
 			return ErrTPMLockout
-		case tpm2.IsTPMSessionError(err, tpm2.ErrorPolicyFail, command, 1):
-			return ErrInvalidLockoutAuthPolicy
```
pedronis, line 341: why was this dropped?

tpm2/provisioning_test.go
```diff
@@ -95,22 +97,48 @@ func (s *provisioningSimulatorSuite) SetUpTest(c *C) {
...
+	syncLockoutAuthData := func(data []byte) error {
+		c.Logf("%s", string(data))
+		c.Assert(expectedLockoutAuthData, Not(HasLen), 0)
+		expected := expectedLockoutAuthData[0]
+		expectedLockoutAuthData = expectedLockoutAuthData[1:]
+		c.Check(data, DeepEquals, expected)
+		return nil
+	}
+
+	opts := []EnsureProvisionedOption{WithLockoutAuthValue(nil), WithProvisionNewLockoutAuthValue(bytes.NewReader(data.lockoutAuthBytes), syncLockoutAuthData)}
```
pedronis: these options is what to try if we didn't have a stored auth data/params yet?

tpm2/provisioning.go
```diff
@@ -248,20 +251,22 @@ func WithLockoutAuthData(data []byte) EnsureProvisionedOption {
...
-// then this option has no effect.
-func WithProvisionNewLockoutAuthValue(authValue []byte) EnsureProvisionedOption {
+// WithProvisionNewLockoutAuthValue tells [Connection.EnsureProvisioned] to set or update the authorization
+// value and authorization policy for the lockout hierarchy. The caller supplies a callback which is used
+// to store the authorization data to persistent storage. This will be called multiple times during the
+// update.
+//
+// This option will also resume a previously interrupted update, as long as the most recent authorization
+// data is supplied to [WithLockoutAuthData].
+func WithProvisionNewLockoutAuthValue(rand io.Reader, syncData func([]byte) error) EnsureProvisionedOption {
```
valentindavid: Would it make more sense to call it `WithProvisionNewLockoutAuthData` (Data not Value) so people do not think they should call `WithLockoutAuthValue`. Or maybe use some types for []byte to make them distinct.

tpm2/lockoutauth.go
```diff
@@ -115,35 +140,414 @@ func (p *lockoutAuthParams) UnmarshalJSON(data []byte) error {
...
+			// dictionary attack protection. This is only a transient state so this is
+			// ok but not ideal. I would have liked to have used a TPM2_PolicyCpHash
+			// assertion here to bind the policy to the new authorization value so that
+			// this branch would require knowledge of both the old and new values.
+			// However, TPM2_PolicyCpHash doesn't work here because we use parameter
+			// encryption and the CpHash is computed after the command parameters are
+			// encrypted.
+			//
+			// As the policy is not hardcoded in the metadata, it would be trivial to
+			// update this policy branch in the future in order to make improvements.
```
valentindavid, line 215: Does that mean that the device is stolen right in this state, they could disable lockout?

tpm2/lockoutauth.go
```diff
@@ -153,68 +557,99 @@ func (t *Connection) authorizeLockout(authParams *lockoutAuthParams, command tpm
...
-	if !lockoutAuthSet {
-		return ErrEmptyLockoutAuthValue
+	switch {
+	case len(t.LockoutHandleContext().AuthValue()) == 0:
+		// authorization was performed with an empty auth value.
+		return ErrLockoutAuthNotInitialized
+	case params.AuthPolicy != nil && session.Handle().Type() == tpm2.HandleTypeHMACSession:
+		// authorization was performed with a HMAC session when we have policy data,
+		// which only happens if the lockout hierarchy has no policy set.
+		return ErrLockoutAuthNotInitialized
```
pedronis, line 645: tests don't' end up here

tpm2/lockoutauth.go
```diff
@@ -115,35 +140,414 @@ func (p *lockoutAuthParams) UnmarshalJSON(data []byte) error {
...
+
+	// We use command parameter encryption here to protect the new authorization value.
+	switch {
+	case session.Handle().Type() == tpm2.HandleTypePolicySession:
+		// We're using policy auth so need to supply the HMAC session as an extra
+		// session for parameter encryption.
+		err = m.tpm.HierarchyChangeAuth(m.tpm.LockoutHandleContext(), m.authParams.NewAuthValue, session, m.tpm.HmacSession().IncludeAttrs(tpm2.AttrCommandEncrypt))
+	default:
+		// We're using HMAC auth
+		err = m.tpm.HierarchyChangeAuth(m.tpm.LockoutHandleContext(), m.authParams.NewAuthValue, session.IncludeAttrs(tpm2.AttrCommandEncrypt))
```
pedronis: this case is not reached by tests

pedronis (commented):
thank you, did a pass

tpm2/lockoutauth.go
```diff
@@ -115,35 +140,414 @@ func (p *lockoutAuthParams) UnmarshalJSON(data []byte) error {
...
+		policy    *policyutil.Policy
+		policyAlg tpm2.HashAlgorithmId
+	)
+	if authParams.AuthPolicy != nil || authParams.NewAuthPolicy != nil {
+		// Only do this if there is policy data.
+		switch currentPolicyDigest, err := t.GetCapabilityAuthPolicy(tpm2.HandleLockout); {
+		case tpm2.IsTPMParameterError(err, tpm2.ErrorValue, tpm2.CommandGetCapability, 1):
+			// TPM_CAP_AUTH_POLICIES is unsupported on TPMs older than v1.38 of the
+			// reference library spec.
+			return nil, nil, errLockoutAuthPolicyNotSupported
```
valentindavid, line 494: What if we have customers with old TPMs? It seems the new code will not allow to set a old fashion authorization value on a new installation. So we are dropping support for those?

chrisccoulson: This case still works and there is a test for it: https://github.com/canonical/secboot/pull/539/changes#diff-a27fbdd788c4c6dee55d96aab23aa9819f6ad51d7140573d2e74c24546857a28R223.

It's handled here: https://github.com/canonical/secboot/pull/539/changes#diff-cebc0b5ead0879a4638a43750d7b66c2f5bb421cd1942fae8ab1348dbc1fb324R353

tpm2/provisioning_test.go
chrisccoulson: Yes, although perhaps there should be an explicit `WithUnconfiguredLockoutAuth` option instead.

tpm2/lockoutauth.go
valentindavid: Nevermind, I see we do `setAuthValueWithoutPolicy` in that case.

tpm2/provisioning.go
chrisccoulson: It's impossible to hit this case now because `authorizeLockout` returns `ErrInvalidLockoutAuthPolicy` instead.

tpm2/lockoutauth.go
chrisccoulson: It doesn't - this branch permits changing the authorization value but it still requires knowledge of the original value which is a 32 byte key on TPMs with TPM_PT_CONTEXT_HASH=sha256, as this is used to derive the key required to sign the authorization. What it means is that this branch isn't subject to dictionary attack protection.

tpm2/provisioning.go
chrisccoulson: That sounds ok.

tpm2/lockoutauth.go
```diff
@@ -115,35 +140,414 @@ func (p *lockoutAuthParams) UnmarshalJSON(data []byte) error {
...
+
+	m.authParams.AuthPolicy = m.authParams.NewAuthPolicy
+	m.authParams.NewAuthPolicy = nil
+
+	return func() (lockoutAuthValueUpdateStateMachineState, error) {
+		return m.setNewAuthValue(policyAlg)
+	}, nil
+}
+
+func (m *lockoutAuthValueUpdateStateMachine) setAuthValueWithoutPolicy() (lockoutAuthValueUpdateStateMachineState, error) {
```
valentindavid, line 382: What are the drawbacks in that case compared to policies? Is that we could get in a situation we do not know which authorization value is the right one?
Do we write the new authorization value to the data before we set this?

chrisccoulson: This case only works when going from an empty authorization value to a non-empty value (ie, during the initial install). If `WithProvisionNewLockoutAuthValue` is used again, then an error is returned (`ErrLockoutAuthUpdateUnsupported`)

tpm2/lockoutauth.go
chrisccoulson: Oh, that's weird - `lockoutauthSuite.TestResetDictionaryAttackLockAuthPolicyUnset` is meant to hit this.

chrisccoulson: Ah, it was a bug in the test :) https://github.com/canonical/secboot/pull/539/changes/a2ef3960b09a87900430f7d3e44eddfcdcd58d62

tpm2/lockoutauth.go
chrisccoulson: It turns out that it's not actually possible to hit this case, so I've removed it.

chrisccoulson:
Ok, I've pushed some updates:

- `WithProvisionNewLockoutAuthValue` is now `WithProvisionNewLockoutAuthData`.
- There is a new `WithUnconfiguredLockoutAuth` option to avoid having to use `WithLockoutAuthValue(nil)` on a fresh install.
- Along with this, there is a new `ErrLockoutAuthInitialized` which will be returned if `Connection.EnsureProvisioned` is called with the new option if the lockout hierarchy already has an authorization value (as an alternative to just tripping the lockout).
- I've merged `ErrInvalidLockoutAuthPolicy` and `ErrLockoutAuthUpdateInterrupted` into a new error - `ErrLockoutAuthInvalid`, as there's no need to distinguish between these 2 cases given that they have the same resolution (need to call `Connection.EnsureProvisioned` again with the `WithProvisionNewLockoutAuthData` option).

pedronis (approved):
thank you

valentindavid (approved):

PR #545, opened 2026-04-30
Changes going from 77bc2457cc76 (2.75.2) to 3f8b98c2db70 (2.74.1)

ernestl:


crypt.go
ernestl, line 1: NOT APPLICABLE

efi/fw_load_handler.go
ernestl, line 1: RESEALING RELATED

efi/preinstall/result.go
ernestl, line 1: EFFECTS PROFILE USED FOR SEALING

efi/preinstall/result.go
```diff
@@ -171,6 +177,8 @@ func (f *CheckResultFlags) UnmarshalJSON(data []byte) error {
 			val = NoSecureBootPolicyProfileSupport
 		case "request-partial-dtpm-reset-attack-mitigation":
 			val = RequestPartialDiscreteTPMResetAttackMitigation
+		case "require-lock-platform-firmware":
```
valentindavid, line 180: Likely this will go to `default:` instead an error with `unrecognized flag "require-lock-platform-firmware"`

PR #546, opened 2026-04-30
efi/secureboot.go: automatically detect deployedmode variable measurement

valentindavid:
e25b8214ea08f7784e792786da37631dee06ea34 (#508) regressed the fix from 1e93244b804eb9958160042a983b33187833a128 (#497). We need to read the event log for Ubuntu Core systems because we do not use the pre-install checks. So here we instead provide a
`WithAutoAllowSecureBootUserMode()` that can be called unconditionally on Core system and let the profile be calculated depdending what is in the event log.

### DIFF
--- a/efi/fw_load_handler.go
+++ b/efi/fw_load_handler.go
@@ -208,6 +208,19 @@ func (h *fwLoadHandler) measureSecureBootPolicyPreOS(ctx pcrBranchContext) error
 	// permitted with the WithSecureBootUserMode option and they are being included in this
 	// branch.
 	includeUserMode := boolParamOrFalse(ctx.Params(), includeSecureBootUserModeParamKey)
+	autoIncludeUserMode := boolParamOrFalse(ctx.Params(), autoIncludeSecureBootUserModeParamKey)
+	if autoIncludeUserMode {
+		for _, e := range h.log.Events {
+			switch {
+			case e.PCRIndex == internal_efi.SecureBootPolicyPCR && e.EventType == tcglog.EventTypeEFIVariableDriverConfig:
+				if data, ok := e.Data.(*tcglog.EFIVariableData); ok {
+					if data.VariableName == efi.GlobalVariable && data.UnicodeName == "AuditMode" {
+						includeUserMode = true
+					}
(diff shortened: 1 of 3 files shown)


efi/fw_load_handler.go
```diff
@@ -208,6 +208,19 @@ func (h *fwLoadHandler) measureSecureBootPolicyPreOS(ctx pcrBranchContext) error
 	// permitted with the WithSecureBootUserMode option and they are being included in this
 	// branch.
 	includeUserMode := boolParamOrFalse(ctx.Params(), includeSecureBootUserModeParamKey)
+	autoIncludeUserMode := boolParamOrFalse(ctx.Params(), autoIncludeSecureBootUserModeParamKey)
+	if autoIncludeUserMode {
+		for _, e := range h.log.Events {
+			switch {
+			case e.PCRIndex == internal_efi.SecureBootPolicyPCR && e.EventType == tcglog.EventTypeEFIVariableDriverConfig:
+				if data, ok := e.Data.(*tcglog.EFIVariableData); ok {
+					if data.VariableName == efi.GlobalVariable && data.UnicodeName == "AuditMode" {
```
pedronis, line 217: it's enough to check one of the two vars?

efi/secureboot.go
```diff
@@ -343,6 +344,26 @@ func (o allowSecureBootUserModeOption) ApplyOptionTo(visitor internal_efi.PCRPro
 	return nil
 }
 
+type autoAllowSecureBootUserModeOption struct{}
+
+func (o autoAllowSecureBootUserModeOption) ApplyOptionTo(visitor internal_efi.PCRProfileOptionVisitor) error {
+	visitor.AddImageLoadParams(func(params ...loadParams) []loadParams {
+		var out []loadParams
+		for _, v := range []bool{false, true} {
```
pedronis, line 352: why do we need to set both to true and false, in case the firmware behavior changes?

pedronis (commented):
couple of questions

valentindavid:
We might not need it. We still have a branch where the option is false.

PR #548, opened 2026-05-29
Remove flags field from NewAtomicFile calls

olivercalder:
The `osutil.AtomicWriteFlags` has had its only flag value removed from snapd, as that flag had been unused since 2016. Now, since all the `flags osutil.AtomicWriteFlags` fields were ignored, we want to remove those fields from snapd as well. However, snapd vendors secboot, and secboot uses `osutil.NewAtomicFile` from snapd, so we can't make the change in snapd without first making the change in secboot.

Once this secboot change lands and a new secboot build is available, then we can bump the secboot version used by snapd within the PR which changes the `osutil` function signatures. Then, after that PR within snapd lands, we can bump the snapd version which secboot builds against.

That PR within snapd is https://github.com/canonical/snapd/pull/17129

### DIFF
--- a/keydata_file.go
+++ b/keydata_file.go
@@ -69,7 +69,7 @@ type FileKeyDataWriter struct {
 }
 
 func (w *FileKeyDataWriter) Commit() error {
-	f, err := osutil.NewAtomicFile(w.path, 0600, 0, sys.UserID(osutil.NoChown), sys.GroupID(osutil.NoChown))
+	f, err := osutil.NewAtomicFile(w.path, 0600, sys.UserID(osutil.NoChown), sys.GroupID(osutil.NoChown))
 	if err != nil {
 		return xerrors.Errorf("cannot create new atomic file: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


frederic-hoerni:
I understand the will to simplify code, however I have a few worries there:
- this breaks the unit tests, so the PR cannot be merged, right?
- if this PR still gets merged, and if programs that use secboot upgrade to this revision, they will be broken as well
- after this PR and snapd's PR get merged, we'll bump the version of snapd used by secboot. I havn't analyzed the impact of this, and in particular if this will require additional changes in secboot. Have you?

Another solution (that unfortunately does not simplify things in the short term) would be:
- snapd: create `NewAtomicFile2`, that is like `NewAtomicFile`, but without the flags argument, and mark `NewAtomicFile` as deprecated
- secboot: use `NewAtomicFile2`
- snapd: later, remove `NewAtomicFile`

valentindavid:
I suppose this has to wait for the PR to land on snapd, and then update go.sum.

olivercalder:
Closing for now, as this would require careful coordination with snapd and other projects which consume secboot, for little to no real benefit.

PR #554, opened 2026-07-21
Quality improvements

frederic-hoerni:
- efi/pcr_profile: minor quality improvements
- Makefile: add check-efi
- efi: add pcr number on panic message
- efi/preinstall: add option --event-log

### DIFF
--- a/efi/pcr_profile_test.go
+++ b/efi/pcr_profile_test.go
@@ -28,8 +28,8 @@ import (
 
 	efi "github.com/canonical/go-efilib"
 	"github.com/canonical/go-tpm2"
+	"github.com/canonical/go-tpm2/policyutil"
 	tpm2_testutil "github.com/canonical/go-tpm2/testutil"
-	"github.com/canonical/go-tpm2/util"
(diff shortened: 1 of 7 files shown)


efi/pcr_profile_test.go
```diff
@@ -121,7 +121,7 @@ func (s *pcrProfileMockedSuite) TestAddPCRProfileSimple(c *C) {
 	profile := secboot_tpm2.NewPCRProtectionProfile()
 
 	var digests tpm2.DigestList
-	for i := 0; i <= 2; i++ {
+	for i := 0; i <= 1; i++ {
```
pedronis, line 124: what's this change about?

pedronis (commented):
one question

efi/pcr_profile_test.go
frederic-hoerni: The original loop creates 3 items in the list, but only the first 2 are used in the test.
This modification is to remove useless code and processing.

valentindavid (approved):

pedronis (approved):
+1

PR #555, opened 2026-07-21
AddPCRProfile: add WithAllowSecurityLevelDowngraded

### LINKED ISSUE #553: unexpected event type (EV_EFI_ACTION) found in log (Thunderbolt related)
A user reports the following issue.
Some BIOS firmware on NUC8v5PNB devices measures an event of type EV_EFI_ACTION saying "Security Level is Downgraded to 0", which breaks the computation of the PCR policy, with this error:

> cannot measure secure boot policy: unexpected event type (EV_EFI_ACTION) found in log

This can be fixed on these devices either by configuring the BIOS to the most secure option for Thunderbolt, or by updating the firmware.

Having a Thunderbolt security level of zero means that in some situations a malicious Thunderbolt device could access potentially sensitive data in your system’s memory.

In spite of that, because of operational constraints, the user would like an option to workaround this issue and still be able to use TPM-backed encryption.

frederic-hoerni:


efi/fw_load_handler.go
```diff
@@ -302,6 +312,23 @@ func (h *fwLoadHandler) measureSecureBootPolicyPreOS(ctx pcrBranchContext) error
...
+			// Now we use the includeThunderboltSecurityLevel0 flag to determine if
+			// this run of fwLoadHandler should really measure it to this profile
+			// branch as well. Including a branch in the profile that skips this
+			// measurement makes it possible for the firmware setting to be corrected
+			// without invalidating the profile.
+			// Future runs can then drop the option to permit it.
+			if includeThunderboltSecurityLevel0 {
+				ctx.ExtendPCR(internal_efi.SecureBootPolicyPCR, e.Digests[ctx.PCRAlg()])
+			}
+			allowThunderboltSecurityLevel0 = false // Only allow this event to appear once
```
valentindavid, line 334: I think we might be able to set `allowInsufficientDMAProtection` to false here. And also set `allowThunderboltSecurityLevel0` to false in previous case. I think those 2 measurements (dma protection disabled and security level downgraded) are mutually exclusive, right?

valentindavid (approved):
Looks fine. Though I think it is mutually exclusive with "DMA Protection Disabled". But this is merge-able as is.

internal/efitest/log.go
```diff
@@ -166,19 +187,20 @@ const (
...
-	NoCallingEFIApplicationEvent      bool                            // omit the EV_EFI_ACTION "Calling EFI Application from Boot Option" event.
-	IncludeOSPresentFirmwareAppLaunch efi.GUID                        // include a flash based application launch in the log as part of the OS-present phase
-	NoSBAT                            bool                            // omit the SbatLevel measurement to mimic older versions of shim
-	PreOSVerificationUsesDigests      crypto.Hash                     // Whether Driver or SysPrep launches are verified using a digest
-	DisableDeployedMode               bool                            // Whether deployed/audit modes are disabled and we have UEFI 2.5
+	StartupLocality                   uint8                               // specify a startup locality other than 0
+	FirmwareDebugger                  bool                                // indicate a firmware debugger endpoint is enabled
+	DMAProtection                     DMAProtectionDisabledEventFlags     // whether DMA protection is disabled
+	ThunderboltSecurityLevel0         ThunderboltSecurityLevel0EventFlags // whether security level is downgraded
```
pedronis: why not just a bool?

efi/pcr_profile_test.go
```diff
@@ -1378,6 +1378,72 @@ func (s *pcrProfileSuite) TestAddPCRProfileUC20WithDbxUpdateWithAllowInsufficien
...
+			NewImageLoadActivity(shim).Loads(
+				NewImageLoadActivity(grub, KernelCommandlineParams("console=ttyS0 console=tty1 panic=-1 systemd.gpt_auto=0 snapd_recovery_mode=recover")).Loads(
+					NewImageLoadActivity(grub, KernelCommandlineParams("console=ttyS0 console=tty1 panic=-1 systemd.gpt_auto=0 snapd_recovery_mode=run")).Loads(
+						NewImageLoadActivity(runKernel),
+					),
+				),
+			),
+		),
+		expected: []tpm2.PCRValues{
+			// Not including DMA protection string // TODO
```
pedronis: what is the TODO about?

pedronis (commented):
thanks, one nitpick/some questions

Btw do we need full WithAutoTCGPCRProfile integration with ActionProceed for this? should that be a follow up?

cc @valentindavid

valentindavid:
> Btw do we need full WithAutoTCGPCRProfile integration with ActionProceed for this? should that be a follow up?

We could. But given this is only very old firmware, maybe we should not do it. That probably will not add that many devices. If that opened for many devices that would be a bit different.

internal/efitest/log.go
frederic-hoerni: changed to a bool.

efi/fw_load_handler.go
frederic-hoerni: ok, done.

efi/pcr_profile_test.go
frederic-hoerni: fixed.

frederic-hoerni:
I took into account the remarks of the review. I will squash the related commits before merge.

efi/fw_load_handler.go
pedronis: does this change need a test?

pedronis (commented):
question

efi/fw_load_handler.go
frederic-hoerni: I will add a test for the "mutually exclusive" part.

frederic-hoerni: test added: TestMeasureImageStartErrAllowThunderboltSecurityLevel0WithInsufficientDMA

pedronis (approved):
+1

PR #556, opened 2026-07-26
efi: support HP pre-OS PCR measurements

ebrig:
## Summary

HP firmware can produce two pre-OS measurements that are not currently represented in generated EFI PCR profiles:

- a vendor-defined PCR2 event after `EV_SEPARATOR`, but before the initial OS loader is authorized or launched
- an `EV_EFI_ACTION` in PCR7 containing the enabled SVM and DMA protection configuration when additional DMA settings are measured to PCR7

This causes the generated profile to diverge from the TPM values and prevents a resealed key from unlocking normally.

This change:

- replays vendor-defined PCR2 events only within the interval between the PCR2 separator and the first OS authorization or application launch
- requires that an OS boundary is observed rather than accepting a truncated log
- recognizes HP's exact enabled SVM/DMA/all-PCIe pre-boot protection event in PCR7
- validates the PCR7 event digest and accepts the event only once, before Secure Boot configuration measurements
- applies the same rules in profile generation and preinstall validation

## Hardware validation

Tested on an HP ZBook Ultra G1a 14 inch Mobile Workstation PC with BIOS X89 01.05.07 (2026-05-05), using:

- SVM CPU Virtualization: Enable
- DMA protection: Enable
- Pre-boot DMA protection: All PCIe devices
- Measure Additional DMA Settings: PCR7

After resealing with the patched package, TPM-passphrase unlock succeeded across a reboot.

## Tests

- `GOTOOLCHAIN=go1.23.12 ./run-tests --with-mssim`
- `go build ./...` with Go 1.18.10 and Go 1.23.12
- `go vet` with Go 1.18.10 and Go 1.23.12
- focused PCR2 and PCR7 tests with Go 1.18.10 and Go 1.23.12

### DIFF
--- a/efi/fw_load_handler.go
+++ b/efi/fw_load_handler.go
@@ -392,18 +392,49 @@ func (h *fwLoadHandler) measurePlatformFirmware(ctx pcrBranchContext) error {
 }
 
 func (h *fwLoadHandler) measureDriversAndApps(ctx pcrBranchContext) error {
+	seenSeparator := false
+
 	for _, event := range h.log.Events {
+		// Some firmware extends vendor-defined events into PCR2 after the
+		// separator but before authorizing or launching the initial OS loader.
+		// Retain those events, but do not copy any PCR2 measurements made once
+		// OS image processing has begun.
+		if seenSeparator &&
+			((event.PCRIndex == internal_efi.SecureBootPolicyPCR &&
+				event.EventType == tcglog.EventTypeEFIVariableAuthority) ||
+				(event.PCRIndex == internal_efi.BootManagerCodePCR &&
+					event.EventType == tcglog.EventTypeEFIBootServicesApplication)) {
+			return nil
+		}
+
 		if event.PCRIndex != internal_efi.DriversAndAppsPCR {
 			continue
 		}
 
-		if event.EventType == tcglog.EventTypeSeparator {
-			return h.measureSeparator(ctx, internal_efi.DriversAndAppsPCR, event)
+		if !seenSeparator {
+			if event.EventType == tcglog.EventTypeSeparator {
+				if err := h.measureSeparator(ctx, internal_efi.DriversAndAppsPCR, event); err != nil {
+					return err
+				}
+				seenSeparator = true
+				continue
+			}
+			ctx.ExtendPCR(internal_efi.DriversAndAppsPCR, event.Digests[ctx.PCRAlg()])
+			continue
+		}
+
+		if !internal_efi.IsVendorEventType(event.EventType) {
+			return fmt.Errorf(
+				"unexpected post-separator event type %v found in PCR %d",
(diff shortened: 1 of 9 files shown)


ebrig:
@chrisccoulson @valentindavid, would either of you be able to review this when you have a chance?

frederic-hoerni:
Hi @ebrig , thank you for this PR.

Could you please send the BIOS measurement of your platform (`/sys/kernel/security/tpm0/binary_bios_measurements`), so that we can get the full picture?

Could you please also agree on the [Canonical contributor licence agreement](https://canonical.com/legal/contributors)?

ironhalik:
I implemented this PR on my AMD Thinkpad x13 g3. Works well.

ironhalik:
@ebrig could you continue this PR? Or would it be ok if I took over via a new PR?

ebrig:
Thanks, @frederic-hoerni. I’ve signed the Canonical contributor licence agreement.

Attached is a gzip-compressed copy of the requested raw `binary_bios_measurements` log from the HP ZBook Ultra G1a with BIOS X89 01.05.07 and the firmware configuration described in this PR.

[binary_bios_measurements.gz](https://github.com/user-attachments/files/31890307/binary_bios_measurements.gz)

SHA-256 of the uncompressed log:

`94b00512b68eae0cef26fe4c3f6aee036f66140a6687e4dd091b9b8efab0915f`

@ironhalik, thank you for testing this on the AMD ThinkPad X13 Gen 3. Yes, I’m continuing this PR, so there’s no need to open a replacement PR.

The PR is ready for continued review. Please let me know if any additional logs, tests, or changes would help.

pkit:
@ebrig looks better than my change. Works well on my hardware (Thinkpad P14s Gen 6 AMD)

pkit:
@frederic-hoerni @ebrig the vendor PCR2 event after separator is a `TSME: on/off` measurement in the new AMD AI cpus.
I disassembled the firmware on multiple AMD laptops to verify.

frederic-hoerni:
@ebrig, could you please rebase your branch on top of `master`?

valentindavid:
I feel like those should be detected not from reading the event log, but from the preinstall checks in `efi/preinstall/check_host_security_amd.go`. Then it should be verified it is present when expected from the checks of event logs.

Also do we know how the measurement of TSME  configuration looks like? It would be better to verify the value in the checks, and generate the expected value from the policy generation.

ebrig:
Addressed. The preinstall checks now derive the TSME state and HP DMA measurement configuration from sysfs, validate the corresponding event data and digests, and use those independently detected values for profile generation.

Linux does not currently expose the HSTI bit that indicates whether TSME measurement is supported, so the TSME event is validated when present, but its presence cannot be required. I also verified on the affected hardware that the generated PCR2 and PCR7 values match the live TPM.

PR #560, opened 2026-08-19
efi/preinstall: enable checks on arm64 platforms, and DGX Spark in particular

alexclewontin:
Opening a draft PR for initial feedback.

This is quite noisy, but a significant portion of the delta is just refactoring checks_test.go and checks_context_test.go to not be so Intel-hyperspecific (and even then, it's mostly just putting the body of each test into a per-fixture loop, so if you look at whitespace insensitive diffs it's much much smaller).

Interested in feedback on the approach at the platform level: the ARM ecosystem is a bit fragmented so I don't really see any way around the switch on vendor/switch on platform approach. Is DMI a sufficient source of info here?

NVIDIA will provide documentation on reading HW ROT and debugger fusing, which will be incorporated before this should be merged.

(first two things are tiny and things that Claude kept calling out so I figured I'd just throw them in as standalone commits)


efi/preinstall/check_host_security_nvidia.go
```diff
@@ -0,0 +1,54 @@
...
+
+package preinstall
+
+import (
+	"fmt"
+
+	internal_efi "github.com/snapcore/secboot/internal/efi"
+)
+
+const nvidiaDGXSparkProductName = "DGX Spark"
```
antlassagne: AI tells me:

> The production environment reads `/sys/class/dmi/id/product_name` and compares it exactly with `"DGX Spark"`. On the DGX Spark used for this review, that file contains `NVIDIA_DGX_Spark`. 

Is it possible that you based this on your device that may have a prototype modalias?

efi/preinstall/check_host_security_nvidia.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func checkHostSecurityNVIDIADGXSpark(env internal_efi.HostEnvironmentARM64) (platformFirmwareIntegrityConfig, error) {
+	// TODO: Implement proper HW ROT fusing checks, once we have the documentation
+	// from NVIDIA to do so. This will involve checking fuses and will return
+	// platformFirmwareIntegrityVerified if set correctly.
+
+	// TODO: Implement proper debug authentication checks, once we have the documentation
+	// from NVIDIA to do so.
+	return platformFirmwareIntegrityMeasured, nil
```
antlassagne: Should you raise a not implemented error instead of returning a success?

efi/preinstall/check_tpm_arm64.go
```diff
@@ -0,0 +1,107 @@
...
+	switch parent.Properties()["DRIVER"] {
+	case "optee-ftpm", "ftpm-tee":
+		return true, nil
+	default:
+		return false, nil
+	}
+}
+
+// isTPMDiscreteNvidia determines whether the default TPM is discrete on NVIDIA systems
+func isTPMDiscreteNvidia(env internal_efi.HostEnvironmentARM64) (bool, error) {
```
antlassagne: AI tells me:

> TPM classification can reject a TPM that was already opened successfully

> After `RunChecks` has opened and exercised the default TPM, the ARM64 reset-mitigation check independently searches sysfs for `DEVNAME=tpm0` and requires exactly one result with a readable parent. If udev enumeration returns zero or duplicate matches, cannot read the parent, or represents the opened TPM differently, the whole check now fails with `TPM2DeviceError` even though the TPM itself was usable.
> This second lookup is not correlated with the `TPMDevice` selected earlier. It should either classify the already-selected device or treat unavailable driver metadata as an inability to determine the optional reset mitigation, rather than as a fatal TPM-device failure. Tests should cover zero/multiple matches and parent/enumeration errors through `RunChecks`, not only the direct happy paths.

efi/preinstall/check_tpm_arm64.go
```diff
@@ -0,0 +1,107 @@
...
+	case "NVIDIA":
+		return isTPMDiscreteNvidia(arm64Env)
+	default:
+		return false, &UnsupportedPlatformError{fmt.Errorf("unsupported system vendor: %s", systemVendor)}
+	}
+}
+
+// isTPMFirmwareOptee determines whether the default TPM is an OP-TEE firmware TPM,
+// as identified by its backing kernel driver
+func isTPMFirmwareOptee(env internal_efi.HostEnvironment) (bool, error) {
```
antlassagne: In the spirit of keeping things minimal, do we want to handle optee TPMs in this enablement?

efi/preinstall/check_host_security_arm64.go
```diff
@@ -0,0 +1,108 @@
...
+
+	if err := checkSecureBootPolicyPCRForDegradedFirmwareSettings(log); err != nil {
+		var ce CompoundError
+		if !errors.As(err, &ce) {
+			return platformFirmwareIntegrityNone, fmt.Errorf("encountered an error whilst checking the TCG log for degraded firmware settings: %w", err)
+		}
+		errs = append(errs, ce.Unwrap()...)
+	}
+
+	if err := checkForKernelIOMMU(env); err != nil {
```
antlassagne: AI tells me:

> The new generic ARM64 check treats the existence of any device in the `iommu` subsystem as sufficient DMA protection. This DGX Spark exposes three SMMUs under `/sys/class/iommu`, so that check succeeds, but the boot log reports `iommu: Default domain type: Passthrough`. SMMU availability therefore does not establish that devices, particularly externally accessible PCIe devices, are actually attached to translated/protected domains.

> Before using this as the ARM64 DMA-security gate, the implementation should verify the relevant devices' active IOMMU domains or otherwise establish that passthrough cannot leave attacker-accessible devices unprotected.

efi/preinstall/check_tpm_arm64.go
antlassagne: However, when checking the DGX Spark, we are indeed in the situation of the happy path.

antlassagne (commented):
First piece of review

efi/preinstall/check_host_security_nvidia.go
antlassagne: Also: https://github.com/ubuntu/ubuntu-desktop-documentation/pull/169

efi/preinstall/check_host_security_nvidia.go
alexclewontin: It is possible. We should poll our lab estate to figure out what the right ID is. But I am also open to other suggestions for mechanisms here (could we do something more SoC generic, so that it works on any GB10?)

efi/preinstall/check_tpm_arm64.go
alexclewontin: It is a fair question. The only reason I did include it is because it is as close to a non-fragmented, cross-platform API as I could come up with on ARM (I for a while tried to include CoreSight detection in a crossplatform way but determined it was not meaningfully possible)

efi/preinstall/check_host_security_arm64.go
alexclewontin: Yeah AI flagged this for me too. I *think* a similar gap exists on the AMD64 side, but it is possible that presence of an IOMMU is a weaker guarantee on ARM64 than it is on AMD64? I'll have to dive into that a bit

efi/preinstall/check_host_security_nvidia.go
alexclewontin: This is just a placeholder, and exactly the thing that should change before this gets merged

efi/preinstall/check_host_security_arm64.go
alexclewontin: So I think in the normal case, the guarantees are basically the same: per [the comments above the function definition](https://github.com/canonical/secboot/blob/4622187ba243e490514da37728e699b0805a00b6/efi/preinstall/check_host_security.go#L71) we are not 100% sure that iommu device existence is sufficient, so those guarantees may not be sufficient, but I don't think they are inherently weaker on ARM.

On ARM64, there *are* a couple of extra escape hatches (`CONFIG_ARM_SMMU_DISABLE_BYPASS_BY_DEFAULT=false`, `arm-smmu.disable-bypass=0`). We should ideally not be shipping kernels with the first. Core has a way to put controls on kcmdline arguments, not sure if desktop w FDE has the same. If it does we would deny the second.

I agree that there is danger in allowing iommu passthrough as a default domain type (via kconfig or kcmdline)... but I believe that danger is arch agnostic, and also can change per kernel and even per boot, so checks on the current boot, often in the context of the installer, should not be considered authoritative.

 However it is interesting that your Spark has `iommu: Default domain type: Passthrough`. I checked mine, and I have `iommu.passthrough=0` on the command line and further down the boot log

```
Aug 12 18:08:24 apollo kernel: iommu: Default domain type: Translated (set via kernel command line)
Aug 12 18:08:24 apollo kernel: iommu: DMA domain TLB invalidation policy: lazy mode
```

We should get to the bottom of that disparity

alexclewontin: However the -nvidia kernel looks like it is configured with
```
# CONFIG_ARM_SMMU_DISABLE_BYPASS_BY_DEFAULT is not set
# CONFIG_IOMMU_DEFAULT_DMA_STRICT is not set
# CONFIG_IOMMU_DEFAULT_DMA_LAZY is not set
CONFIG_IOMMU_DEFAULT_PASSTHROUGH=y
```
:scream:

efi/preinstall/check_host_security_nvidia.go
MitchellAugustin: I'm seeing `NVIDIA_DGX_Spark` on my home Spark's resolute install (and its Noble FastOS install), as well as 202511-38080, 202511-38082, 202511-38083, 202511-38084 in the lab (some on RR, some on NN). (these are our control unit and release quality devices).

I suspect `DGX Spark` was just used in the earlier samples as well.

efi/preinstall/check_host_security_arm64.go
MitchellAugustin: @alexclewontin FWIW I also have
```
mitchell@spark-prod:~$ sudo dmesg | grep "iommu" | grep "Passthrough"
[    0.105847] iommu: Default domain type: Passthrough
```
on my Spark's stock Resolute install. 

On its Noble-based FastOS partition though, I have the same as you with `iommu.passthrough=0` being set (via `/etc/default/grub.d/iommu.cfg`).

There are a handful of other grub cmdline differences that NVIDIA ships in FastOS (and I don't think we include every single one in dgx-desktop-defaults) - we will probably just need to keep a close eye on these and make sure that anything that needs to be built into the snapped kernel image for FastOS to work does end up there.

efi/preinstall/check_host_security_arm64.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return dtpmPartialResetAttackMitigationUnknown, &TPM2DeviceError{err}
+	}
+	if !discreteTPM {
+		return dtpmPartialResetAttackMitigationNotRequired, nil
+	}
+
+	// ARM64 has no generic mechanism to establish that the TPM startup locality
+	// is protected by the hardware root of trust, so PCR0 binding cannot be relied
+	// on to mitigate an independent reset of a discrete TPM.
```
MitchellAugustin: The claim that there's no generic mechanism for this seems essentially correct based on my read of [the spec](https://trustedcomputinggroup.org/wp-content/uploads/Trusted-Platform-Module-2.0-Library-Part-1-Architecture_Version-185_pub.pdf).

However, did we confirm that this definitely isn't relevant for Spark platforms _specifically_? (i.e., do we know for sure whether or not the startup locality is available to the OS? (since it seems like that's determining factor of whether this mitigation is enabled in the amd64 implementation)?)

efi/preinstall/check_tpm_arm64.go
```diff
@@ -0,0 +1,107 @@
...
+	parent, err := devices[0].Parent()
+	if err != nil {
+		return false, fmt.Errorf("cannot obtain parent of tpm0 device: %w", err)
+	}
+	if parent == nil {
+		return false, fmt.Errorf("internal error: tpm0 device has no parent")
+	}
+
+	switch parent.Properties()["DRIVER"] {
+	case "optee-ftpm", "ftpm-tee":
```
MitchellAugustin: Looks right based on [this](https://github.com/torvalds/linux/blob/cb8a75eec0877810b50aa1c5a833f929525cd2ee/drivers/char/tpm/tpm_ftpm_tee.c#L332) and [this](https://github.com/torvalds/linux/blob/cb8a75eec0877810b50aa1c5a833f929525cd2ee/drivers/char/tpm/tpm_ftpm_tee.c#L354), but I don't have the hardware to check myself

efi/preinstall/check_tpm_arm64.go
```diff
@@ -0,0 +1,107 @@
...
+func isTPMDiscreteNvidia(env internal_efi.HostEnvironmentARM64) (bool, error) {
+	productName, err := env.ProductName()
+	if err != nil {
+		return false, &UnsupportedPlatformError{fmt.Errorf("cannot determine product name: %w", err)}
+	}
+
+	switch productName {
+	// We just happen to know that the NVIDIA DGX Spark has a dTPM
+	case nvidiaDGXSparkProductName:
+		return true, nil
```
MitchellAugustin: I suppose this is fine for now (as there is already necessarily quite a lot of platform-specific hardcoding needed for various things across secboot), but I wonder if there isn't some other more generic check we could do for this, akin to the fTPM checks.

(although, on the other hand, I guess that may not add any real value for this or other platforms, given all the platform-specific quirks.)

internal/efi/default_env_arm64.go
```diff
@@ -0,0 +1,63 @@
...
+package efi
+
+import (
+	"fmt"
+	"strings"
+)
+
+var (
+	dmiSystemVendorPath = "/sys/class/dmi/id/sys_vendor"
+	dmiProductNamePath  = "/sys/class/dmi/id/product_name"
```
MitchellAugustin: I believe this is fine for the desktop DGX Spark and OEM flavors, but it might not be universal across RTX Spark OEM flavors. (I know that at least "family name" isn't, and Scott has been telling us to use `SMBIOS Type 4 table for CPU Name starts with "NVIDIA RTX Spark"` as the criterion for when to install our default packages.)

I'll check with him to see if product_name will be standardized, but I expect it won't be.

MitchellAugustin (changes requested):
Looks mostly good, and most of the boilerplate looks analogous to the existing amd64 TPM device implementations.
I do have some questions and one important suggested change inline around your usage of `/sys/class/dmi/id/product_name` as a device identifier.

I did not thoroughly review the tests yet (assuming it may be best for me to do that after the implementation has stabilized), and will need to come back after the stubs are implemented to review those.

internal/efi/default_env_arm64.go
MitchellAugustin: See MM conversation

MitchellAugustin (changes requested):

efi/preinstall/check_tpm_arm64.go
alexclewontin: Not that I could tell and I looked quite a bit. Even on the AMD64 side the implementation of the equivalent appears to be split between AMD and Intel. On ARM, I would assume that anything that isn't standardized by EFI is mega-fragmented. 

The OP-TEE fTPM check is a nice short circuit heuristic we can use based on the fact that we know it's a common implementation. But I would not be shocked if it missed an equivalent implemented on top of Qualcomm's QTEE. And you can't do the opposite: the CRB driver that the Spark uses can also be used by fTPMs.

And yeah, at the moment the ARM ecosystem is still in a place where we'd have to add HW ROT fusing checks per platform, so it doesn't really matter. TBF AMD64 doesn't look much better, but the fact that it's just two vendors instead of a half dozen makes it a little easier I suppose

efi/preinstall/check_host_security_nvidia.go
alexclewontin: Closing this because we probably want to use a slightly different field based to anticipate compatibility, with future similar platforms.

internal/efi/default_env_arm64.go
antlassagne: The only thing that stays the same accross the entire DGX Spark family s product_family.

Here are examples of the modalias for different oem devices. I collected these before kernel 6.19 so the product family was not in there yet. I should definitely have asked for the entire dmidecode output at the time...

```
# prototypes - the one alex has
dmi:bvnAmericanMegatrendsInternational,LLC.:bvr5.36_0ACUM021:bd09/26/2025:br5.36:svnNVIDIA:pnDGXSpark:pvr5.36_0ACUM021:rvnNVIDIA:rnDGXSpark:rvr5.36_0ACUM021:cvnDefaultstring:ct3:cvrDefaultstring:skuDefaultstring:

# founder edition
dmi:bvnAmericanMegatrendsInternational,LLC.:bvr5.36_0ACUM018:bd08/06/2025:br5.36:svnNVIDIA:pnNVIDIA_DGX_Spark:pvrA.7:rvnNVIDIA:rnP4242:rvrA04:cvnNVIDIA:ct17:cvrA04:sku0000:

# lenovo
dmi:bvnLENOVO:bvr S0QKTOAA:bd10/21/2025:br1.10:efr2.66:svnLENOVO:pn30KL0005
GF:pvrThinkStationPGX:rvnLENOVO:rnINVALID:rvrNotDefined:cvnLENOVO:ct3:cvrN
one:skuLENOVO_MT_30KL_BU_Think_FM_ThinkStationPGX:

# dell
dmi:bvnDellInc.:bvr5.36_1.1.1:bd08/15/2025:br5.36:svnDellInc.:pnDellProMaxwithGB10FCM1253:pvr:rvnDellInc.:rn038RY7:rvrA00:cvnDellInc.:ct3:cvr:sku0DA7:

# asus
dmi:bvnAmericanMegatrendsInternational,LLC.:bvrGX10DGX.0102.2025.1111.1531:bd11/11/2025:br5.36:svnASUSTeKCOMPUTERINC.:pnGX10:pvr5.36_GX10DGX:rvnASUSTeKCOMPUTERINC.:rnGX10:rvrDefaultstring:cvnASUSTeKCOMPUTERINC.:ct17:cvrDefaultstring:skuDefaultstring:
```

antlassagne: So after digging, if this is about identifying the manufacturer to stay aligned with the x86 implementation
- x86 relies on /proc/cpuinfo vendor_id for that identification
- we can't because arm does not expose it. However we can identify the CPU in several ways, for example
    - Cortex-A725 / Cortex X925 with lscpu
    - The manufacturer of the processor with `sudo dmidecode -t 4` and a bit of parsing. I confirm the manufacturer is NVIDIA for Spark's founder edition and for the laptops
I'd go with option 2.

However, in the end, that's simply to determine if the TPM is discrete or not. I see that you hardcode the fact that DGX Spark has a discrete TPM. For that, I would use the product family.

aglinserer: Instead of `sudo dmidecode -t 4` I'd suggest using `dmidecode --string processor-version` or the likes to avoid having to parse potentially differently structured output.

efi/preinstall/check_host_security_arm64.go
```diff
@@ -0,0 +1,108 @@
...
+	switch systemVendor {
+	case "NVIDIA":
+		return checkHostSecurityNVIDIA(env)
+	default:
+		return platformFirmwareIntegrityNone, &UnsupportedPlatformError{fmt.Errorf("unsupported system vendor: %s", systemVendor)}
+	}
+}
+
+// checkHostSecurity is the main entry point for verifying that the host security
+// is sufficient. Errors that can't be resolved or which should prevent further checks from running
```
antlassagne: what is sufficient?

efi/preinstall/check_host_security_arm64.go
```diff
@@ -0,0 +1,108 @@
...
+		return platformFirmwareIntegrityNone, err
+	}
+
+	return checkHostSecurityARM64Generic(env, log, integrity)
+}
+
+func checkHostSecurityARM64Generic(env internal_efi.HostEnvironment, log *tcglog.Log, integrity platformFirmwareIntegrityConfig) (platformFirmwareIntegrityConfig, error) {
+	var errs []error
+
+	if err := checkSecureBootPolicyPCRForDegradedFirmwareSettings(log); err != nil {
```
antlassagne: That would need a test against what's expected on DGX Spark ideally

efi/preinstall/check_host_security_nvidia.go
```diff
@@ -0,0 +1,54 @@
...
+
+const nvidiaDGXSparkProductName = "DGX Spark"
+
+func checkHostSecurityNVIDIA(env internal_efi.HostEnvironmentARM64) (platformFirmwareIntegrityConfig, error) {
+	productName, err := env.ProductName()
+	if err != nil {
+		return platformFirmwareIntegrityNone, &UnsupportedPlatformError{fmt.Errorf("cannot determine platform product name: %w", err)}
+	}
+
+	switch productName {
```
antlassagne: See other comments, the right identifier for dgx spark & family is product_family. The problem is that it might not be the right identifier for other platforms, but we don't have to solve every future problem right now

efi/preinstall/check_host_security_nvidia.go
```diff
@@ -0,0 +1,54 @@
...
+	case nvidiaDGXSparkProductName:
+		return checkHostSecurityNVIDIADGXSpark(env)
+	default:
+		return platformFirmwareIntegrityNone, &UnsupportedPlatformError{fmt.Errorf("unsupported NVIDIA product: %s", productName)}
+	}
+}
+
+func checkHostSecurityNVIDIADGXSpark(env internal_efi.HostEnvironmentARM64) (platformFirmwareIntegrityConfig, error) {
+	// TODO: Implement proper HW ROT fusing checks, once we have the documentation
```
antlassagne, line 54: Placeholder to keep a thread open until these TODOs are filled

efi/preinstall/check_tpm_arm64.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// isTPMDiscreteNvidia determines whether the default TPM is discrete on NVIDIA systems
+func isTPMDiscreteNvidia(env internal_efi.HostEnvironmentARM64) (bool, error) {
+	productName, err := env.ProductName()
+	if err != nil {
+		return false, &UnsupportedPlatformError{fmt.Errorf("cannot determine product name: %w", err)}
+	}
+
+	switch productName {
```
antlassagne: Again, we can't rely on the product name for DGX Spark

efi/preinstall/checks_context_test.go
```diff
@@ -1,7 +1,7 @@
 // -*- Mode: Go; indent-tabs-mode: t -*-
```
antlassagne, line 1: This file currently is +4829 -5962
Maybe it's expected, but for the sake of having this branch accepted faster by the maintainers is it possible to make the diff smaller?

internal/efi/env.go
```diff
@@ -83,6 +83,16 @@ type HostEnvironmentAMD64 interface {
 	ReadMSRs(msr uint32) (map[uint32]uint64, error)
 }
 
+// HostEnvironmentARM64 is an interface that abstracts out a host environment specific
+// to ARM64 platforms.
+type HostEnvironmentARM64 interface {
```
antlassagne, line 98: One more occurrence of the open question about identifying the devices

internal/efi/default_env_arm64_null.go
```diff
@@ -0,0 +1,27 @@
+//go:build !arm64
```
antlassagne: I'm a bit confused by this file. from what I understand, i file finishing with `*_arm64.go` will only build on arm64. In this repo they work with this mecanism, and add `*_null.go` files for the "default" path. But how does a `*_arm64_null.go` make sense?

efi/preinstall/export_test.go
```diff
@@ -87,7 +87,6 @@ var (
 	NewX509CertificateID                                  = newX509CertificateID
 	OpenAndCheckTPM2Device                                = openAndCheckTPM2Device
 	ReadOrderedLoadOptionVariables                        = readOrderedLoadOptionVariables
-	RestrictedTPMLocalitiesIntel                          = restrictedTPMLocalitiesIntel
```
antlassagne, line 90: Is that on purpose?

frederic-hoerni:
I would prefer if the refactoring bits were in separate commits. This would make reviewing easier.

antlassagne (changes requested):
Here is a deeper review. Overall my understanding is
- It brings up the whole arm64 checks structure, which makes it a bit verbose
- The actual path that matters is 
    - "is it arm? yes --> is it nvidia? yes --> is it DGX Spark? --> yes--> TPM is discrete.
    - Then the TCG event log is analyzed
    - Then we skip HW ROT fusing checks

There is definitely something fishy in the 2 test files that has some many changes.

efi/preinstall/export_test.go
alexclewontin: Yes, this is only used `check_host_security_intel_test.go` which gains a `//go:build amd64` tag in the same commit, so this moves to `export_amd64_test.go`

internal/efi/default_env_arm64_null.go
alexclewontin: This is the alternative, empty, stub implementation of the `HostEnvironment.ARM64()` method for non-ARM64 architectures. 

[`defaultEnvImpl` ](https://github.com/alexclewontin/secboot/blob/722fa4cc7f3530c76cba704e4862e3aa5a8bf064/internal/efi/default_env.go#L81) implements the [`HostEnvironment`](https://github.com/alexclewontin/secboot/blob/722fa4cc7f3530c76cba704e4862e3aa5a8bf064/internal/efi/env.go#L141) interface, which as of this PR has

- an `AMD64()` method returning a `HostEnvironmentAMD64, error` pair,
- an `ARM64()` method returning a `HostEnvironmentARM64, error` pair

You should only call `AMD64()` on amd64 architectures, and `ARM64()` on arm64 architectures. If you call the wrong one, you get an `ErrNot<arch>Host`, but you need to define and implement each for any struct that implements the `HostEnvironment` interface.

On AMD64, the (real meaningful) `HostEnvironment.AMD64` implementation comes from [`internal/efi/default_env_amd64.go`](https://github.com/alexclewontin/secboot/blob/722fa4cc7f3530c76cba704e4862e3aa5a8bf064/internal/efi/default_env_amd64.go#L114). The (empty stub) `HostEnvironment.ARM64` implementation comes from [`internal/efi/default_env_arm64_null.go`](https://github.com/alexclewontin/secboot/blob/722fa4cc7f3530c76cba704e4862e3aa5a8bf064/internal/efi/default_env_arm64_null.go#L24)

On ARM64, you get the inverse. The (empty stub) `HostEnvironment.AMD64` implementation comes from [`internal/efi/default_env_amd64_null.go`](https://github.com/alexclewontin/secboot/blob/722fa4cc7f3530c76cba704e4862e3aa5a8bf064/internal/efi/default_env_amd64_null.go#L24), and the (real meaningful) `HostEnvironment.ARM64` implementation comes from [`internal/efi/default_env_arm64.go`](https://github.com/alexclewontin/secboot/blob/722fa4cc7f3530c76cba704e4862e3aa5a8bf064/internal/efi/default_env_arm64.go#L60)

On a theoretical third architecture, you'd get empty stub implementations of both `HostEnvironment.AMD64` and `HostEnvironment.ARM64`.

I agree that it is a little bit awkward. We are relying on build constraints and runtime errors to enforce something that would ideally be enforced by a type system. But the issue is that it is hard to come up with an abstraction that is both sufficiently abstract so as to work for both architectures, but also concrete enough to not be meaningless and just push this down a layer. And also I don't want to introduce more churn into this PR

antlassagne: So, it's not possible to simply have one _amd64.go, one _arm64.go, and one _null.go file? that would be cleaner and remove complexity from this PR?

efi/preinstall/checks_context_test.go
alexclewontin: I am going to look at @frederic-hoerni's comment about breaking up the commits a bit. Basically, the churn in this and `checks_test.go` come from taking all of the tests from

```
func Test() {
   const fixtureData = foo
   doAThing(fixtureData)
   doASecondThing(fixtureData)
}
```
to
```
fixtures := [foo, bar, baz]

func Test() {
    for fixtureData in fixtures {
        doAThing(fixtureData)
        doASecondThing(fixtureData)
    }
}
```
The issue is essentially that in this example, fixture data is a bunch of *super* intel-specific tests that wouldn't meaningfully exercise the arm64 codepaths. We could keep them how they are, but then we would have to add a bunch of redundant new code testing the arm64 codepaths. This way, we can piggyback on the logic encoded in the `doAThing(), doASecondThing()` sequence and apply it to any arbitrary test fixture

alexclewontin: Also FWIW if you ignore whitespace-only changes:

```
$ git diff -w --stat 02f4ce691cbebc106ca3485ecd75f16aaec510da^!
 efi/preinstall/check_tpm_intel_test.go                                         |    4 +-
 efi/preinstall/checks_amd64_test.go                                            |  107 +++++
 efi/preinstall/checks_context_amd64_test.go                                    |  107 +++++
 efi/preinstall/checks_context_test.go                                          | 2463 ++++++++++++++++++++++++++++---------------------------------------------------------------------------
 efi/preinstall/checks_fixture_amd64_test.go                                    |  112 +++++
 efi/preinstall/checks_fixture_null_test.go                                     |   45 ++
 efi/preinstall/checks_fixture_test.go                                          |   96 ++++
 efi/preinstall/checks_test.go                                                  | 1573 ++++++++++++++++++------------------------------------------------
 efi/preinstall/export_amd64_test.go                                            |    3 +-
 efi/preinstall/export_test.go                                                  |    3 +-
 internal/efi/{default_env_null.go => default_env_amd64_null.go}                |    2 +-
 internal/efi/{default_env_not_amd64_test.go => default_env_amd64_null_test.go} |    2 +-
 12 files changed, 1553 insertions(+), 2964 deletions(-)
 ```
 Still big, but not as bad

File: internal/efi/default_env_arm64_null.go (hunk 17 above)
Comment (alexclewontin) [reply internal/efi/default_env_arm64_null.go]: You could condense down to just `_amd64.go` and `_arm64.go` (no `_null.go`) by putting the `AMD64()` stub in the `_arm64.go` file, and the `ARM64()` stub in the `_amd64.go` file, but that breaks once you want to add a third arch like RISC-V or something (each file needs to gain a stub implementation for a `RISCV()` method and the `_riscv.go` file needs to re-implement the stub `AMD64()` and `ARM64()` methods)

Comment (alexclewontin) [reply internal/efi/default_env_arm64_null.go]: But open to it if that would truly be better

File: efi/preinstall/check_tpm_arm64.go (hunk 14 above)
Comment (alexclewontin) [reply efi/preinstall/check_tpm_arm64.go]: Moved to DMI type 4

File: efi/preinstall/check_host_security_intel.go [hunk 19]
```diff
@@ -346,7 +343,7 @@ const (
 func checkHostSecurityIntelCPUDebuggingLocked(env internal_efi.HostEnvironmentAMD64) error {
 	// Check for "Silicon Debug Interface", returned in bit 11 of %ecx when calling
 	// cpuid with %eax=1.
-	debugSupported := env.HasCPUIDFeature(cpuid.SDBG)
+	debugSupported := env.HasCPUIDFeature(internal_efi.CPUIDFeatureSDBG)
```
MitchellAugustin, line 346: What's the reason behind changing these intel checks?

internal/efi/default_env.go
```diff
@@ -240,3 +249,92 @@ func (defaultEnvImpl) EnumerateDevices(matcher netlink.Matcher) ([]SysfsDevice,
...
+// offset within the formatted area that holds the 1-based string index.
+//
+// Layout (DMTF SMBIOS specification):
+//   - byte 0:   structure type (must be 4)
+//   - byte 1:   length of the formatted area including the header
+//   - bytes 2-3: handle
+//   - bytes 4…: remaining formatted area
+//   - after the formatted area: NUL-terminated strings; string set ends
+//     with an additional NUL (empty string sentinel)
+func decodeSMBIOSType4Field(data []byte, fieldOffset uint8) (string, error) {
```
MitchellAugustin, line 275: This function LGTM, I copied this function and ran it on both devices, and it worked as intended on both. Also LGTM when comparing with the table spec as described on [page 43 here](https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.9.0.pdf).

(I recall you mentioning a back and forth between @antlassagne and @aglinserer on the feasibility of this but could not find where that conversation was)

efi/preinstall/check_host_security_nvidia.go
```diff
@@ -0,0 +1,61 @@
...
+
+const nvidiaDGXSparkCPUVersion = "GB10"
+const nvidiaRTXSparkCPUVersionPrefix = "NVIDIA RTX Spark"
+
+// isNvidiaSparkCPUVersion reports whether cpuVersion identifies a supported NVIDIA
+// Spark platform: either the DGX Spark (exact match on "GB10") or any RTX Spark
+// variant (prefix match on "NVIDIA RTX Spark").
+func isNvidiaSparkCPUVersion(cpuVersion string) bool {
+	return cpuVersion == nvidiaDGXSparkCPUVersion || strings.HasPrefix(cpuVersion, nvidiaRTXSparkCPUVersionPrefix)
+}
```
MitchellAugustin, line 37: These match criteria LGTM

MitchellAugustin (commented):
Platform matching criteria LGTM. No immediate concerns with the rest, but are the deleted _amd64 files intentional here?

Deferring my review of the tests again until we have the HW ROT fusing items + any other known outstanding bits implemented

efi/preinstall/check_host_security_intel.go
alexclewontin: the cpuid package is amd64 only. In the spirit of truly decoupling test host arch from mocked arch under test, this lets you run the intel checks on an arm64 test host, by mocking the environment.

We maintain a parallel copy of the SDBG constant in the internal_efi package  ( :( ) but a) these are HW features, they should never change, and b) we have tests (that do only run on amd64) that just serve as guardrails to make sure that our constant == cpuid package's constant, so if somehow something did go wrong tests would start failing.

alexclewontin: also this has been split, so this particular change now lives in canonical/secboot#563

PR #563, opened 2026-08-31
efi: allow other architectures in check_host_security and friends

alexclewontin:
Checking host security today makes assumes in many ways secboot will only be used on amd64. This PR attempts to lay the groundwork to allow other architectures to leverage much of the same logic.

Big themes:
 - removing architecture-conditional compilation, to allow unit tests to be mocked and run on any host architecture
 - abstracting the mocked test environment (the "fixture") from the actual code under test. This accounts for the bulk of the PR, as a ton of a couple test files now sit inside loops over fixtures. The whitespace insensitive diff is much smaller.
 - adding switch/case statements to the business logic of checking host security, so that each architecture can implement its own ecosystem-aware checks (as even within amd64 there doesn't seem to be a platform agnostic way to handle this).
 
 
This is based on top of https://github.com/canonical/secboot/pull/562 and is part of an attempt to split up https://github.com/canonical/secboot/pull/560.


efi/preinstall/check_tpm_intel_test.go
```diff
@@ -23,10 +23,18 @@ import (
...
 
-type tpmIntelSuite struct{}
+type tpmIntelSuite struct {
+	snapd_testutil.BaseTest
+}
+
+func (s *tpmIntelSuite) SetUpTest(c *C) {
+	s.BaseTest.SetUpTest(c)
+	s.AddCleanup(MockRuntimeGOARCH("amd64"))
+}
```
frederic-hoerni, line 38: Why is this modification needed?

efi/preinstall/checks_fixture_test.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	devices := append([]internal_efi.SysfsDevice(nil), f.devices...)
+	base = append(base, efitest.WithSysfsDevices(devices...))
+	return efitest.NewMockHostEnvironmentWithOpts(append(base, options...)...)
+}
+
+func (f *runChecksHostFixture) mockRuntimeGOARCH(s interface{ AddCleanup(func()) }) {
+	s.AddCleanup(MockRuntimeGOARCH(f.arch))
+}
+func runChecksPlatformHostFixtures() []runChecksHostFixture {
```
frederic-hoerni, line 121: - please add a description about what this function returns and how it is supposed to be maintained
- missing new line before func

efi/preinstall/checks_fixture_test.go
```diff
@@ -0,0 +1,196 @@
...
+// Platform identity capabilities do not imply security properties. In
+// particular, HasIntelBootGuard does not imply that any TPM startup locality is
+// inaccessible from the OS; fixtures must declare locality properties
+// separately.
+
+// runChecksHostFixture abstracts host-specific details from scenario logic in
+// the shared RunChecks and RunChecksContext tests.
+type runChecksHostFixture struct {
+	name                    string
+	capabilities            runChecksHostCapabilities
```
frederic-hoerni, line 68: If I correctly understand, `capabilities` summarizes the fields below, to make it simpler to select a given set of host fixtures (in `runChecksHostFixturesFor`).

Maybe you could add a comment about that.

efi/preinstall/checks_fixture_test.go
```diff
@@ -0,0 +1,196 @@
...
+	environment             efitest.MockHostEnvironmentOption
+	virtualizationMode      string
+	virtualizationDetection internal_efi.DetectVirtMode
+	devices                 []internal_efi.SysfsDevice
```
frederic-hoerni: Is it worth it to set a more specific name (eg: `sysfsDevices`)? (because we already have "devices" everywhere...)

efi/preinstall/checks_fixture_test.go
```diff
@@ -0,0 +1,196 @@
...
+	base = append(base, efitest.WithSysfsDevices(devices...))
+	return efitest.NewMockHostEnvironmentWithOpts(append(base, options...)...)
+}
+
+func (f *runChecksHostFixture) mockRuntimeGOARCH(s interface{ AddCleanup(func()) }) {
+	s.AddCleanup(MockRuntimeGOARCH(f.arch))
+}
+func runChecksPlatformHostFixtures() []runChecksHostFixture {
+	intelDevices := func(status []byte, withIOMMU bool) []internal_efi.SysfsDevice {
+		attrs := map[string][]byte{
```
frederic-hoerni, line 123: Instead of embedding a func here, wouldn't it be more legible to have it as a plain regular named function? That would also be easier to document and let `runChecksPlatformHostFixtures()` focus on the list that it returns?

efi/preinstall/checks_context_test.go
```diff
@@ -90,6 +89,7 @@ type testRunChecksContextRunParams struct {
 }
 
 func (s *runChecksContextSuite) testRun(c *C, params *testRunChecksContextRunParams) (errs []*WithKindAndActionsError) {
+	s.ResetAndClearTPMSimulatorUsingPlatformHierarchy(c)
```
frederic-hoerni: Why is this needed?

efi/preinstall/checks_context_test.go
```diff
@@ -215,785 +215,632 @@ func (s *runChecksContextSuite) testRun(c *C, params *testRunChecksContextRunPar
...
-		loadedImages:              efiImagesDefault(),
-		profileOpts:               PCRProfileOptionsDefault,
-		actions:                   []actionAndArgs{{action: ActionNone}},
-		expectedPcrAlg:            tpm2.HashAlgorithmSHA256,
-		expectedUsedSecureBootCAs: []*X509CertificateID{NewX509CertificateID(testutil.ParseCertificate(c, msUefiCACert))},
-		expectedFlags:             NoPlatformConfigProfileSupport | NoDriversAndAppsConfigProfileSupport | NoBootManagerConfigProfileSupport,
-		expectedWarningsMatch: `3 errors detected:
+	required := runChecksHostCapabilityValid | runChecksHostCapabilityFirmwareTPM
+	for _, fixture := range runChecksHostFixturesFor(c, required) {
+		c.Logf("running with host fixture %q", fixture.name)
```
frederic-hoerni, line 220: These are worth being displayed when tests are run (for identifying which host fixture raised an error, or for comparison of which host fixtures get tested from a version to another...).

They appear with `-check.vv`, but this option also activates other cumbersome log messages.

So, is there a way to give better visibility when running tests? And possibly ensure that `Makefile` and `.github/workflows/test.yaml` (or `run-tests`) take advantage of it.

efi/preinstall/check_tpm_intel_test.go
alexclewontin: Without it, `runtimeGOARCH` ([check_host_security.go:36](https://github.com/alexclewontin/secboot/blob/3daef882a516e145fce04b879a63df0b0b5f8b12/efi/preinstall/check_host_security.go#L36)) will take the value of whatever the test runner arch is. This will cause failures in all the tests in the file on non-amd64 runners since they all call `env.AMD64` which will fail at runtime if called on a non-amd64 device.

alexclewontin: We could call it on a per-test basis? But this is a universal need for Intel TPM tests so we do it once for the whole suite, and then tear it down again at the end

efi/preinstall/checks_fixture_test.go
alexclewontin: Done

efi/preinstall/checks_fixture_test.go
alexclewontin: Both of these should be addressed, let me know how it looks

efi/preinstall/checks_fixture_test.go
alexclewontin: Added, let me know if this covers it

internal/cpuid/cpuid_null.go
alexclewontin, line 1: This is not a rename, it's an addition of a file & the removal of a file but because both files are small and the license header dominates, it's something like 52% similar so detected as a rename

PR #564, opened 2026-08-31
efi: implement platform-agnostic arm64 host security checks

alexclewontin:
This PR adds a framework for host security checks on arm64 platforms. It builds on https://github.com/canonical/secboot/pull/563 (which represents most of the diff), and is part of an attempt to split up https://github.com/canonical/secboot/pull/560.

It adds a generic framework to check host security on ARM64. Very few truly cross-platform APIs were found to be applicable to that ecosystem, but we can use the backing driver for the TPM to prove that a given TPM is a fTPM: if it is using the OP-TEE fTPM driver we can be sure that it is an fTPM. This is not conclusive (absence of that driver doesn't prove that a TPM is a dTPM), but it is a common reference implementation for the ARM ecosystem, so worth including.


efi/preinstall/check_host_security.go
```diff
@@ -145,3 +153,218 @@ Loop:
...
+	// The startup locality is available to the OS, so the mitigation
+	// is unavailable even though it would have been desired because
+	// PCR0 can be recreated from the OS.
+	return dtpmPartialResetAttackMitigationUnavailable, nil
+}
+
+// checkHostSecurityARM64Platform selects the platform-specific firmware
+// integrity check. Tests replace this to supply synthetic platforms.
+var checkHostSecurityARM64Platform = func(env internal_efi.HostEnvironmentARM64, cpuManufacturer string) (platformFirmwareIntegrityConfig, error) {
+	return platformFirmwareIntegrityNone, &UnsupportedPlatformError{fmt.Errorf("unsupported CPU manufacturer: %s", cpuManufacturer)}
```
MitchellAugustin, line 306: I guess this is intentionally a stub until we have our updates from NVIDIA?

MitchellAugustin (approved):
What has been implemented here so far looks good from my perspective. Main thoughts:

- In my comparison of this PR's diff against https://github.com/canonical/secboot/pull/560's diff, none of the new delta seems problematic to me
- SMBIOS table 4 detection and platform-level items discussed in 560 have been cleanly included in this PR

This was only a static code review (I did not test)

efi/preinstall/check_host_security.go
alexclewontin: yes, although this could be a stub that includes a switch case statement

PR #565, opened 2026-09-04
efi: allow other architectures in check_host_security and friends

alexclewontin:
This is the continuation of #563 as part of a stack

Checking host security today makes assumes in many ways secboot will only be used on amd64. This PR attempts to lay the groundwork to allow other architectures to leverage much of the same logic.

Big themes:

removing architecture-conditional compilation, to allow unit tests to be mocked and run on any host architecture
abstracting the mocked test environment (the "fixture") from the actual code under test. This accounts for the bulk of the PR, as a ton of a couple test files now sit inside loops over fixtures. The whitespace insensitive diff is much smaller.
adding switch/case statements to the business logic of checking host security, so that each architecture can implement its own ecosystem-aware checks (as even within amd64 there doesn't seem to be a platform agnostic way to handle this).


efi/preinstall/checks_context_test.go
```diff
@@ -215,785 +215,632 @@ func (s *runChecksContextSuite) testRun(c *C, params *testRunChecksContextRunPar
...
-			tpm2.PropertyManufacturer:      uint32(tpm2.TPMManufacturerINTC),
-		},
-		enabledBanks:              []tpm2.HashAlgorithmId{tpm2.HashAlgorithmSHA256},
-		loadedImages:              efiImagesDefault(),
-		profileOpts:               PCRProfileOptionsDefault,
-		actions:                   []actionAndArgs{{action: ActionNone}},
-		expectedPcrAlg:            tpm2.HashAlgorithmSHA256,
-		expectedUsedSecureBootCAs: []*X509CertificateID{NewX509CertificateID(testutil.ParseCertificate(c, msUefiCACert))},
-		expectedFlags:             NoPlatformConfigProfileSupport | NoDriversAndAppsConfigProfileSupport | NoBootManagerConfigProfileSupport,
-		expectedWarningsMatch: `3 errors detected:
```
alexclewontin, line 259: Porting @frederic-hoerni's comment from the last PR:

These are worth being displayed when tests are run (for identifying which host fixture raised an error, or for comparison of which host fixtures get tested from a version to another...).

They appear with `-check.vv`, but this option also activates other cumbersome log messages.

So, is there a way to give better visibility when running tests? And possibly ensure that `Makefile` and `.github/workflows/test.yaml` (or `run-tests`) take advantage of it.

efi/preinstall/checks_fixture_test.go
```diff
@@ -0,0 +1,213 @@
...
+// for testing. Each fixture can define an EFI environment with CPU features, a set of sysfs
+// devices, virtualization information, and host ISA. Each fixture also includes a set of
+// "capabilities" which can be used by tests to decide whether they should expect a
+// certain fixture to work for a given test scenario or not.
+//
+// To add fixtures, append them to the array returned by this function. To add capabilities,
+// add them to the const block at the top of this file and then audit the existing fixtures
+// and add the new capability to any existing fixtures to which it might apply.
+func runChecksPlatformHostFixtures() []runChecksHostFixture {
+	intelDevices := func(status []byte, withIOMMU bool) []internal_efi.SysfsDevice {
```
alexclewontin, line 122: Porting @frederic-hoerni's comment from the last PR:

Instead of embedding a func here, wouldn't it be more legible to have it as a plain regular named function? That would also be easier to document and let runChecksPlatformHostFixtures() focus on the list that it returns?

alexclewontin (commented):
Ported comments from last PR

efi/preinstall/checks_fixture_test.go
```diff
@@ -0,0 +1,239 @@
...
+	return efitest.NewMockHostEnvironmentWithOpts(append(base, options...)...)
+}
+
+func (f *runChecksHostFixture) mockRuntimeGOARCH(s interface{ AddCleanup(func()) }) {
+	s.AddCleanup(MockRuntimeGOARCH(f.arch))
+}
+
+// logRunChecksHostFixture prints the fixture with the calling gocheck test's name.
+// It must be called directly from a test method.
+func logRunChecksHostFixture(fixture runChecksHostFixture) {
```
alexclewontin, line 118: Porting @frederic-hoerni's comment from the last PR (this was originally misplaced, and now actually points to the code addressing the comment):

These are worth being displayed when tests are run (for identifying which host fixture raised an error, or for comparison of which host fixtures get tested from a version to another...).

They appear with `-check.vv`, but this option also activates other cumbersome log messages.

So, is there a way to give better visibility when running tests? And possibly ensure that `Makefile` and `.github/workflows/test.yaml` (or `run-tests`) take advantage of it.

efi/preinstall/check_tpm_test.go
```diff
@@ -1149,3 +1150,70 @@ func (s *tpmSuite) TestOpenAndCheckTPM2DeviceLockoutAvailabilitySkipped(c *C) {
 	c.Assert(tpm, NotNil)
 	c.Check(dev.NumberOpen(), Equals, int(1))
 }
+
+type tpmAmd64Suite struct {
+	snapd_testutil.BaseTest
+}
```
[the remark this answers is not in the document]
alexclewontin: should be fixed

pedronis (commented):
the basic idea here seems sounds

here are some generally minor issues found by a LLM review:

4. and 6. might definitely be OK as is

## 1. Unreachable test assertions (`additionalExpectedFlags`)

The `additionalExpectedFlags` field is declared in `efi/preinstall/checks_fixture_test.go` (line 86)
but **no fixture ever sets it**, so it is always `0`. That makes these two blocks dead code:

- `efi/preinstall/checks_context_test.go` (line 892)
- `efi/preinstall/checks_context_test.go` (line 2481)

Each contains a full alternative set of assertions (`ErrorKindNoSuitablePCRBank`, specific PCR0
digests) plus a `continue` that would skip the trailing `c.Assert(errs, HasLen, 0)`. This is
anticipatory arm64 code that has never been executed and whose hardcoded digests cannot have been
validated. It is also ~75 `| fixture.additionalExpectedFlags` OR-ins across
`efi/preinstall/checks_test.go` and `efi/preinstall/checks_context_test.go` that are all no-ops
today.

Suggestion: drop the mechanism and reintroduce it together with the arm64 fixture that needs it.

## 2. Accidentally deleted TODO

`// TODO: Good test case for invalid PCR1 when we support it.` was dropped from
`efi/preinstall/checks_context_test.go` during the restructuring of `TestRunGoodInvalidPCR0Value`.
It looks unintentional and should be restored.

## 3. Layering smell in the CPUID drift guard

`CPUIDFeatureSMX` and `CPUIDFeatureSDBG` are declared in `internal/efi/env.go` (lines 70 and 74),
but the drift-guard test that pins them to the upstream `github.com/canonical/cpuid` values lives in
`internal/cpuid/cpuid_amd64_test.go` and therefore has to import `internal/efi` — the reverse of the
production dependency direction.

Suggestion: declare the constants in `internal/cpuid` (and reference or re-export them from
`internal/efi`), keeping the test local to the package that owns the knowledge.

## 4. CI never exercises the non-amd64 paths

`.github/workflows/test.yaml` (line 9) runs only on `ubuntu-24.04` (x64), so
`internal/cpuid/cpuid_null.go` and the general arch-neutrality claim are unverified by CI. Since the
point of the PR is "compiled on every architecture", a cheap
`GOARCH=arm64 go build ./... && GOARCH=arm64 go vet ./...` step would lock that in. It passes today.

## 5. Silent-skip risk in the fixture selector

`runChecksHostFixturesFor` in `efi/preinstall/checks_fixture_test.go` (line 89) calls `c.Skip(...)`
when no fixture matches the required capabilities. If a capability is mistyped or a fixture is
removed, a test silently stops running rather than failing.

Suggestion: at minimum, make the skip message name the missing capabilities.

## 6. Stub semantics are silently permissive

`internal/cpuid/cpuid_null.go` returns `false` from `HasFeature` (line 38) and `""` from
`VendorIdentificator` (line 26) on non-amd64. These are unreachable today because
`DefaultEnv.AMD64()` gates on `runtimeGOARCH` first, but a future direct caller would silently get a
"feature absent" answer for a security-relevant check. A comment is present; consider whether
panicking would be safer.

## 7. `logRunChecksHostFixture` implementation

`efi/preinstall/checks_fixture_test.go` (line 118) uses `fmt.Printf` plus `runtime.Caller` and
`panic` on name-parse failure. `c.Logf` would be more idiomatic and avoid the
reflection-on-function-names fragility, though it does not give the always-on visibility that was
requested in review.

alexclewontin:
> ## 1. Unreachable test assertions (`additionalExpectedFlags`)
> The `additionalExpectedFlags` field is declared in `efi/preinstall/checks_fixture_test.go` (line 86) but **no fixture ever sets it**, so it is always `0`. That makes these two blocks dead code:
> 
> * `efi/preinstall/checks_context_test.go` (line 892)
> * `efi/preinstall/checks_context_test.go` (line 2481)
> 
> Each contains a full alternative set of assertions (`ErrorKindNoSuitablePCRBank`, specific PCR0 digests) plus a `continue` that would skip the trailing `c.Assert(errs, HasLen, 0)`. This is anticipatory arm64 code that has never been executed and whose hardcoded digests cannot have been validated. It is also ~75 `| fixture.additionalExpectedFlags` OR-ins across `efi/preinstall/checks_test.go` and `efi/preinstall/checks_context_test.go` that are all no-ops today.
> 
> Suggestion: drop the mechanism and reintroduce it together with the arm64 fixture that needs it.

Fixed

> ## 2. Accidentally deleted TODO
> `// TODO: Good test case for invalid PCR1 when we support it.` was dropped from `efi/preinstall/checks_context_test.go` during the restructuring of `TestRunGoodInvalidPCR0Value`. It looks unintentional and should be restored.

Fixed

> ## 3. Layering smell in the CPUID drift guard
> `CPUIDFeatureSMX` and `CPUIDFeatureSDBG` are declared in `internal/efi/env.go` (lines 70 and 74), but the drift-guard test that pins them to the upstream `github.com/canonical/cpuid` values lives in `internal/cpuid/cpuid_amd64_test.go` and therefore has to import `internal/efi` — the reverse of the production dependency direction.
> 
> Suggestion: declare the constants in `internal/cpuid` (and reference or re-export them from `internal/efi`), keeping the test local to the package that owns the knowledge.

Fixed

> ## 4. CI never exercises the non-amd64 paths
> `.github/workflows/test.yaml` (line 9) runs only on `ubuntu-24.04` (x64), so `internal/cpuid/cpuid_null.go` and the general arch-neutrality claim are unverified by CI. Since the point of the PR is "compiled on every architecture", a cheap `GOARCH=arm64 go build ./... && GOARCH=arm64 go vet ./...` step would lock that in. It passes today.
> 

Concur this is a non-issue, as it is addressed in #566 (I like PR stacks, but lack of visibility on things like this is a downside).

> ## 5. Silent-skip risk in the fixture selector
> `runChecksHostFixturesFor` in `efi/preinstall/checks_fixture_test.go` (line 89) calls `c.Skip(...)` when no fixture matches the required capabilities. If a capability is mistyped or a fixture is removed, a test silently stops running rather than failing.
> 
> Suggestion: at minimum, make the skip message name the missing capabilities.

This is a good catch, and after some consideration I think that a test with no fixtures that meet the required capabilities should be considered a fail, not a skip, to avoid the risk of easy-to-overlook regression. Now errors out.

> ## 6. Stub semantics are silently permissive
> `internal/cpuid/cpuid_null.go` returns `false` from `HasFeature` (line 38) and `""` from `VendorIdentificator` (line 26) on non-amd64. These are unreachable today because `DefaultEnv.AMD64()` gates on `runtimeGOARCH` first, but a future direct caller would silently get a "feature absent" answer for a security-relevant check. A comment is present; consider whether panicking would be safer.

I mostly concur this is a non-issue, since this should only be called during testing of amd64 business logic on non-amd64 platforms. The only thing that might make it nicer is if we provided two alternative implementations, one which is silently permissive for testing, and one which panics if this is ever reached during actual operation. Not sure if the complexity is actually worth it though.

> 
> ## 7. `logRunChecksHostFixture` implementation
> `efi/preinstall/checks_fixture_test.go` (line 118) uses `fmt.Printf` plus `runtime.Caller` and `panic` on name-parse failure. `c.Logf` would be more idiomatic and avoid the reflection-on-function-names fragility, though it does not give the always-on visibility that was requested in review.

This would change the behavior. I have no strong opinion on *what* the desired behavior here should be, but if we want to implement what @frederic-hoerni suggested I believe this is the right *how*

PR #566, opened 2026-09-04
efi: implement platform-agnostic arm64 host security checks

alexclewontin:
This PR adds a framework for host security checks on arm64 platforms. It moves #564 to the stack.

It adds a generic framework to check host security on ARM64. Very few truly cross-platform APIs were found to be applicable to that ecosystem, but we can use the backing driver for the TPM to prove that a given TPM is a fTPM: if it is using the OP-TEE fTPM driver we can be sure that it is an fTPM. This is not conclusive (absence of that driver doesn't prove that a TPM is a dTPM), but it is a common reference implementation for the ARM ecosystem, so worth including.

### DIFF
--- a/efi/preinstall/check_host_security.go
+++ b/efi/preinstall/check_host_security.go
@@ -166,6 +166,8 @@ func checkHostSecurity(env internal_efi.HostEnvironment, log *tcglog.Log) (platf
 	switch runtimeGOARCH {
 	case "amd64":
 		return checkHostSecurityAMD64(env, log)
+	case "arm64":
+		return checkHostSecurityARM64(env, log)
 	default:
 		return platformFirmwareIntegrityNone, &UnsupportedPlatformError{fmt.Errorf("checking host security is not implemented on %s", runtimeGOARCH)}
 	}
@@ -178,6 +180,8 @@ func checkDiscreteTPMPartialResetAttackMitigationStatus(env internal_efi.HostEnv
 	switch runtimeGOARCH {
 	case "amd64":
 		return checkDiscreteTPMPartialResetAttackMitigationStatusAMD64(env, logResults)
+	case "arm64":
+		return checkDiscreteTPMPartialResetAttackMitigationStatusARM64(env, logResults)
 	default:
 		return dtpmPartialResetAttackMitigationNotRequired, nil
 	}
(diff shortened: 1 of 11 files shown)


efi/preinstall/check_host_security.go
```diff
@@ -295,3 +299,72 @@ func checkDiscreteTPMPartialResetAttackMitigationStatusAMD64(env internal_efi.Ho
...
+
+	integrity, err := checkHostSecurityARM64Platform(arm64Env, cpuManufacturer)
+	if err != nil {
+		return platformFirmwareIntegrityNone, err
+	}
+
+	return checkHostSecurityARM64Generic(env, log, integrity)
+}
+
+func checkHostSecurityARM64Generic(env internal_efi.HostEnvironment, log *tcglog.Log, integrity platformFirmwareIntegrityConfig) (platformFirmwareIntegrityConfig, error) {
```
frederic-hoerni, line 328: This does not look very Arm64 specific and duplicates code from checkHostSecurityAMD64.
Could both be merged? (does it make sense?)

Or do you intend to make them diverge in the next PR?

efi/preinstall/checks_fixture_test.go
```diff
@@ -139,6 +166,14 @@ func runChecksPlatformHostFixtures() []runChecksHostFixture {
 		)))
 	}
 
+	newDevices := func(tpmDriver string, withIOMMU bool) []internal_efi.SysfsDevice {
```
frederic-hoerni, line 169: same question as for `intelDevices` in PR#565: https://github.com/canonical/secboot/pull/565#discussion_r3934634871

efi/preinstall/checks_fixture_test.go
```diff
@@ -109,6 +109,33 @@ func (f *runChecksHostFixture) mockRuntimeGOARCH(s interface{ AddCleanup(func())
 	s.AddCleanup(MockRuntimeGOARCH(f.arch))
 }
 
+const (
+	exampleARM64CPUManufacturer = "Example Manufacturer"
+	exampleARM64CPUVersion      = "Example OP-TEE SoC"
+)
+
+func init() {
+	RegisterARM64TestPlatform(exampleARM64CPUManufacturer, exampleARM64CPUVersion)
```
frederic-hoerni, line 118: Please add a comment about what is the purpose of this, and why this is done here.

Wouldn't it be more explicit to have this done in the setup of one of the test suites instead?

efi/preinstall/export_test.go
```diff
@@ -188,3 +191,22 @@ func MockRuntimeGOARCH(arch string) (restore func()) {
 	runtimeGOARCH = arch
 	return func() { runtimeGOARCH = orig }
 }
+
+func RegisterARM64TestPlatform(cpuManufacturer, cpuVersion string) {
```
frederic-hoerni, line 195: Please add a description of this function.

internal/efi/default_env.go
```diff
@@ -50,10 +51,15 @@ var (
...
 	osReadlink               = os.Readlink
+	// runtimeGOARCH is the architecture that host security checks are performed
+	// for. It is a variable so that tests can run the checks for architectures
+	// other than the one the test binary was built for.
 	runtimeGOARCH            = runtime.GOARCH
 	tpm2_deviceDefaultDevice = tpm2_device.DefaultDevice
 
 	eventLogPath = "/sys/kernel/security/tpm0/binary_bios_measurements" // Path of the TCG event log for the default TPM, in binary form
+
+	dmiProcessorInfoPath = "/sys/firmware/dmi/entries/4-0/raw"
```
frederic-hoerni, line 62: Does "4" means type 4?
Does "0" means processor 0? Do we make any assumption there (eg: that other CPUs are the same as CPU 0...)
Please document these.

alexclewontin: Documented

alexclewontin: We do indeed assume that CPUs are all homogenous

PR #574, opened 2026-09-22
crypt.go: add function to verify key for specific keyslot

valentindavid:
In order to be able to prevent removing keyslots for which we have the only key for the containers, we need to be able to check specific keyslots.

We typically only know the key that got registered in keyring during boot. And they come from sealed key that we usually cannot unseal again later. If we remove any of those keyslots because an unexpected keyslot was used for activation, then we lose ability to act on the container, until we reboot and a different keyslot gets used.

Unexpected keyslot might used if we for example lose power or hard reset in the middle of reprovision.


crypt_test.go
```diff
@@ -4137,3 +4137,41 @@ func (s *cryptSuite) TestNameLegacyLUKS2ContainerKeyNameAlreadyUsed(c *C) {
...
+	}
+	c.Assert(InitializeLUKS2Container(path, "disk", key, initOptions), IsNil)
+
+	otherKey := s.newPrimaryKey()
+	c.Assert(AddLUKS2ContainerUnlockKey(path, "other", key, otherKey), IsNil)
+
+	recoveryKey := s.newRecoveryKey()
+	c.Assert(AddLUKS2ContainerRecoveryKey(path, "recovery", key, recoveryKey), IsNil)
+
+	check := func (name string, testKey []byte, expected bool) {
```
frederic-hoerni: `go fmt` detects a discrepancy here (space between "func" and parenthesis)

internal/luks2/cryptsetup.go
```diff
@@ -506,3 +506,13 @@ func SetSlotPriority(devicePath string, slot int, priority SlotPriority) error {
 func TestContainerKey(devicePath string, key []byte) bool {
 	return cryptsetupCmd(bytes.NewReader(key), "open", "--test-passphrase", "--key-file", "-", devicePath) == nil
 }
+
+// Check if key is valid key for LUKS2 container at devicePath for a specific keyslot.
+func TestContainerKeyForKeyslot(devicePath string, slot int, key []byte) bool {
+	err := cryptsetupCmd(bytes.NewReader(key), "open", "--test-passphrase", "--key-slot", strconv.Itoa(slot), "--key-file", "-", devicePath)
+	if err != nil {
+		fmt.Printf("got error: %v", err)
```
frederic-hoerni: In secboot, errors are generally printed to stderr.

crypt.go
```diff
@@ -1044,3 +1046,19 @@ func NameLegacyLUKS2ContainerKey(devicePath string, keyslot int, newName string)
...
+		return false, xerrors.Errorf("cannot obtain LUKS header view: %w", err)
+	}
+
+	_, id, exists := view.TokenByName(name)
+
+	if !exists {
+		return false, ErrKeyslotNameNotExist
+	}
+
+	return luks2.TestContainerKeyForKeyslot(devicePath, id, key), nil
```
frederic-hoerni: Is `id` the identifier of the token or the keyslot? They may have the same value in some cases, but this is not always true.

The id of the keyslot should be given here.

frederic-hoerni (changes requested):

internal/luks2/cryptsetup.go
valentindavid: Oh, that was debugging...

crypt.go
valentindavid: Oops.

frederic-hoerni (approved):

pedronis (approved):
thanks

